PR #213, opened 2026-03-28
feat: geo_filter polygon overlay on map and live pages (Go backend)

efiten:
## Summary

- Adds `GeoFilter` struct to `Config` in `cmd/server/config.go` so `geo_filter.polygon` and `bufferKm` from `config.json` are parsed by the Go backend
- Adds `GET /api/config/geo-filter` endpoint in `cmd/server/routes.go` returning the polygon + bufferKm to the frontend
- Restores the blue polygon overlay (solid inner + dashed buffer zone) on the **Map** page (`public/map.js`)
- Restores the same overlay on the **Live** page (`public/live.js`), toggled via the "Mesh live area" checkbox

## Test plan

- [x] `GET /api/config/geo-filter` returns `{ polygon: [...], bufferKm: N }` when configured
- [x] `GET /api/config/geo-filter` returns `{ polygon: null, bufferKm: 0 }` when not configured
- [x] Map page shows blue polygon overlay when `geo_filter.polygon` is set in config
- [x] Live page shows same overlay, checkbox state shared via localStorage
- [x] Checkbox is hidden when no polygon is configured

### DIFF
--- a/cmd/server/config.go
+++ b/cmd/server/config.go
@@ -47,12 +47,23 @@ type Config struct {
 	Retention *RetentionConfig `json:"retention,omitempty"`
 
 	PacketStore *PacketStoreConfig `json:"packetStore,omitempty"`
+
+	GeoFilter *GeoFilterConfig `json:"geo_filter,omitempty"`
 }
 
 // PacketStoreConfig controls in-memory packet store limits.
 type PacketStoreConfig struct {
 	RetentionHours float64 `json:"retentionHours"` // max age of packets in hours (0 = unlimited)
 	MaxMemoryMB    int     `json:"maxMemoryMB"`     // hard memory ceiling in MB (0 = unlimited)
+}
+
+type GeoFilterConfig struct {
+	Polygon  [][2]float64 `json:"polygon,omitempty"`
+	BufferKm float64      `json:"bufferKm,omitempty"`
+	LatMin   *float64     `json:"latMin,omitempty"`
+	LatMax   *float64     `json:"latMax,omitempty"`
+	LonMin   *float64     `json:"lonMin,omitempty"`
+	LonMax   *float64     `json:"lonMax,omitempty"`
 }
 
 type RetentionConfig struct {
--- a/cmd/server/routes.go
+++ b/cmd/server/routes.go
@@ -102,6 +102,7 @@ func (s *Server) RegisterRoutes(r *mux.Router) {
 	r.HandleFunc("/api/config/regions", s.handleConfigRegions).Methods("GET")
 	r.HandleFunc("/api/config/theme", s.handleConfigTheme).Methods("GET")
 	r.HandleFunc("/api/config/map", s.handleConfigMap).Methods("GET")
+	r.HandleFunc("/api/config/geo-filter", s.handleConfigGeoFilter).Methods("GET")
 
 	// System endpoints
 	r.HandleFunc("/api/health", s.handleHealth).Methods("GET")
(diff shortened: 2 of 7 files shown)


Kpa-clawbot (changes requested):
## Review: PR #213 — geo_filter polygon overlay

**Verdict: Request Changes** — Three blocking bugs in map.js and two AGENTS.md rule violations.

---

### 🔴 Blocking Bugs

**1. geoFilterLayer not declared in map.js — ReferenceError**
map.js uses 'use strict' (line 2). The diff adds geoFilterLayer = L.layerGroup(...) but never declares let geoFilterLayer. This will throw a ReferenceError at runtime. Compare with live.js where the diff correctly adds it to the declaration on line 8.

Fix: Add geoFilterLayer to the variable declarations at the top of map.js (after heatLayer on line 14).

**2. No HTML checkbox element in map.js — feature won't work on map page**
The diff calls getElementById('mcGeoFilterLabel') and getElementById('mcGeoFilter') but no HTML element with those IDs is created in map.js or index.html. The map controls template (lines 92-109) has mcClusters, mcHeatmap, mcHashLabels, mcNeighbors — but no mcGeoFilter. The null-check prevents a crash, but the checkbox will never appear and the overlay is unreachable on the map page.

Fix: Add a <label id="mcGeoFilterLabel" ...><input type="checkbox" id="mcGeoFilter"> Mesh live area</label> to the Display fieldset in the map controls template.

**3. No cleanup of geoFilterLayer in map.js destroy() — memory leak**
The destroy() function (line 720) cleans up markerLayer, 
outeLayer, heatLayer but the diff doesn't add geoFilterLayer cleanup. In contrast, live.js correctly adds geoFilterLayer = null to its cleanup line.

Fix: Add if (geoFilterLayer) { geoFilterLayer = null; } in destroy().

---

### 🟡 AGENTS.md Violations

**4. No tests added (Rule #1)**
The PR adds a new Go API endpoint (/api/config/geo-filter) and frontend logic (buffer zone computation) with zero tests. Rule #1: *"Every change that touches logic MUST have unit tests."* At minimum:
- Go test: /api/config/geo-filter returns correct shape with and without config
- The buffer zone computation has testable math

**5. Cache busters not bumped (Rule #3)**
public/index.html is not in the changed files, but both public/live.js and public/map.js were modified. Rule #3: *"Every time you change a .js or .css file in public/, bump the cache buster in index.html."* This has caused **7 production regressions** before.

---

### 🟠 DRY Violation

**6. ~40 lines of identical code copy-pasted between map.js and live.js**
The entire polygon overlay logic (fetch, inner polygon, buffer computation, layer group, checkbox binding, localStorage) is duplicated verbatim. AGENTS.md Engineering Principles: *"If the same logic exists in two places, it MUST be extracted into a shared function."*

Suggested: Extract a shared function like createGeoFilterOverlay(map, checkboxId, labelId) into a shared module (e.g., a new geo-overlay.js or added to an existing shared file). Both pages would call it with their respective element IDs.

---

### 💡 Minor Suggestions

**7. Hardcoded #3b82f6 in Leaflet polygon options**: live.js already has a cssVar() helper. Consider cssVar('--accent') || '#3b82f6' for theme consistency. Non-blocking since existing analytics code has the same pattern.

**8. Go struct has unused fields**: GeoFilterConfig defines LatMin/LatMax/LonMin/LonMax pointer fields, but the handler only returns polygon and ufferKm. Either these are for future use (YAGNI?) or they should be documented.

---

### ✅ What's Good

- Go endpoint handler is clean with proper nil/empty checks
- live.js implementation is complete (declaration, HTML, cleanup all present)
- localStorage state sharing between pages is a nice UX touch
- Silent fallback on API error is appropriate

jjkroell:
Is this a feature to filter out nodes outside the boundary or nodes and packets? I had this going pretty good before the GO change.

PR #215, opened 2026-03-28
feat: geo_filter enforcement, DB pruning, geofilter-builder tool, HB column

efiten:
## Summary

Several features and fixes from a live deployment of the Go v3.0.0 backend.

### geo_filter — full enforcement

- **Go backend config** (`cmd/server/config.go`, `cmd/ingestor/config.go`): added `GeoFilterConfig` struct so `geo_filter.polygon` and `bufferKm` from `config.json` are parsed by both the server and ingestor
- **Ingestor** (`cmd/ingestor/geo_filter.go`, `cmd/ingestor/main.go`): ADVERT packets from nodes outside the configured polygon + buffer are dropped *before* any DB write — no transmission, node, or observation data is stored
- **Server API** (`cmd/server/geo_filter.go`, `cmd/server/routes.go`): `GET /api/config/geo-filter` endpoint returns the polygon + bufferKm to the frontend; `/api/nodes` responses filter out any out-of-area nodes already in the DB
- **Frontend** (`public/map.js`, `public/live.js`): blue polygon overlay (solid inner + dashed buffer zone) on Map and Live pages, toggled via "Mesh live area" checkbox, state shared via localStorage

### Automatic DB pruning

- Add `retention.packetDays` to `config.json` to delete transmissions + observations older than N days on a daily schedule (1 min after startup, then every 24h). Nodes and observers are never pruned.
- `POST /api/admin/prune?days=N` for manual runs (requires `X-API-Key` header if `apiKey` is set)

```json
"retention": {
  "nodeDays": 7,
  "packetDays": 30
}
```

### tools/geofilter-builder.html

Standalone HTML tool (no server needed) — open in browser, click to place polygon points on a Leaflet map, set `bufferKm`, copy the generated `geo_filter` JSON block into `config.json`.

### scripts/prune-nodes-outside-geo-filter.py

Utility script to clean existing out-of-area nodes from the database (dry-run + confirm). Useful after first enabling geo_filter on a populated DB.

### HB column in packets table

Shows the hop hash size in bytes (1–4) decoded from the path byte of each packet's raw hex. Displayed as **HB** between Size and Type columns, hidden on small screens.

## Test plan

- [x] ADVERT from node outside polygon is not stored (no new row in nodes or transmissions)
- [x] `GET /api/config/geo-filter` returns polygon + bufferKm when configured, `{polygon: null, bufferKm: 0}` when not
- [x] `/api/nodes` excludes nodes outside polygon even if present in DB
- [x] Map and Live pages show blue polygon overlay when configured; checkbox toggles it
- [x] `retention.packetDays: 30` deletes old transmissions/observations on startup and daily
- [x] `POST /api/admin/prune?days=30` returns `{deleted: N, days: 30}`
- [x] `tools/geofilter-builder.html` opens standalone, draws polygon, copies valid JSON
- [x] HB column shows 1–4 for all packets in grouped and flat view

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,6 +9,7 @@ ARG BUILD_TIME=unknown
 # Build server
 WORKDIR /build/server
 COPY cmd/server/go.mod cmd/server/go.sum ./
+COPY internal/geofilter/ ../../internal/geofilter/
 RUN go mod download
 COPY cmd/server/ ./
 RUN go build -ldflags "-X main.Version=${APP_VERSION} -X main.Commit=${GIT_COMMIT} -X main.BuildTime=${BUILD_TIME}" -o /corescope-server .
--- a/cmd/ingestor/config.go
+++ b/cmd/ingestor/config.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"os"
 	"strings"
+
+	"github.com/meshcore-analyzer/geofilter"
 )
 
 // MQTTSource represents a single MQTT broker connection.
@@ -34,7 +36,11 @@ type Config struct {
 	ChannelKeys     map[string]string `json:"channelKeys,omitempty"`
 	HashChannels    []string          `json:"hashChannels,omitempty"`
 	Retention       *RetentionConfig  `json:"retention,omitempty"`
+	GeoFilter       *GeoFilterConfig  `json:"geo_filter,omitempty"`
 }
+
+// GeoFilterConfig is an alias for the shared geofilter.Config type.
+type GeoFilterConfig = geofilter.Config
 
 // RetentionConfig controls how long stale nodes are kept before being moved to inactive_nodes.
 type RetentionConfig struct {
--- a/cmd/ingestor/geo_filter.go
+++ b/cmd/ingestor/geo_filter.go
@@ -1,0 +1,73 @@
+package main
+
+import "math"
+
+// NodePassesGeoFilter returns true if the node should be kept.
+// Nodes with no GPS coordinates are always allowed.
+func NodePassesGeoFilter(lat, lon *float64, gf *GeoFilterConfig) bool {
+	if gf == nil {
+		return true
+	}
+	if lat == nil || lon == nil {
+		return true
+	}
+	if *lat == 0 && *lon == 0 {
+		return true
+	}
+	if len(gf.Polygon) >= 3 {
+		if pointInPolygon(*lat, *lon, gf.Polygon) {
+			return true
+		}
+		if gf.BufferKm > 0 {
+			n := len(gf.Polygon)
+			for i := 0; i < n; i++ {
+				j := (i + 1) % n
+				if distToSegmentKm(*lat, *lon, gf.Polygon[i], gf.Polygon[j]) <= gf.BufferKm {
+					return true
+				}
+			}
+		}
+		return false
+	}
+	// Legacy bounding box
+	if gf.LatMin != nil && gf.LatMax != nil && gf.LonMin != nil && gf.LonMax != nil {
+		return *lat >= *gf.LatMin && *lat <= *gf.LatMax && *lon >= *gf.LonMin && *lon <= *gf.LonMax
+	}
+	return true
+}
+
+func pointInPolygon(lat, lon float64, polygon [][2]float64) bool {
+	inside := false
+	n := len(polygon)
+	j := n - 1
+	for i := 0; i < n; i++ {
+		yi, xi := polygon[i][0], polygon[i][1]
+		yj, xj := polygon[j][0], polygon[j][1]
+		if (yi > lat) != (yj > lat) {
+			if lon < (xj-xi)*(lat-yi)/(yj-yi)+xi {
+				inside = !inside
+			}
+		}
+		j = i
+	}
+	return inside
+}
+
+func distToSegmentKm(lat, lon float64, a, b [2]float64) float64 {
+	lat1, lon1 := a[0], a[1]
+	lat2, lon2 := b[0], b[1]
+	cosLat := math.Cos((lat1 + lat2) / 2.0 * math.Pi / 180.0)
+	ax := (lon1 - lon) * 111.0 * cosLat
+	ay := (lat1 - lat) * 111.0
+	bx := (lon2 - lon) * 111.0 * cosLat
+	by := (lat2 - lat) * 111.0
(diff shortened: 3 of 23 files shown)


Kpa-clawbot (changes requested):
## PR #215 Review — Kobayashi (Lead)

Good feature set (geo_filter enforcement, pruning, HB column, tooling). The architecture decisions are sound — filtering ADVERTs at ingestion time is the right call, and the prune design (separate rw connection, transaction) is correct. However, there are several issues that need to be addressed before merge.

---

### 🔴 Blocking

**1. Missing cache busters (AGENTS.md Rule 3)**
Changed `public/live.js`, `public/map.js`, `public/packets.js`, `public/style.css` — but `public/index.html` is not in the diff. This rule exists because it caused **7 production regressions**. Must bump cache busters in the same commit.

**2. No unit tests (AGENTS.md Rule 1)**
Zero test files in this PR. The geo filter logic (`pointInPolygon`, `distToSegmentKm`, `NodePassesGeoFilter`), HB column parsing (`(parseInt(hex.slice(2,4), 16) >> 6) + 1`), and `PruneOldPackets` all need tests. This is non-negotiable per AGENTS.md: *"Every change that touches logic MUST have unit tests."*

Needed at minimum:
- `cmd/ingestor/geo_filter_test.go` — inside/outside polygon, buffer zone edge, null coords, 0/0 coords, legacy bbox, nil config
- `cmd/server/geo_filter_test.go` — same plus `toFloat64` type switch coverage
- `PruneOldPackets` — test with seeded DB rows

**3. `/api/nodes` geo filter breaks pagination and counts**
The geo filter is applied **after** `GetNodes()` returns a paginated, counted result:
`go
nodes, total, counts, err := s.db.GetNodes(limit, offset, ...)
// ... later ...
if s.cfg.GeoFilter != nil {
    filtered := nodes[:0]  // filter post-pagination
    total = len(filtered)  // total is wrong (reflects this page only)
    // counts is NEVER updated — role counts include filtered-out nodes
}
`

Two bugs:
- **`counts` is stale**: role counts from the DB include nodes outside the polygon. Frontend sidebar showing "50 repeaters" when only 30 are visible.
- **Pagination is broken**: if DB returns 50 rows and filter removes 20, the client gets 30 rows for a page of 50. Offset-based pagination will skip/duplicate nodes unpredictably.

Fix: push the geo filter into the DB query (`WHERE` clause with lat/lon bounds) or fetch all nodes first, filter, then paginate in Go.

**4. Admin prune endpoint open when no `apiKey` configured**
`go
if s.cfg.APIKey != "" && r.Header.Get("X-API-Key") != s.cfg.APIKey {
`
If `apiKey` is empty (the default), anyone can `POST /api/admin/prune?days=1` and delete all packet data. This is a destructive endpoint — it should require auth unconditionally. Either require an API key to be set for this endpoint, or add a separate admin key.

---

### 🟡 Important

**5. DRY: geo filter geometry duplicated 3×**
`pointInPolygon` and `distToSegmentKm` are copy-pasted across:
- `cmd/ingestor/geo_filter.go`
- `cmd/server/geo_filter.go`  
- `scripts/prune-nodes-outside-geo-filter.py`

The two Go copies are identical except function name prefixes (`geo`-prefixed in server). Extract to a shared `internal/geofilter/` package. The Python script can stay separate (utility tool).

**6. DRY: `GeoFilterConfig` struct duplicated**
Same struct in `cmd/ingestor/config.go` and `cmd/server/config.go`. Extract to shared package (same `internal/geofilter/`).

**7. Hardcoded hex colors in frontend (AGENTS.md violation)**
`map.js` and `live.js` both hardcode `#3b82f6` for the polygon overlay. Per AGENTS.md: *"All colors MUST use CSS variables."* Add a `--geo-filter-color` CSS variable and reference it.

**8. Frontend geo filter overlay duplicated between `map.js` and `live.js`**
~43 lines of identical code. Extract to a shared function (e.g., in a helper or `roles.js`) that takes a map instance and returns a layer group.

**9. Python script has hardcoded Belgium/Netherlands polygon**
`scripts/prune-nodes-outside-geo-filter.py` has a deployment-specific polygon baked in. Should read from `config.json` (it even has a comment suggesting this) or accept a `--config` flag. Shipping specific coordinates in a public repo is questionable.

---

### 🟢 Minor / Notes

**10. `time.Tick` in prune goroutine (line ~380 in main.go)**
`time.Tick` leaks the underlying ticker (can't be GC'd). Use `time.NewTicker` + defer `Stop()`. Minor since goroutine lives forever, but bad practice.

**11. Buffer zone approximation**
Centroid-based vertex scaling fails on concave polygons (vertices can cross). Acceptable as documented ("approximate"), but worth a code comment noting the concave limitation.

**12. 11 commits for 4 features**
AGENTS.md Rule 6: one commit per logical change. This bundles geo_filter + pruning + HB column + tools. Consider squash-merging or splitting into focused PRs.

---

### Summary
Request changes on the 4 blocking items. The geo filter at ingestion is well-designed, prune architecture is solid, and the geofilter-builder tool is a nice addition. Fix the pagination/counts bug, add tests, bump cache busters, and secure the prune endpoint, then this is ready.

Kpa-clawbot (approved):
## PR #215 Re-review — Kobayashi (Lead)

All 4 blockers from the previous review are resolved:

| # | Blocker | Status | Evidence |
|---|---------|--------|----------|
| 1 | Missing cache busters | ✅ Fixed | All refs in \index.html\ bumped \=1774690966\ → \=1780272000\ |
| 2 | No tests for geo filter | ✅ Fixed | \cmd/ingestor/geo_filter_test.go\ (7 tests), \cmd/server/geo_filter_test.go\ (multiple tests), \TestPruneOldPackets\ in \db_test.go\ |
| 3 | /api/nodes geo filter breaks pagination & counts | ✅ Fixed | \handleNodes\ now fetches all nodes when geo filter is active, filters in-memory, recomputes \	otal\ + \counts\, THEN applies offset/limit. Non-geo path unchanged. |
| 4 | Prune endpoint unauthenticated | ✅ Fixed | \handleAdminPrune\ requires \piKey\ configured + valid \X-API-Key\ header. Returns 401 without both. |

**Nit (non-blocking):** The geo-filtered \/api/nodes\ path fetches up to 1M rows to filter in Go. Fine for typical deployments, but worth a TODO comment if node counts ever reach 5-figures — push the polygon check into SQL at that point.

Good work addressing all the feedback. Ship it. 🚢

PR #222, opened 2026-03-28
perf: eliminate O(n) slice prepend on every packet ingest

efiten:
## Problem

`s.packets` and `s.byPayloadType[t]` were maintained in newest-first order by **prepending** every new packet:

```go
s.packets = append([]*StoreTx{tx}, s.packets...)
s.byPayloadType[pt] = append([]*StoreTx{tx}, s.byPayloadType[pt]...)
```

Each prepend allocates a new backing array and copies the entire existing slice. With a large dataset (e.g. 30 days of data ≈ 2–3 M packets × 8 bytes/pointer), this means **~24 MB copied per ingest cycle**. The poller runs every second, so any active MQTT traffic causes continuous memory copying and GC pressure — observed as sustained 100–140 % CPU in production.

## Fix

- Store both slices **oldest-first** (append to tail — zero copy).
- `Load()` SQL changed from `ORDER BY first_seen DESC` → `ASC` so the initial load appends naturally.
- `QueryPackets` DESC (default) now reads backwards from the tail in **O(page\_size)** — no sort or copy needed.
- `QueryPackets` ASC reads forwards as before.
- `GetChannelMessages` switches from reverse-iteration to forward-iteration (slice is now already oldest-first).

## Impact

| | Before | After |
|---|---|---|
| Cost per new packet ingest | O(n) slice copy | O(1) append |
| QueryPackets DESC (default) | O(1) slice window | O(page\_size) reverse read |
| QueryPackets ASC | O(n log n) sort | O(1) slice window |
| GetChannelMessages | O(n) reverse loop | O(n) forward loop |

All existing tests pass.

## Test plan
- [x] Run `go test ./...` in `cmd/server/` — all pass
- [x] Deploy and observe CPU drop on a production instance with a large DB
- [x] Verify packets list shows newest-first, pagination works correctly
- [x] Verify channel messages show in chronological order

### DIFF
--- a/cmd/server/store.go
+++ b/cmd/server/store.go
@@ -176,7 +176,7 @@ func (s *PacketStore) Load() error {
 			FROM transmissions t
 			LEFT JOIN observations o ON o.transmission_id = t.id
 			LEFT JOIN observers obs ON obs.rowid = o.observer_idx
-			ORDER BY t.first_seen ASC, o.timestamp ASC`
+			ORDER BY t.first_seen ASC, o.timestamp DESC`
 	} else {
 		loadSQL = `SELECT t.id, t.raw_hex, t.hash, t.first_seen, t.route_type,
 				t.payload_type, t.payload_version, t.decoded_json,
@@ -184,7 +184,7 @@ func (s *PacketStore) Load() error {
 				o.snr, o.rssi, o.score, o.path_json, o.timestamp
 			FROM transmissions t
 			LEFT JOIN observations o ON o.transmission_id = t.id
-			ORDER BY t.first_seen ASC, o.timestamp ASC`
+			ORDER BY t.first_seen ASC, o.timestamp DESC`
 	}
 
 	rows, err := s.db.conn.Query(loadSQL)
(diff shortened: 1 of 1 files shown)


Kpa-clawbot (approved):
Reviewed cmd/server/store.go in depth.

✅ What looks good:
- Core perf fix is correct: replacing prepend (ppend([]*StoreTx{...}, ...)) with tail append removes O(n) copy on ingest.
- Ordering updates are consistently propagated across impacted paths (Load, QueryPackets, QueryMultiNodePackets, GetTimestamps, GetChannelMessages, GetNodeHealth).
- No data-loss behavior introduced (in-memory indexes and DB writes remain intact; this is read/order logic only).
- AGENTS.md-specific frontend rules (cache busters/colors) are not applicable since this PR is backend-only.

⚠️ Validation note:
- I could not run go test ./... locally due environment toolchain mismatch (modernc.org/libc ... undefined: syscall.SyscallN, module requires Go 1.20). This appears environment-related, not caused by this diff.

Given code consistency and risk profile, approving.

PR #269, opened 2026-03-29
feat: add multi-byte hash usage matrix with stats and improved tooltips

jjkroell:
- Add 1/2/3-byte selector to Hash Issues analytics page
- 1-byte and 2-byte modes show 16×16 matrix with stat cards (nodes tracked, using N-byte ID, prefix space used, prefix collisions)
- 3-byte mode shows summary stat cards instead of unrenderable grid
- Fix "Nodes tracked" to always show total node count across all modes
- Use CSS variable colours for matrix cells (light/dark mode compatible)
- Replace native title tooltips with custom styled popovers
- Hide collision risk card when 3-byte mode is selected
- Fix double-tooltip bug on mode switch via _matrixTipInit guard
- Fix tooltip persisting outside matrix grid on mouseleave

https://dev.ve7kod.ca/#/analytics

Hash Issues

### DIFF
--- a/public/analytics.js
+++ b/public/analytics.js
@@ -1167,8 +1167,9 @@
     if (bytes === 3) {
       const total = totalNodes.length;
       const threeByteNodes = allNodes.filter(n => n.hash_size === 3).length;
+      const nodesForByte = allNodes.filter(n => n.hash_size === 3 || !n.hash_size);
       const prefixMap = {};
-      for (const n of allNodes) {
+      for (const n of nodesForByte) {
         const p = n.public_key.slice(0, 6).toUpperCase();
         if (!prefixMap[p]) prefixMap[p] = 0;
         prefixMap[p]++;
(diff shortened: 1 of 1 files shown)


Kpa-clawbot (commented):
## Review: Multi-byte hash matrix

Nice feature — the byte selector UX and stat cards are clean. A few issues:

### 🔴 Missing cache busters
Both `analytics.js` and `style.css` are modified but `index.html` cache busters aren't bumped. This is the #1 pitfall in AGENTS.md (happened 7 times before). Please update in the same commit.

### 🟡 Performance: O(n²) prefix filtering
In both 1-byte and 2-byte modes, `allNodes.filter(n => n.public_key.toUpperCase().startsWith(hex))` runs inside a loop of 256 iterations. For a mesh with hundreds of nodes, that's tens of thousands of string operations. Consider building a single prefix map in one pass:
```js
const prefixNodes = {};
for (const n of allNodes) {
  const hex = n.public_key.slice(0, 2).toUpperCase();
  (prefixNodes[hex] ||= []).push(n);
}
```
Same pattern applies to the 2-byte `groupNodes.filter` inner loop.

### 🟡 Tooltip element lifecycle
`initMatrixTooltip` appends a `div` to `document.body` and guards against re-init via `el._matrixTipInit`. But when switching byte modes, `el.innerHTML` is replaced — the tooltip `div` stays orphaned in the body. On rapid switching this could accumulate. Consider: (a) removing the tooltip div on mode switch, or (b) using a single shared tooltip div created once at the module level.

### 🟡 `hash_size` property assumption
The code uses `n.hash_size === 1`, `n.hash_size === 2`, etc. for the "Using N-byte ID" stat cards. Is this field reliably present on node objects from the API? If not, these counts will always be 0. If it's a new field, mention it in the PR description.

### 🟡 No tests for new logic
The 2-byte grouping logic, 3-byte summary stats, and collision detection changes have meaningful branching. Consider adding test coverage — especially for the `renderCollisions` refactor where the fallback to node-scanning for 2/3-byte modes is new behavior.

### ✅ No server.js changes — good
### ✅ CSS uses variables for theming — good
### ✅ 3-byte "too large to visualize" is a smart UX decision

PR #328, opened 2026-03-31
perf: fix slow /api/packets and /api/channels on large stores

efiten:
## Problem

Two endpoints were slow on larger installations:

**`/packets?limit=50000&groupByHash=true` — 16s+**
`QueryGroupedPackets` did two expensive things on every request:
1. O(n × observations) scan per packet to find `latest` timestamp
2. Held `s.mu.RLock()` during the O(n log n) sort, blocking all concurrent reads

**`/channels` — 13s+**
`GetChannels` iterated all payload-type-5 packets and JSON-unmarshaled each one while holding `s.mu.RLock()`, blocking all concurrent reads for the full duration.

## Fix

**Packets (`QueryGroupedPackets`):**
- Add `LatestSeen string` to `StoreTx`, maintained incrementally in all three observation write paths. Eliminates the per-packet observation scan at query time.
- Build output maps under the read lock, sort the local copy after releasing it.
- Cache the full sorted result for 3 seconds keyed by filter params.

**Channels (`GetChannels`):**
- Copy only the fields needed (firstSeen, decodedJSON, region match) under the read lock, then release before JSON unmarshaling.
- Cache the result for 15 seconds keyed by region param.
- Invalidate cache on new packet ingestion.

## Test plan
- [ ] Open packets page on a large store — load time should drop from 16s to <1s
- [ ] Open channels page — should load in <100ms instead of 13s+
- [ ] `[SLOW API]` warnings gone for both endpoints
- [ ] Packet/channel data is correct (hashes, counts, observer counts)
- [ ] Filters (region, type, since/until) still work correctly

### DIFF
--- a/cmd/server/routes_test.go
+++ b/cmd/server/routes_test.go
@@ -2402,3 +2402,206 @@ func min(a, b int) int {
 	}
 	return b
 }
+
+// TestLatestSeenMaintained verifies that StoreTx.LatestSeen is populated after Load()
+// and is >= FirstSeen for packets that have observations.
+func TestLatestSeenMaintained(t *testing.T) {
+	db := setupTestDB(t)
+	defer db.Close()
+	seedTestData(t, db)
+
+	store := NewPacketStore(db, nil)
+	if err := store.Load(); err != nil {
+		t.Fatalf("store.Load failed: %v", err)
+	}
+
+	store.mu.RLock()
+	defer store.mu.RUnlock()
+
+	if len(store.packets) == 0 {
+		t.Fatal("expected packets in store after Load")
+	}
+
+	for _, tx := range store.packets {
+		if tx.LatestSeen == "" {
+			t.Errorf("packet %s has empty LatestSeen (FirstSeen=%s)", tx.Hash, tx.FirstSeen)
+			continue
+		}
+		// LatestSeen must be >= FirstSeen (string comparison works for RFC3339/ISO8601)
+		if tx.LatestSeen < tx.FirstSeen {
+			t.Errorf("packet %s: LatestSeen %q < FirstSeen %q", tx.Hash, tx.LatestSeen, tx.FirstSeen)
+		}
+		// For packets with observations, LatestSeen must be >= all observation timestamps.
+		for _, obs := range tx.Observations {
+			if obs.Timestamp != "" && obs.Timestamp > tx.LatestSeen {
+				t.Errorf("packet %s: obs.Timestamp %q > LatestSeen %q", tx.Hash, obs.Timestamp, tx.LatestSeen)
+			}
+		}
+	}
+}
+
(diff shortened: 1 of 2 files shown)


Kpa-clawbot (approved):
## Code Review: perf fix for grouped packets and channels

**Tests:** All pass (`go test ./...` — 3.75s) ✅

### What was reviewed

Read the full diff for `store.go` (+360/-79) and `routes_test.go` (+206) against master.

### Algorithmic changes — Correct ✅

**LatestSeen maintenance:** Added to all three write paths (Load, IngestNewFromDB, IngestNewObservations) with consistent `if obs.Timestamp > tx.LatestSeen` logic. Initialized to `FirstSeen` on creation. This eliminates the O(observations) scan per packet at query time. The invariant (LatestSeen ≥ FirstSeen, LatestSeen ≥ max(obs.Timestamp)) is well-tested.

**Lock scope reduction:** Both `QueryGroupedPackets` and `GetChannels` now copy data under the read lock and do expensive work (sort / JSON unmarshal) outside it. This is the right pattern — no correctness issues since the copied data is owned by the goroutine after unlock.

**Caching:** Grouped packets get a 3s TTL cache keyed by all filter dimensions. Channels get a 15s TTL cache invalidated on ingestion. Both are sensible choices — grouped cache's short TTL makes explicit invalidation unnecessary, and channels cache IS invalidated on new packet ingestion in both `IngestNewFromDB` and `IngestNewObservations`.

### Edge cases — Handled ✅

- Empty store: `filterPackets` returns empty slice → empty entries → cache stores empty result → pagination returns `{Packets: [], Total: 0}`
- Single packet with no observations: `LatestSeen = FirstSeen` (set at creation), sort works fine
- Offset beyond total: `pagePacketResult` returns empty slice with correct total
- Cache key covers all filter dimensions including optional pointer fields (Type, Route)

### One minor observation (non-blocking)

`pagePacketResult` returns a sub-slice of the cached `PacketResult.Packets`. Since callers only JSON-serialize the result, there's no mutation risk. But if a future caller ever modified a returned map in-place, it would corrupt the cache. A comment noting this shared-reference would be defensive, but it's fine as-is given the current usage.

### Test coverage — Good ✅

5 new tests covering:
- `TestLatestSeenMaintained` — invariant after Load
- `TestQueryGroupedPacketsSortedByLatest` — sort order correctness (old first_seen with recent obs sorts before new first_seen with old obs)
- `TestQueryGroupedPacketsCacheReturnsConsistentResult` — cache consistency
- `TestGetChannelsCacheReturnsConsistentResult` — cache consistency
- `TestGetChannelsNotBlockedByLargeLock` — lock-copy pattern produces correct results

### Code style — Consistent ✅

Follows existing patterns (separate cache mutex, same naming conventions, same map building style).

LGTM — clean performance fix with good test coverage.

PR #332, opened 2026-03-31
fix: add migration for missing observations.timestamp index

efiten:
## Problem

On installations where the database predates the `idx_observations_timestamp` index, `/api/stats` takes 30s+ because `GetStoreStats()` runs two full table scans:

```sql
SELECT COUNT(*) FROM observations WHERE timestamp > ?  -- last hour
SELECT COUNT(*) FROM observations WHERE timestamp > ?  -- last 24h
```

The index is only created in the `if !obsExists` block, so any database where the `observations` table already existed before that code was added never gets it.

## Fix

Adds a one-time migration (`obs_timestamp_index_v1`) that runs at ingestor startup:

```sql
CREATE INDEX IF NOT EXISTS idx_observations_timestamp ON observations(timestamp)
```

On large installations this index creation may take a few seconds on first startup after the upgrade, but subsequent stats queries become instant.

## Test plan
- [ ] Restart ingestor on an older database and confirm `[migration] observations timestamp index created` appears in logs
- [ ] Confirm `/api/stats` response time drops from 30s+ to <100ms

### DIFF
--- a/cmd/ingestor/db.go
+++ b/cmd/ingestor/db.go
@@ -280,6 +280,17 @@ func applySchema(db *sql.DB) error {
 		log.Println("[migration] node telemetry columns added")
 	}
 
+	// One-time migration: add timestamp index on observations for fast stats queries.
+	// Older databases created before this index was added suffer from full table scans
+	// on COUNT(*) WHERE timestamp > ?, causing /api/stats to take 30s+.
+	row = db.QueryRow("SELECT 1 FROM _migrations WHERE name = 'obs_timestamp_index_v1'")
+	if row.Scan(&migDone) != nil {
+		log.Println("[migration] Adding timestamp index on observations...")
+		db.Exec(`CREATE INDEX IF NOT EXISTS idx_observations_timestamp ON observations(timestamp)`)
+		db.Exec(`INSERT INTO _migrations (name) VALUES ('obs_timestamp_index_v1')`)
+		log.Println("[migration] observations timestamp index created")
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


Kpa-clawbot (approved):
Code review of the migration logic and tests:

**✅ Migration pattern** — Follows the exact same pattern as existing migrations (`advert_count_unique_v1`, `noise_floor_real_v1`, `node_telemetry_v1`): check `_migrations` table → run if absent → record completion.

**✅ Idempotency** — `CREATE INDEX IF NOT EXISTS` handles the case where the index already exists (e.g., DB created after the inline schema added it at line 168).

**✅ New DB path** — New databases get the index inline during table creation (line 168), so the migration becomes a no-op (migration row may or may not exist, but the index is there either way). Tests verify this.

**✅ Test coverage** — Two cases covered: (1) fresh DB has the index, (2) old DB without the index gets it via migration. Both verify index existence in `sqlite_master` and migration record in `_migrations`. Thorough.

**✅ PR description** — Accurate, explains the problem (full table scans on older DBs), the fix, and the expected behavior.

**Minor note (non-blocking):** The `db.Exec` calls for `CREATE INDEX` and `INSERT INTO _migrations` don't check errors. If the index creation fails silently, the migration still records as done and won't retry. This matches the pattern of older migrations (`advert_count_unique_v1`, `noise_floor_real_v1`) but differs from the newer `node_telemetry_v1` which does check errors. Consider adding error checks for consistency with the newer pattern — but this is cosmetic since `CREATE INDEX IF NOT EXISTS` on SQLite is unlikely to fail if the table exists.

LGTM 👍

PR #459, opened 2026-04-01
fix: VCR timeline and clock respect UTC/local timezone setting

### LINKED ISSUE #324: bug: VCR timeline ignores UTC/local timezone setting
## Problem

The VCR timeline on the Live page always shows local time, ignoring the UTC/local timezone setting in the customizer Display tab.

## Expected behavior

The timeline should respect the global timestamp timezone setting from the customizer. When set to UTC, timeline labels should show UTC times.

## Related
- #286 (timestamp display enhancement) — the shared `formatTimestamp()` utility respects the setting, but the VCR timeline likely uses its own time formatting that wasn't updated.

## Files to check
- `public/live.js` — VCR timeline rendering, look for `toLocaleTimeString` or direct Date formatting that bypasses the shared formatter

efiten:
## Problem

Fixes #324. The VCR LCD clock and timeline hover/touch tooltip always showed local time, ignoring the UTC/local timezone setting in the customizer Display tab.

## Root cause

Three sites in `live.js` bypassed the shared `getTimestampTimezone()` utility:

- `updateVCRClock()` — used `d.getHours()` / `d.getMinutes()` / `d.getSeconds()` (always local)
- Timeline mousemove tooltip — used `d.toLocaleTimeString()` (always local)
- Timeline touchmove tooltip — same

## Fix

Added `vcrFormatTime(tsMs)` helper that checks `getTimestampTimezone()` and uses `getUTC*` methods when set to `'utc'`, otherwise local `get*`. Applied to all three sites. Exposed as `window._vcrFormatTime` for testing.

## Tests

4 new unit tests in `test-frontend-helpers.js` covering UTC mode, local mode, and zero-padding.

## Checklist
- [x] Branches from `upstream/master`
- [x] No Matomo or local-only commits
- [x] Cache busters bumped (`v=1775073838`)
- [x] 233 tests pass, 0 fail

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -22,9 +22,9 @@
   <meta name="twitter:title" content="CoreScope">
   <meta name="twitter:description" content="Real-time MeshCore LoRa mesh network analyzer — live packet visualization, node tracking, channel decryption, and route analysis.">
   <meta name="twitter:image" content="https://raw.githubusercontent.com/Kpa-clawbot/corescope/master/public/og-image.png">
-  <link rel="stylesheet" href="style.css?v=1775062162">
-  <link rel="stylesheet" href="home.css?v=1775062162">
-  <link rel="stylesheet" href="live.css?v=1775062162">
+  <link rel="stylesheet" href="style.css?v=1775073838">
+  <link rel="stylesheet" href="home.css?v=1775073838">
+  <link rel="stylesheet" href="live.css?v=1775073838">
   <link rel="stylesheet" href="https://unpkg.com/leaflet@1.9.4/dist/leaflet.css"
     integrity="sha256-p4NxAoJBhIIN+hmNHrzRCf9tD/miZyoHS5obTRR9BMY="
     crossorigin="anonymous">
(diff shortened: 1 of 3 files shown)


Kpa-clawbot (approved):
## Independent review

Reviewed the full diff cold. This is a clean, well-scoped fix.

### What it does
Extracts a `vcrFormatTime(tsMs)` helper that checks `getTimestampTimezone()` and uses `getUTC*()` vs `get*()` accordingly. Applied consistently to all 3 sites that were bypassing the timezone setting (VCR clock, mousemove tooltip, touchmove tooltip).

### Observations

**Positive:**
- Defensive `typeof getTimestampTimezone === 'function'` guard — won't break if `customize.js` fails to load
- DRY: single helper replaces duplicated logic in 3 places
- Tests cover UTC mode, local mode, zero-padding, and exposure
- Cache busters bumped consistently

**Minor (non-blocking):**
- The tooltip previously used `toLocaleTimeString()` which respected locale formatting (e.g. 12h vs 24h). The new helper always uses 24h format. This is arguably more consistent with the VCR/LCD aesthetic but is a subtle behavior change for users with 12h locale in local-time mode. Worth noting in case users report it.

**No issues found.** 229 existing tests pass on master. The 4 new tests are well-structured.

✅ Approved.

PR #461, opened 2026-04-01
perf: upsert known nodes in-place on ADVERT, skip full reload

### LINKED ISSUE #399: perf(frontend): nodes page invalidates entire cache and refetches 5K nodes on every ADVERT batch
## Summary
The nodes page WebSocket handler invalidates the node cache and reloads ALL nodes on every ADVERT packet. While `debouncedOnWS` with 5s delay batches multiple adverts, each batch still triggers a full `/nodes?limit=5000` fetch and complete table re-render.

## Problematic Code
**File:** `public/nodes.js`, ~line 280:

```js
wsHandler = debouncedOnWS(function (msgs) {
  if (msgs.some(isAdvertMessage)) {
    _allNodes = null;                    // Invalidate ALL cached nodes
    invalidateApiCache('/nodes');         // Bust API cache too
    loadNodes(true);                     // Full reload from server
  }
}, 5000);
```

## Root Cause
A single new advert invalidates the entire node cache and fetches all 5000 nodes again. The `refreshOnly` flag only skips rebuilding the filter bar — it still fetches all data.

## Severity: **Minor**
The 5s debounce and API TTL caching (90s) limit the impact. But on an active mesh with frequent adverts, this still refreshes every 5s.

## Suggested Fix
1. Merge the new ADVERT data directly into `_allNodes` without re-fetching everything
2. Extract the new node info from the WS message and upsert into the cached array
3. Only fetch from server on initial load or explicit refresh

## Related
- Architecture: "fetch once, filter client-side"

efiten:
## Problem

Fixes #399. On every ADVERT WebSocket batch the nodes page invalidated the entire `_allNodes` cache and triggered a full `/nodes?limit=5000` fetch — even when every advertising node was already cached. The 90s API TTL was actively bypassed.

## Root cause

```js
wsHandler = debouncedOnWS(function (msgs) {
  if (msgs.some(isAdvertMessage)) {
    _allNodes = null;               // wipe cache unconditionally
    invalidateApiCache('/nodes');   // bust API TTL
    loadNodes(true);                // full 5k fetch
  }
}, 5000);
```

## Fix

ADVERT decoded payloads include `pubKey`, `name`, `lat`, `lon` — enough to update known nodes in place:

- **Known node** (pubKey found in `_allNodes`): upsert `name`, `lat`, `lon`, `last_seen` directly — no fetch, no cache bust, just re-render.
- **New node** (pubKey not in cache) or **no pubKey** in payload: fall back to full reload as before.

This covers the common case on an active mesh: all advertising nodes are already cached. The full reload path is preserved for node discovery.

## Tests

2 new unit tests: known-node upsert (asserts 0 API calls, fields updated) and unknown-node fallback (asserts full reload triggered). All 231 tests pass.

## Checklist
- [x] Branches from `upstream/master`
- [x] No Matomo or local-only commits
- [x] Cache busters bumped
- [x] 231 tests pass, 0 fail

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -22,9 +22,9 @@
   <meta name="twitter:title" content="CoreScope">
   <meta name="twitter:description" content="Real-time MeshCore LoRa mesh network analyzer — live packet visualization, node tracking, channel decryption, and route analysis.">
   <meta name="twitter:image" content="https://raw.githubusercontent.com/Kpa-clawbot/corescope/master/public/og-image.png">
-  <link rel="stylesheet" href="style.css?v=1775062162">
-  <link rel="stylesheet" href="home.css?v=1775062162">
-  <link rel="stylesheet" href="live.css?v=1775062162">
+  <link rel="stylesheet" href="style.css?v=1775077121">
+  <link rel="stylesheet" href="home.css?v=1775077121">
+  <link rel="stylesheet" href="live.css?v=1775077121">
   <link rel="stylesheet" href="https://unpkg.com/leaflet@1.9.4/dist/leaflet.css"
     integrity="sha256-p4NxAoJBhIIN+hmNHrzRCf9tD/miZyoHS5obTRR9BMY="
     crossorigin="anonymous">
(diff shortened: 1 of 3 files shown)


Kpa-clawbot (approved):
## Independent Review

**Verdict: APPROVE** ✅

Reviewed the full diff cold (3 commits across 3 files, +115/-30).

### What it does
Optimizes the ADVERT WebSocket handler on the nodes page: instead of wiping the cache and doing a full `/nodes?limit=5000` fetch on every ADVERT batch, it upserts known nodes in-place (by pubKey) and only falls back to full reload for genuinely new nodes.

### Code quality
- **Logic is correct.** The upsert path mutates `_allNodes` entries in-place, then calls `loadNodes(true)` which skips the API fetch (guarded by `if (!_allNodes)`) and just re-renders. Clean.
- **Map optimization** (commit 6efbe2c) — O(1) lookup per advert instead of O(n) `.find()`. Map is rebuilt per debounced batch (5s), which is fine.
- **Edge cases handled well:**
  - `_allNodes` is null (not yet loaded) → full reload
  - Missing pubKey in payload → full reload
  - Unknown pubKey → full reload
  - Mixed known/unknown batch → breaks early on first unknown, does full reload (safe, slightly conservative but correct)
- **`payload.lat != null`** — correctly allows `0` as a valid coordinate value (vs falsy check).
- **Test hooks** (`_nodesGetAllNodes`, `_nodesSetAllNodes`) follow the existing pattern.

### Tests
All 231 tests pass. The 2 new tests cover:
1. Known-node upsert: verifies 0 API calls, 0 cache invalidations, and fields updated in-place
2. Unknown-node fallback: verifies full reload + cache invalidation triggered

### Minor observations (non-blocking)
- The `gh pr diff` output was stale (showed `.find()` instead of `Map`), but the actual branch head has the Map fix. Probably a GitHub caching artifact.
- Cache buster bumped ✅

Clean, well-scoped perf fix. LGTM.

PR #462, opened 2026-04-01
fix: null-guard animLayer and liveAnimCount in nextHop after destroy

efiten:
## Summary
- `nextHop()` schedules `setInterval`/`setTimeout` callbacks that can fire after `destroy()` has set `animLayer = null` and removed DOM elements
- This caused three console errors on the Live page when navigating away mid-animation: `Cannot read properties of null (reading 'hasLayer')` and `Cannot set properties of null (setting 'textContent')`
- Added null guards at each async callback site; no behavioral change when the page is active

## Changes
- `public/live.js`: early return if `animLayer` is null at start of `nextHop()`; null-safe `animLayer.hasLayer` checks in `setInterval`/`setTimeout`; null-safe `liveAnimCount` element access
- `public/index.html`: cache buster bumped
- `test-frontend-helpers.js`: 4 source-inspection tests verifying the null guards are present

## Test plan
- [ ] Open Live page, trigger some packet animations, navigate away quickly — no console errors
- [ ] `node test-frontend-helpers.js` passes (233 tests, 0 failures)

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -22,9 +22,9 @@
   <meta name="twitter:title" content="CoreScope">
   <meta name="twitter:description" content="Real-time MeshCore LoRa mesh network analyzer — live packet visualization, node tracking, channel decryption, and route analysis.">
   <meta name="twitter:image" content="https://raw.githubusercontent.com/Kpa-clawbot/corescope/master/public/og-image.png">
-  <link rel="stylesheet" href="style.css?v=1775062162">
-  <link rel="stylesheet" href="home.css?v=1775062162">
-  <link rel="stylesheet" href="live.css?v=1775062162">
+  <link rel="stylesheet" href="style.css?v=1775078686">
+  <link rel="stylesheet" href="home.css?v=1775078686">
+  <link rel="stylesheet" href="live.css?v=1775078686">
   <link rel="stylesheet" href="https://unpkg.com/leaflet@1.9.4/dist/leaflet.css"
     integrity="sha256-p4NxAoJBhIIN+hmNHrzRCf9tD/miZyoHS5obTRR9BMY="
     crossorigin="anonymous">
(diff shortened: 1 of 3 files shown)


Kpa-clawbot (approved):
## Independent review

**Verdict: Approve** ✅

### Summary
Clean, minimal fix for a real race condition — async callbacks (/) in `nextHop()` firing after `destroy()` nulls `animLayer` and removes DOM elements.

### What I checked
1. **Diff review** — All 4 null guards are correctly placed at the right async boundaries:
   - Early return at top of `nextHop()` if `animLayer` is null
   - `setInterval` pulse callback guards `animLayer` before `hasLayer()`
   - `setTimeout` cleanup guards `animLayer` before `hasLayer()` and `removeLayer()`
   - `liveAnimCount` DOM element null-checked before `.textContent` assignment
2. **No behavioral change** when page is active — guards only trigger post-destroy
3. **Tests** — ran `node test-frontend-helpers.js`: 233 passed, 0 failed
4. **Cache buster** bumped in index.html (cosmetic, correct)

### Minor observations (non-blocking)
- The source-inspection tests check for exact string presence rather than behavioral correctness. This is fine for regression guarding but brittle if the code is refactored. Not worth changing for this PR.
- The `activeAnims` decrement still runs even when `animLayer` is null (the early return is *after* the count update block). This is correct — the counter should still decrement to stay accurate.

No critical or major issues found.

Kpa-clawbot (approved):
## Independent review

**Verdict: Approve** ✅

### Summary
Clean, minimal fix for a real race condition — async callbacks (setInterval/setTimeout) in nextHop() firing after destroy() nulls animLayer and removes DOM elements.

### What I checked
1. **Diff review** — All 4 null guards are correctly placed at the right async boundaries:
   - Early return at top of nextHop() if animLayer is null
   - setInterval pulse callback guards animLayer before hasLayer()
   - setTimeout cleanup guards animLayer before hasLayer() and removeLayer()
   - liveAnimCount DOM element null-checked before .textContent assignment
2. **No behavioral change** when page is active — guards only trigger post-destroy
3. **Tests** — ran node test-frontend-helpers.js: 233 passed, 0 failed
4. **Cache buster** bumped in index.html (cosmetic, correct)

### Minor observations (non-blocking)
- The source-inspection tests check for exact string presence rather than behavioral correctness. Fine for regression guarding but brittle if code is refactored. Not worth changing for this PR.
- The activeAnims decrement still runs even when animLayer is null (the early return is after the count update block). This is correct — the counter should still decrement to stay accurate.

No critical or major issues found.

PR #491, opened 2026-04-02
fix: use packet timestamp in bufferPacket instead of arrival time (#475)

### LINKED ISSUE #475: Live map - message history
On the live map page - the message history timestamp always shows as "0s ago".  Time stamp does not change.

efiten:
## Summary
- `bufferPacket()` was overwriting `_ts` with `Date.now()` (receive time) for every live WS packet
- Packets arriving in the same batch all got identical timestamps, making the message history show the same "Xs ago" for every entry (e.g., all show "5s ago")
- Fix: use `pkt.timestamp || pkt.created_at` (mirroring `dbPacketToLive`) so each packet reflects its actual origination time, falling back to `Date.now()` only when the packet has no timestamp

## Root cause
```js
// before
pkt._ts = Date.now();

// after
pkt._ts = new Date(pkt.timestamp || pkt.created_at || Date.now()).getTime();
```

The WS broadcast includes `timestamp` (= `tx.FirstSeen`) in the packet map (store.go:1182), so the field is always present for real packets.

## Test plan
- [x] Open Live page, observe packets arriving — each should show its own relative time, not all the same value
- [x] `node test-frontend-helpers.js` passes (235 tests, 0 failures)

Closes #475

### DIFF
--- a/public/live.js
+++ b/public/live.js
@@ -482,7 +482,7 @@
   });
 
   function bufferPacket(pkt) {
-    pkt._ts = Date.now();
+    pkt._ts = new Date(pkt.timestamp || pkt.created_at || Date.now()).getTime();
     const entry = { ts: pkt._ts, pkt };
     VCR.buffer.push(entry);
     // Keep buffer capped at ~2000 — adjust playhead to avoid stale indices (#63)
--- a/public/live.js
+++ b/public/live.js
@@ -481,8 +481,13 @@
     }
   });
 
+  function packetTimestamp(pkt) {
+    return new Date(pkt.timestamp || pkt.created_at || Date.now()).getTime();
+  }
+  if (typeof window !== 'undefined') window._live_packetTimestamp = packetTimestamp;
+
   function bufferPacket(pkt) {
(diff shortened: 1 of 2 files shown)


Kpa-clawbot (approved):
## Independent review

Clean, correct fix. The bug is real — `bufferPacket` was overwriting `_ts` with `Date.now()`, causing all batch-arriving packets to show identical timestamps.

**What's good:**
- `packetTimestamp()` correctly mirrors the existing `dbPacketToLive` logic (`pkt.timestamp || pkt.created_at`) with an additional `Date.now()` fallback for robustness
- Server confirms `timestamp` is always present in WS broadcast (`store.go` sets it to `tx.FirstSeen`)
- Tests verify the priority chain (timestamp > created_at > Date.now()) and aren't tautological — they check actual behavior
- Test exposure pattern (`window._live_packetTimestamp`) follows established codebase convention

**Minor note (out-of-scope):** `dbPacketToLive` (line 441) lacks the `Date.now()` fallback — if both `timestamp` and `created_at` are missing, it would produce `NaN`. Not introduced by this PR, but worth noting.

LGTM. ✅

Issue #303, opened 2026-03-31
Repeater Hash Stats not changing based on most recent advert


kizniche:
Under Analytics and Hash Stats, there is a section By Repeaters, showing 1-byte, 2-byte, and 3-byte repeaters. I have 66 repeaters listed as 1-byte, then I changed my repeater to 2-byte and made a flood advert (which CoreScope saw my repeater Advert listed as a 2-byte hash size), but the stats didn't change my repeater to a 2-byte repeater, it stayed as 66 1-byte repeaters and 0 2-byte repeaters. My assumption would be the stats are generated based on the most recent advert.

efiten:
I think I may have addressed this already: https://github.com/Kpa-clawbot/CoreScope/pull/285

vaizki:
Hey @Kpa-clawbot hey @Kpa-clawbot can you reopen this and check if it's possible to ignore the zero-hop adverts when updating hash size? I had to revert this on my local install but you can fix it, right?

PR #493, opened 2026-04-02
fix: skip zero-hop adverts when checking node hash size

vaizki:
Fixes issue router IDs flapping between 1byte and multi-byte as described in https://github.com/Kpa-clawbot/CoreScope/issues/303 with a minimal patch + test coverage.

This fix is critical for regions using multi-byte IDs.

Closes https://github.com/Kpa-clawbot/CoreScope/issues/303

### DIFF
--- a/cmd/server/routes_test.go
+++ b/cmd/server/routes_test.go
@@ -2256,6 +2256,46 @@ func TestGetNodeHashSizeInfoOnlyDirectZeroHopIgnored(t *testing.T) {
 	}
 }
 
+func TestGetNodeHashSizeInfoDirectNonZeroHopCounted(t *testing.T) {
+	// A DIRECT advert with non-zero hop count should NOT be skipped —
+	// only zero-hop DIRECT adverts misreport hash size.
+	db := setupTestDB(t)
+	seedTestData(t, db)
+	store := NewPacketStore(db, nil)
+	if err := store.Load(); err != nil {
+		t.Fatalf("store.Load failed: %v", err)
+	}
+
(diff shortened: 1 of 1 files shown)


Kpa-clawbot (changes requested):
## Independent Review

**The core fix is correct** — DIRECT zero-hop adverts misreport hash size and should be excluded from `computeNodeHashSizeInfo`. The implementation is clean and minimal.

### Must-fix

1. **Missing test: DIRECT advert with non-zero hops should NOT be skipped.**
   The two new tests cover (a) DIRECT zero-hop mixed with FLOOD, and (b) DIRECT zero-hop only. But there is no test for a DIRECT advert with a non-zero hop count (e.g. `pathByte = 0x41` — hop count 1, hash size 2). This is the boundary condition: the fix skips `(pathByte & 0x3F) == 0`, so a DIRECT with hops > 0 should still be counted. Without this test, a future refactor could accidentally skip ALL direct adverts. Add a test like `TestGetNodeHashSizeInfoDirectNonZeroHopCounted` that inserts a DIRECT advert with non-zero path byte and asserts it IS included in hash size info.

### Looks good

- Route type extraction (`header & 0x03`) matches `decoder.go` constants
- The `RouteDirect` constant is correctly referenced (value 2)
- Test raw hex values are consistent with the protocol (header byte encodes route type in bits 0-1, path byte upper 2 bits = hash size)
- `TestGetNodeHashSizeInfoOnlyDirectZeroHopIgnored` correctly asserts `nil` — node with only ignored adverts should not appear in results
- Minimal, surgical change to production code — no unnecessary refactoring

PR #497, opened 2026-04-02
fix: apply region filter in GetNodes — was silently ignored (#496)

### LINKED ISSUE #496: fix: node list region filter has no effect (GetNodes ignores region param)
### Summary
Selecting a region in the node list filter has no effect — all nodes are always shown regardless of which region is selected.

### Root Cause
`db.GetNodes()` in `cmd/server/db.go` accepts a `region string` parameter (passed correctly from the HTTP handler at line 846) but **never uses it** when building the SQL WHERE clause. The parameter is silently ignored.

All other region-filtered endpoints (channels, analytics, bulk health) correctly apply the region filter via observer IATA joins. `GetNodes` simply never received this implementation.

### Expected Behavior
When a region is selected (e.g. SJC), only nodes whose ADVERT transmissions were observed by observers with that IATA code should be returned.

### Fix
Add a subquery to `GetNodes` that filters `nodes.public_key` against `JSON_EXTRACT(transmissions.decoded_json, '$.pubKey')` for ADVERT transmissions (payload_type=4) observed by observers with matching IATA codes. Handles both v2 (`observer_id TEXT`) and v3 (`observer_idx INTEGER`) schemas.

### Related
- Node list region column and active/stale column are separate feature requests (not covered by this fix)

efiten:
## Summary
- `db.GetNodes` accepted a `region` param from the HTTP handler but never used it — every region-filter selection was silently ignored and all nodes were always returned
- Added a subquery filtering `nodes.public_key` against ADVERT transmissions (payload_type=4) observed by observers with matching IATA codes
- Handles both v2 (`observer_id TEXT`) and v3 (`observer_idx INT`) schemas

## Test plan
- [x] 4 new subtests added to `TestGetNodesFiltering`: SJC (1 node), SFO (1 node), SJC,SFO multi (1 node deduped), AMS unknown (0 nodes)
- [x] All existing Go tests still pass
- [x] Deploy to staging, open `/nodes`, select a region in the filter bar — only nodes observed by observers in that region should appear

Closes #496

### DIFF
--- a/cmd/server/db.go
+++ b/cmd/server/db.go
@@ -698,6 +698,32 @@ func (db *DB) GetNodes(limit, offset int, role, search, before, lastHeard, sortB
 		}
 	}
 
+	if region != "" {
+		codes := normalizeRegionCodes(region)
+		if len(codes) > 0 {
+			placeholders := make([]string, len(codes))
+			regionArgs := make([]interface{}, len(codes))
+			for i, c := range codes {
+				placeholders[i] = "?"
+				regionArgs[i] = c
+			}
+			joinCond := "obs.rowid = o.observer_idx"
+			if !db.isV3 {
+				joinCond = "obs.id = o.observer_id"
+			}
+			subq := fmt.Sprintf(`public_key IN (
+				SELECT DISTINCT JSON_EXTRACT(t.decoded_json, '$.pubKey')
+				FROM transmissions t
+				JOIN observations o ON o.transmission_id = t.id
+				JOIN observers obs ON %s
+				WHERE t.payload_type = 4
+				AND UPPER(TRIM(obs.iata)) IN (%s)
+			)`, joinCond, strings.Join(placeholders, ","))
+			where = append(where, subq)
+			args = append(args, regionArgs...)
+		}
+	}
+
 	w := ""
 	if len(where) > 0 {
 		w = "WHERE " + strings.Join(where, " AND ")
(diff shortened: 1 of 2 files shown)


Kpa-clawbot (changes requested):
## Independent Review

The fix is correct and well-structured — the SQL subquery follows existing patterns in the codebase (e.g., lines 551, 613–615 in db.go), parameterization prevents SQL injection, and `JSON_EXTRACT` returning NULL on missing/malformed `decoded_json` is a safe no-match.

### Must-fix

**1. v2 schema path (`obs.id = o.observer_id`) is completely untested.**

`setupTestDB` hardcodes `isV3: true`. All four new region-filter subtests exercise only the v3 join (`obs.rowid = o.observer_idx`). The v2 branch could have a typo or wrong column reference and nothing would catch it.

Add at least one region-filter subtest using a v2-schema helper (with `observer_id TEXT` in observations and `isV3: false`). This is a real code path — if any production instance still runs v2 schema, a broken join would return zero nodes silently, which is the exact same class of bug this PR fixes.

---

Everything else (normalizeRegionCodes logic, DISTINCT dedup, UPPER/TRIM on IATA codes, test assertions, performance characteristics of the subquery) looks solid.

Kpa-clawbot (approved):
## Independent Review (Re-review)

Previous must-fix (v2 schema test coverage) has been addressed. Verified in the diff:

- `setupTestDBV2` helper creates in-memory DB with `observer_id TEXT` schema and `isV3: false`
- `TestGetNodesRegionFilterV2` covers both match (LAX) and no-match (JFK) cases
- The v2 join path (`obs.id = o.observer_id`) is now exercised

The core fix (region subquery in `GetNodes`) is clean — parameterized SQL, handles both schema versions, DISTINCT dedup, UPPER/TRIM normalization. No issues found.

✅ Approved.

PR #499, opened 2026-04-02
perf: skip JSON parse in indexByNode when no pubkey fields present (#376)

### LINKED ISSUE #376: perf: indexByNode parses full JSON for every packet during Load (store.go)
## Problem

`indexByNode()` in `store.go` parses the full JSON for every packet to extract pubkeys:

```go
func (s *PacketStore) indexByNode(tx *StoreTx) {
    if tx.DecodedJSON == "" {
        return
    }
    var decoded map[string]interface{}
    if json.Unmarshal([]byte(tx.DecodedJSON), &decoded) != nil {
        return
    }
    for _, field := range []string{"pubKey", "destPubKey", "srcPubKey"} {
        ...
    }
}
```

Called during `Load()` for every packet (30K+), and during `IngestNewFromDB()` for every new packet. JSON parsing is expensive relative to the simple field extraction needed.

## Severity

**Minor** — happens during load (once) and ingest (low rate). The JSON is typically small. But with 30K+ packets at startup, this adds up.

## Fix

Use `strings.Contains` to check for field presence first, then parse only if needed. Or use a faster JSON field extractor (e.g., `gjson.Get`).

Related: #360

efiten:
## Summary
- `indexByNode` was calling `json.Unmarshal` for every packet during `Load()` and `IngestNewFromDB()`, even channel messages and other payloads that can never contain node pubkey fields
- All three target fields (`"pubKey"`, `"destPubKey"`, `"srcPubKey"`) share the common substring `"ubKey"` — added a `strings.Contains` pre-check that skips the JSON parse entirely for packets that don't match
- At 30K+ packets on startup, this eliminates the majority of `json.Unmarshal` calls in `indexByNode` (channel messages, status packets, etc. all bypass it)

## Test plan
- [x] 5 new subtests in `TestIndexByNodePreCheck`: ADVERT with pubKey indexed, destPubKey indexed, channel message skipped, empty JSON skipped, duplicate hash deduped
- [x] All existing Go tests pass
- [x] Deploy to staging and verify node-filtered packet queries still work correctly

Closes #376

### DIFF
--- a/cmd/server/coverage_test.go
+++ b/cmd/server/coverage_test.go
@@ -3738,6 +3738,14 @@ func TestIndexByNodePreCheck(t *testing.T) {
 		}
 	})
 
+	t.Run("indexes srcPubKey", func(t *testing.T) {
+		tx := &StoreTx{Hash: "h2b", DecodedJSON: `{"srcPubKey":"112233","type":"TXT_MSG"}`}
+		store.indexByNode(tx)
+		if len(store.byNode["112233"]) != 1 {
+			t.Errorf("expected 1 entry for srcPubKey 112233, got %d", len(store.byNode["112233"]))
+		}
+	})
+
 	t.Run("skips channel message without pubKey", func(t *testing.T) {
 		beforeLen := len(store.byNode)
 		tx := &StoreTx{Hash: "h3", DecodedJSON: `{"type":"CHAN","channel":"#test","text":"hello"}`}
--- a/cmd/server/coverage_test.go
+++ b/cmd/server/coverage_test.go
@@ -3715,3 +3715,52 @@ func TestGetChannelMessagesAfterIngest(t *testing.T) {
 		t.Errorf("newest message should be 'brand new message', got %q", lastMsg["text"])
 	}
 }
+
+func TestIndexByNodePreCheck(t *testing.T) {
+	store := &PacketStore{
+		byNode:     make(map[string][]*StoreTx),
+		nodeHashes: make(map[string]map[string]bool),
+	}
+
+	t.Run("indexes ADVERT with pubKey", func(t *testing.T) {
+		tx := &StoreTx{Hash: "h1", DecodedJSON: `{"pubKey":"AABBCC","type":"ADVERT"}`}
+		store.indexByNode(tx)
+		if len(store.byNode["AABBCC"]) != 1 {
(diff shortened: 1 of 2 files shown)


Kpa-clawbot (changes requested):
## Independent Review

### The optimization itself is correct and safe

The `strings.Contains(tx.DecodedJSON, "ubKey")` pre-check is sound. It can only produce false positives (unnecessary parse when a *value* contains "ubKey"), never false negatives (all three target field names contain "ubKey" as a substring). The early return for empty `DecodedJSON` was already there; this layers on top cleanly.

Tests cover the key scenarios (each pubkey field type, skip on no match, empty JSON, dedup). No tautological tests — they exercise real `indexByNode` behavior.

### Must-fix

**1. Missing performance proof — required by AGENTS.md Rule 0**

The PR title is `perf:` and the description claims this "eliminates the majority of `json.Unmarshal` calls." AGENTS.md Rule 0 is explicit:

> Perf claims require proof. "This is faster" without data is not acceptable. Every PR claiming to fix or improve performance MUST include one of:
> - A benchmark test (before/after timings with realistic data sizes)
> - Profile output or timing measurements
> - A test assertion that enforces the perf characteristic

None of these are present. Add a `BenchmarkIndexByNode` (or equivalent) showing before/after, or at minimum a test that asserts the skip behavior at scale (e.g., 10K packets, measure that <N% trigger JSON parse). The project rules are clear — no proof, no merge.

Kpa-clawbot (approved):
## Independent Review (re-review)

Previous must-fix (missing benchmark proof) has been addressed. Benchmark added in `BenchmarkIndexByNode` showing ~17x speedup for the skip path vs full JSON parse.

**Optimization correctness:** The `strings.Contains(tx.DecodedJSON, "ubKey")` pre-check is sound — all three target fields (`pubKey`, `destPubKey`, `srcPubKey`) contain `ubKey`. False positives (value containing "ubKey") are harmless (just an unnecessary parse). False negatives are impossible.

**Tests:** 6 subtests + benchmark. Cover all three pubkey field types, skip on no match, empty JSON, and dedup. Not tautological — they exercise real `indexByNode` behavior.

**Code change:** 4 lines in `store.go`, clean and minimal.

No must-fix items. Approved. ✅

PR #500, opened 2026-04-02
fix: restore color-coded hex breakdown in packet detail (#329)

### LINKED ISSUE #329: Regression: packet hex breakdown lost color-coded section highlighting
## Description

The packet details pane hex breakdown displays sections (Header, Transport Codes, Path, Payload) but all sections use identical styling — there is no color differentiation between sections to visually distinguish which bytes belong to which part of the packet.

## Expected Behavior

Each section of the hex breakdown should have a distinct background tint so users can quickly identify header bytes vs transport codes vs path hops vs payload data at a glance — similar to how Wireshark color-codes protocol layers.

For example:
- **Header** rows: blue tint
- **Transport Codes** rows: green tint
- **Path** rows: amber/yellow tint
- **Payload** rows: purple tint

## Actual Behavior

All section header rows use the same CSS class `section-row` with a single background color (`var(--section-bg, #eef2ff)`), and all field rows under every section share identical unstyled `<tr>` elements with no section-specific class.

## Root Cause

The `sectionRow()` and `fieldRow()` functions in `public/packets.js` have never applied per-section CSS classes:

```js
// packets.js lines 1573-1577 (current)
function sectionRow(label) {
    return `<tr class="section-row"><td colspan="4">${label}</td></tr>`;
}
function fieldRow(offset, name, value, desc) {
    return `<tr><td class="mono">${offset}</td><td>${name}</td><td class="mono">${value}</td><td class="text-muted">${desc || ''}</td></tr>`;
}
```

And `public/style.css` has only one rule for all section rows:

```css
.field-table .section-row td {
  background: var(--section-bg, #eef2ff); font-weight: 700; font-size: 11px;
  text-transform: uppercase; letter-spacing: .5px; color: var(--accent);
}
```

This has been the case since the initial commit (`4634917`). No per-section color coding was ever implemented — the uniform styling was carried forward through all subsequent changes to the hex breakdown (commits adding transport codes, path rendering, payload decoding, etc.).

## Suggested Fix

1. Pass a section identifier to `sectionRow()` and `fieldRow()`:
   ```js
   function sectionRow(label, section) {
       return `<tr class="section-row section-${section}"><td colspan="4">${label}</td></tr>`;
   }
   function fieldRow(offset, name, value, desc, section) {
       return `<tr class="field-${section}"><td class="mono">${offset}</td><td>${name}</td><td class="mono">${value}</td><td class="text-muted">${desc || ''}</td></tr>`;
   }
   ```

2. Add CSS variables and rules for each section with distinct background tints (respecting both light and dark themes via `--section-bg-header`, `--section-bg-transport`, etc.).

efiten:
## Summary
- `BuildBreakdown` was never ported from the deleted Node.js `decoder.js` to Go — the server has returned `breakdown: {}` since the Go migration (commit `742ed865`), so `createColoredHexDump()` and `buildHexLegend()` in the frontend always received an empty `ranges` array and rendered everything as monochrome
- Implemented `BuildBreakdown()` in `decoder.go` — computes labeled byte ranges matching the frontend's `LABEL_CLASS` map: `Header`, `Transport Codes`, `Path Length`, `Path`, `Payload`; ADVERT packets get sub-ranges: `PubKey`, `Timestamp`, `Signature`, `Flags`, `Latitude`, `Longitude`, `Name`
- Wired into `handlePacketDetail` (was `struct{}{}`)
- Also adds per-section color classes to the field breakdown table (`section-header`, `section-transport`, `section-path`, `section-payload`) so the table rows get matching background tints

## Test plan
- [x] Open any packet detail pane — hex dump should show color-coded sections (red header, orange path length, blue transport codes, green path hops, yellow/colored payload)
- [x] Legend below action buttons should appear with color swatches
- [x] ADVERT packets: PubKey/Timestamp/Signature/Flags each get their own distinct color
- [x] Field breakdown table section header rows should be tinted per section
- [x] 8 new Go tests: all pass

Closes #329

### DIFF
--- a/cmd/server/decoder.go
+++ b/cmd/server/decoder.go
@@ -397,6 +397,106 @@ func DecodePacket(hexString string) (*DecodedPacket, error) {
 	}, nil
 }
 
+// HexRange represents a labeled byte range for the hex breakdown visualization.
+type HexRange struct {
+	Start int    `json:"start"`
+	End   int    `json:"end"`
+	Label string `json:"label"`
+}
+
+// Breakdown holds colored byte ranges returned by the packet detail endpoint.
+type Breakdown struct {
+	Ranges []HexRange `json:"ranges"`
+}
+
+// BuildBreakdown computes labeled byte ranges for each section of a MeshCore packet.
+// The returned ranges are consumed by createColoredHexDump() and buildHexLegend()
+// in the frontend (public/app.js).
+func BuildBreakdown(hexString string) *Breakdown {
+	hexString = strings.ReplaceAll(hexString, " ", "")
+	hexString = strings.ReplaceAll(hexString, "\n", "")
+	hexString = strings.ReplaceAll(hexString, "\r", "")
+	buf, err := hex.DecodeString(hexString)
+	if err != nil || len(buf) < 2 {
+		return &Breakdown{Ranges: []HexRange{}}
+	}
+
+	var ranges []HexRange
+	offset := 0
+
+	// Byte 0: Header
+	ranges = append(ranges, HexRange{Start: 0, End: 0, Label: "Header"})
+	offset = 1
+
+	header := decodeHeader(buf[0])
+
+	// Bytes 1-4: Transport Codes (TRANSPORT_FLOOD / TRANSPORT_DIRECT only)
+	if isTransportRoute(header.RouteType) {
+		if len(buf) < offset+4 {
+			return &Breakdown{Ranges: ranges}
+		}
+		ranges = append(ranges, HexRange{Start: offset, End: offset + 3, Label: "Transport Codes"})
+		offset += 4
+	}
+
+	if offset >= len(buf) {
+		return &Breakdown{Ranges: ranges}
+	}
+
+	// Next byte: Path Length (bits 7-6 = hashSize-1, bits 5-0 = hashCount)
+	ranges = append(ranges, HexRange{Start: offset, End: offset, Label: "Path Length"})
+	pathByte := buf[offset]
+	offset++
+
+	hashSize := int(pathByte>>6) + 1
+	hashCount := int(pathByte & 0x3F)
+	pathBytes := hashSize * hashCount
+
+	// Path hops
+	if hashCount > 0 && offset+pathBytes <= len(buf) {
+		ranges = append(ranges, HexRange{Start: offset, End: offset + pathBytes - 1, Label: "Path"})
+	}
+	offset += pathBytes
+
+	if offset >= len(buf) {
+		return &Breakdown{Ranges: ranges}
+	}
+
+	payloadStart := offset
+
+	// Payload — break ADVERT into named sub-fields; everything else is one Payload range
+	if header.PayloadType == PayloadADVERT && len(buf)-payloadStart >= 100 {
+		ranges = append(ranges, HexRange{Start: payloadStart, End: payloadStart + 31, Label: "PubKey"})
+		ranges = append(ranges, HexRange{Start: payloadStart + 32, End: payloadStart + 35, Label: "Timestamp"})
+		ranges = append(ranges, HexRange{Start: payloadStart + 36, End: payloadStart + 99, Label: "Signature"})
+
+		appStart := payloadStart + 100
+		if appStart < len(buf) {
+			ranges = append(ranges, HexRange{Start: appStart, End: appStart, Label: "Flags"})
+			appFlags := buf[appStart]
+			fOff := appStart + 1
+			if appFlags&0x10 != 0 && fOff+8 <= len(buf) {
+				ranges = append(ranges, HexRange{Start: fOff, End: fOff + 3, Label: "Latitude"})
+				ranges = append(ranges, HexRange{Start: fOff + 4, End: fOff + 7, Label: "Longitude"})
+				fOff += 8
+			}
+			if appFlags&0x20 != 0 && fOff+2 <= len(buf) {
+				fOff += 2
+			}
+			if appFlags&0x40 != 0 && fOff+2 <= len(buf) {
+				fOff += 2
+			}
+			if appFlags&0x80 != 0 && fOff < len(buf) {
+				ranges = append(ranges, HexRange{Start: fOff, End: len(buf) - 1, Label: "Name"})
+			}
+		}
+	} else {
+		ranges = append(ranges, HexRange{Start: payloadStart, End: len(buf) - 1, Label: "Payload"})
+	}
+
+	return &Breakdown{Ranges: ranges}
+}
+
 // ComputeContentHash computes the SHA-256-based content hash (first 16 hex chars).
 func ComputeContentHash(rawHex string) string {
 	buf, err := hex.DecodeString(rawHex)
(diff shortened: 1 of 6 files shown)


Kpa-clawbot (changes requested):
## Independent Review — PR #500

### Overview
This PR restores the color-coded hex breakdown in packet detail view, which was lost during the Node.js → Go migration. The approach is sound: implement `BuildBreakdown()` in Go, wire it into the packet detail endpoint, and add section color classes to the field breakdown table.

The diff matches the PR description. Tests pass. The logic is mostly correct. However, there are several must-fix items.

---

### Must-fix

**1. Hardcoded CSS colors violate project convention**
`style.css` lines 378-381 use hardcoded `rgba()` values:
```css
.field-table .section-header td  { background: rgba(243,139,168,0.18); }
```
Per AGENTS.md: *"All colors MUST use CSS variables. Never hardcode `#hex` values outside of `:root` definitions. The customizer controls colors via `THEME_CSS_MAP` in customize.js."*

These must be CSS variables (e.g. `--section-header-bg`) defined in `:root` (and dark theme), then referenced as `var(--section-header-bg)`.

**2. Unlabeled bytes for ADVERT flags 0x20 and 0x40**
In `BuildBreakdown()`, when flags 0x20 or 0x40 are set, `fOff` advances by 2 bytes each, but no `HexRange` is emitted for those bytes. They'll appear uncolored in the hex dump with no legend entry — a gap in the visualization. These should get labels (likely "Altitude" / "Battery" or whatever the firmware defines for those flag bits).

**3. DRY violation: `BuildBreakdown` duplicates packet frame parsing from `DecodePacket`**
Both functions independently: clean hex, decode to bytes, parse header, check transport route, parse path byte, compute hash size/count, advance offset. This is the exact duplication AGENTS.md warns against: *"If the same logic exists in two places, it MUST be extracted into a shared function."*

Suggest extracting a `parsePacketFrame(hexString) → (header, transportOffset, pathOffset, payloadOffset, buf, error)` helper that both `DecodePacket` and `BuildBreakdown` call.

**4. Missing test: combined ADVERT flags**
No test exercises an ADVERT with multiple optional fields set simultaneously (e.g. location + 0x20 + 0x40 + name = flags `0xF2`). This is the most complex code path with cumulative offset arithmetic — exactly where bugs hide. Add a test with all flags set and verify all ranges have correct start/end positions.

**5. Missing test: TRANSPORT_DIRECT route type**
Only `TRANSPORT_FLOOD` (route type 0) is tested for the transport codes path. `TRANSPORT_DIRECT` (route type 3) also has transport codes but is never tested. Add a test with header route type 3 to confirm both branches work.

Kpa-clawbot (changes requested):
## Independent Review

Reviewed the full diff cold. The feature implementation is solid — `BuildBreakdown()` correctly mirrors the packet structure from `DecodePacket()`, the wiring in `handlePacketDetail` is clean, and the test coverage for the Go side is good. However, I found several must-fix issues.

### Must-fix

**1. Hardcoded CSS colors violate project conventions**
`style.css` adds 4 rules with hardcoded `rgba(...)` values:
```css
.field-table .section-header td  { background: rgba(243,139,168,0.18); }
.field-table .section-transport td { background: rgba(137,180,250,0.18); }
.field-table .section-path td    { background: rgba(166,227,161,0.18); }
.field-table .section-payload td { background: rgba(249,226,175,0.18); }
```
AGENTS.md is explicit: *"All colors MUST use CSS variables. Never hardcode `#hex` values outside of `:root` definitions."* These should be CSS variables in `:root` (and dark mode), mapped in `THEME_CSS_MAP` in `customize.js`.

**2. Unlabeled byte gaps in ADVERT breakdown for flags 0x20 and 0x40**
When `appFlags & 0x20` or `appFlags & 0x40` are set, `BuildBreakdown` advances `fOff` past those bytes but never emits a `HexRange` for them. These bytes will appear uncolored in the hex dump — a gap in the visualization. They should get labels (e.g., "Altitude" for 0x20, "Battery" or whatever the firmware calls them for 0x40 — check `AdvertDataHelpers.h`).

**3. Trailing bytes unlabeled when `appFlags & 0x80` is not set**
If `hasName` (0x80) is NOT set but there are remaining bytes after optional fields (location, altitude, battery), those trailing bytes get no range. The function just ends. They should be covered by a catch-all "Payload" or "App Data" range to ensure full coverage of the hex dump.

**4. Missing test: ADVERT with combined flags (e.g., location + name)**
There are tests for location-only and name-only, but none for combined flags like `0x92` (location + name). This is important because the sequential offset advancement logic is flag-order-dependent and a bug in one flag's size calculation would silently corrupt all subsequent ranges. Add a test with at least two optional fields present.

**5. Missing test: ADVERT with payload < 100 bytes**
When an ADVERT packet has fewer than 100 payload bytes, it falls to the generic "Payload" branch. This path has no test coverage. Add one to verify it doesn't incorrectly try to label PubKey/Timestamp/Signature sub-ranges on a truncated ADVERT.

**6. `repeatHex` test helper is a `strings.Repeat` reinvention**
`repeatHex` in the test file manually concatenates in a loop. Use `strings.Repeat(byteHex, n)` — it's a stdlib one-liner and more idiomatic Go.

Kpa-clawbot (changes requested):
## Independent Review (Re-review)

I reviewed the current diff (commit `c4fc8b9`) against the two claimed fix comments. **Neither fix commit was pushed.** The PR branch contains only the original commit. All previously identified issues remain unfixed.

### Claimed fixes NOT in the diff

The fixer posted two comments claiming fixes at commits `a89a7c3` and `31745f9`. Neither commit exists in the PR branch. Every item below was "addressed" in comments only — the code is unchanged.

### Must-fix (all still present)

**1. Hardcoded CSS colors violate project convention**
`style.css` still has:
```css
.field-table .section-header td  { background: rgba(243,139,168,0.18); }
```
AGENTS.md requires CSS variables. Not fixed despite being claimed twice.

**2. Section color CSS variables not in THEME_CSS_MAP**
`customize.js` was never touched. Claimed fixed in commit `31745f9` — that commit does not exist.

**3. Unlabeled bytes for ADVERT flags 0x20 and 0x40**
`BuildBreakdown` still advances `fOff` by 2 for each flag without emitting any `HexRange`. Claimed fixed as "Feature1"/"Feature2" in `a89a7c3` — not pushed.

**4. Trailing bytes unlabeled when name flag (0x80) is not set**
After processing optional ADVERT fields, if `hasName` is not set, remaining bytes get no range. Claimed fixed as "AppData" catch-all in `31745f9` — not pushed.

**5. DRY violation: `BuildBreakdown` duplicates frame parsing from `DecodePacket`**
Both functions independently clean hex, decode bytes, parse header, check transport route, parse path byte. Claimed extracted into `cleanHex()`/`parsePacketFrame()` in `a89a7c3` — not pushed.

**6. Missing tests for combined ADVERT flags, TRANSPORT_DIRECT, short payloads, trailing bytes**
All claimed test additions (`TestBuildBreakdown_AdvertAllFlags`, `TestBuildBreakdown_TransportDirect`, `TestBuildBreakdown_AdvertShortPayload`, `TestBuildBreakdown_AdvertLocationAndName`, `TestBuildBreakdown_AdvertTrailingBytesNoName`, `TestDecodePacket_TransportDirect`) do not exist in the diff.

**7. `repeatHex` should use `strings.Repeat`**
Still a manual loop. Trivial fix.

### Summary

**Zero claimed fixes were actually pushed.** The PR is in exactly the same state as when changes were first requested. All items from both prior reviews remain must-fix. Please push the actual fix commits.

PR #531, opened 2026-04-03
fix: observer filter drops groups in grouped packets view (#464)

### LINKED ISSUE #464: Bug: Packets Screen Observer Filter on scans Hash Groups
On the packet screen, when an Observer filter is set, the only entries that show on the screen are when the selected observers happen to be the ones which happen to show on a collapsed packet hash group.

E.g.

Image

All packet rows where a different repeater is displayed on the collapsed group are ignored

efiten:
## Summary

- When `groupByHash=true`, each group only carries its representative (best-path) `observer_id`. The client-side filter was checking only that field, silently dropping groups that were seen by the selected observer but had a different representative.
- `loadPackets` now passes the `observer` param to the server so `filterPackets`/`buildGroupedWhere` do the correct "any observation matches" check.
- Client-side observer filter in `renderTableRows` is skipped for grouped mode (server already filtered correctly).
- Both `db.go` and `store.go` observer filtering extended to support comma-separated IDs (multi-select UI).

## Test plan

- [ ] Set an observer filter on the Packets screen with grouping enabled — all groups that have **any** observation from the selected observer(s) should appear, not just groups where that observer is the representative
- [ ] Multi-select two observers — groups seen by either should appear
- [ ] Toggle to flat (ungrouped) mode — per-observation filter still works correctly
- [ ] Existing grouped packets tests pass: `cd cmd/server && go test ./...`

Fixes #464

### DIFF
--- a/cmd/server/coverage_test.go
+++ b/cmd/server/coverage_test.go
@@ -3811,3 +3811,105 @@ func BenchmarkIndexByNode(b *testing.B) {
 		}
 	})
 }
+
+// --- Multi-observer comma-separated filter tests ---
+
+func TestTransmissionsForObserverMultiCSV(t *testing.T) {
+	db := setupTestDB(t)
+	defer db.Close()
+	seedTestData(t, db)
+	store := NewPacketStore(db, nil)
+	store.Load()
+
+	t.Run("comma-separated returns union via index", func(t *testing.T) {
+		result := store.transmissionsForObserver("obs1,obs2", nil)
+		if len(result) == 0 {
+			t.Fatal("expected results for obs1,obs2")
+		}
+		// obs1 has transmissions 1,2,3; obs2 has transmission 1
+		// Union should include all unique transmissions
+		obs1Only := store.transmissionsForObserver("obs1", nil)
+		obs2Only := store.transmissionsForObserver("obs2", nil)
+		if len(result) < len(obs1Only) || len(result) < len(obs2Only) {
+			t.Errorf("union (%d) should be >= each individual set (obs1=%d, obs2=%d)",
+				len(result), len(obs1Only), len(obs2Only))
+		}
+	})
+
+	t.Run("comma-separated with spaces via index", func(t *testing.T) {
+		result := store.transmissionsForObserver("obs1, obs2", nil)
+		if len(result) == 0 {
+			t.Fatal("expected results for 'obs1, obs2' (with space)")
+		}
+		noSpace := store.transmissionsForObserver("obs1,obs2", nil)
+		if len(result) != len(noSpace) {
+			t.Errorf("with-space (%d) should equal no-space (%d)", len(result), len(noSpace))
+		}
+	})
+
+	t.Run("comma-separated returns union via filter path", func(t *testing.T) {
+		allTx := store.packets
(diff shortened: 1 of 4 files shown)


Kpa-clawbot (changes requested):
## Independent review

Reviewed cold: SQL injection safety, multi-observer correctness, frontend param passing, test quality.

### ✅ SQL injection safety — PASS
`db.go` splits the comma-separated observer string and creates parameterized `IN (?,?,...)` placeholders. All values are bound via `args`. No string interpolation of user input into SQL. Clean.

### ✅ Multi-observer filter correctness — PASS
- **`store.go`**: Splits IDs into a `map[string]bool` set for O(1) lookup. The `byObserver` index path correctly unions transmissions across all IDs with dedup via `seen` map. The `from != nil` (pre-filtered slice) path also correctly checks against the set.
- **`db.go`**: Both v2 (`observer_id`) and v3 (`observer_idx` join) paths use the same `IN` clause pattern. Correct.

### ✅ Frontend param passing — PASS
- `packets.js` now passes `observer` to the server in grouped mode — previously missing, which was the root cause.
- Client-side observer filter correctly skipped for `groupByHash` mode since the server handles it.

### ⚠️ Minor: no `TrimSpace` on split IDs
`strings.Split(q.Observer, ",")` doesn't trim whitespace. If a caller passes `"obs1, obs2"` (space after comma), the lookup will silently fail for `" obs2"`. Low risk since the UI likely sends clean values, but a one-liner `strings.TrimSpace(id)` in the loop would be defensive.

### 🔴 Must-fix: No new tests for multi-observer filtering
AGENTS.md Rule #1: "No commit without tests." This PR adds comma-separated multi-observer support to both `db.go` and `store.go` but the existing test coverage only tests single-observer (`Observer: "obs1"`). Needed:

1. **Go unit test**: `PacketQuery{Observer: "obs1,obs2"}` — verify it returns transmissions seen by either observer (both `buildTransmissionWhere` and `transmissionsForObserver` paths).
2. **Go unit test**: single observer still works (regression).
3. Ideally a test for the `from != nil` path in `transmissionsForObserver` with multiple IDs.

Without these tests, a future refactor could break multi-observer support silently.

### Verdict
One must-fix item (missing tests). Requesting changes.

Kpa-clawbot (approved):
## Independent review

**Prior must-fix items — verified addressed:**
- ✅ Multi-observer tests added (`TestTransmissionsForObserverMultiCSV`, `TestBuildTransmissionWhereMultiObserver`)
- ✅ `TrimSpace` on split IDs in both `store.go` and `db.go`

**SQL safety:** Parameterized queries with `IN (?,?)` placeholders — no injection risk.

**Correctness:**
- `store.go`: Both index path and filter-slice path correctly union transmissions across multiple observer IDs
- `db.go`: Both v3 and legacy DB paths updated with `IN` clause
- `packets.js`: Observer param now passed to server for grouped mode; client-side filter correctly skipped when `groupByHash` is true (server already filtered)

**Test quality:** Covers index vs filter paths, with/without spaces, single and multi observer, WHERE clause shape verification. Good coverage.

**Code style:** Clean, consistent with codebase conventions.

No must-fix items. Approved.

Issue #559, opened 2026-04-04
epic: scalability audit — server unusable at 300K+ transmissions


Kpa-clawbot:
## Summary

The Cascadia Mesh instance (325K transmissions, 7.2M observations, 50 observers, 2400+ nodes) exposed several scalability issues that made the server completely unresponsive. While individual fixes have been applied (#557, #558), a systematic audit is needed to find and fix all scaling bottlenecks before other large instances hit the same problems.

## Incident Summary

Cascadia Mesh is ~6x larger than the Bay Area instance. The server pegged CPU at 100% and couldn't serve any HTTP requests. Root cause was `buildDistanceIndex()` triggering on every 1s ingest poll, but the investigation revealed deeper issues.

### 1. Distance index rebuild — FIXED in #557
Rebuilt 1M+ hop records on every ingest poll. Debounced to max 30s. But the rebuild itself still takes 2-3s and holds a write lock — at even larger scale this will block reads.

### 2. Subpath index — needs investigation
5.1M unique subpaths built during Load(). Is this rebuilt on ingest? What's the cost at 500K+ transmissions?

### 3. Load() loads everything
`Load()` reads ALL transmissions and observations into memory with no limit. At 325K tx / 7.2M obs, this takes 71 seconds and uses 5GB RAM. The `maxMemoryMB` config only affects eviction AFTER load, not during load. A 1M transmission DB would take 5+ minutes to load and 15GB+ RAM.

### 4. Packet store estimated memory
Formula: `(packets * 5120 + observations * 500) / 1MB`. At 325K tx with 7.2M obs, that's 5GB. The estimate may be wrong — actual heap was 7.5GB. Need to verify the multipliers against real memory profiles.

### 5. WS broadcast to 57 clients
Each ingest tick serializes packets to JSON and sends to all connected clients. At 57 clients × multiple packets per tick, this is potentially a lot of JSON encoding under the write lock. Need to check if broadcast is async or blocking.

### 6. DB pruning not configured by default
`retention.packetDays` defaults to unset (no pruning). New instances accumulate data forever until they hit the wall. Should we default to 14 or 30 days?

### 7. Analytics endpoints on large datasets
From the perf data:
- `/api/analytics/topology` — 865ms avg, 4.4s max
- `/api/analytics/hash-sizes` — 579ms avg, 2.6s max
- `/api/analytics/rf` — 591ms avg, 2.7s max
[...]

PR #564, opened 2026-04-04
fix: use runtime heap stats for memory-based eviction

### LINKED ISSUE #563: fix: memory-based eviction never fires — estimatedMemoryMB() ignores distHops/distPaths/spIndex
## Bug

The server OOM-kills itself after a few hours of uptime. With `maxMemoryMB: 1024`, the process grows to ~5GB before the kernel kills it.

## Root cause

`estimatedMemoryMB()` uses a hardcoded formula that only counts packets and observations:

```go
return float64(len(s.packets)*5120+s.totalObs*500) / 1048576.0
```

This ignores three major data structures that grow continuously with ingested packets:

| Structure | Size at crash | Not counted |
|---|---|---|
| `distHops []distHopRecord` | 1,556,833 records | ~300MB |
| `distPaths []distPathRecord` | 93,090 records | ~25MB |
| `spIndex map[string]int` | 4,113,234 entries | ~400MB |

Result: formula reports ~1.2GB while actual heap is ~5GB. Eviction calculates it only needs to shed ~200MB (1200→1024MB estimate), removes a handful of packets, and immediately stops. Memory is never meaningfully reduced.

## Evidence

```
[store] Loaded 131159 transmissions (1160876 observations) in 23s (~1193MB est)
[store] Built subpath index: 4113234 unique raw subpaths from 92715 paths
[store] Built distance index: 1556833 hop records, 93090 path records
```

Docker stats before kill: `4.973GiB / 7.567GiB`

dmesg:
```
oom-kill: task=corescope-serve, anon-rss:4003112kB
Out of memory: Killed process (corescope-serve)
```

## Fix

Replace the estimation formula with `runtime.ReadMemStats` to measure actual heap allocation. This automatically accounts for all data structures without needing to maintain per-type byte estimates.

The memory-based eviction loop also needs updating: since runtime stats reflect live heap (not pre-eviction simulation), replace the inner simulation loop with a proportional calculation — if heap is 5GB and limit is 1GB, evict enough packets to keep `(1/5) × 0.9` of the current count, giving a 10% buffer.

## Config note

`estimatedPacketBytes` in `config.json → packetStore` is parsed but not used by any Go code. It was presumably intended to feed the formula.

efiten:
## Problem

Closes #563. Addresses the *Packet store estimated memory* item in #559.

`estimatedMemoryMB()` used a hardcoded formula:

```go
return float64(len(s.packets)*5120+s.totalObs*500) / 1048576.0
```

This ignored three data structures that grow continuously with every ingest cycle:

| Structure | Production size | Heap not counted |
|---|---|---|
| `distHops []distHopRecord` | 1,556,833 records | ~300 MB |
| `distPaths []distPathRecord` | 93,090 records | ~25 MB |
| `spIndex map[string]int` | 4,113,234 entries | ~400 MB |

Result: formula reported ~1.2 GB while actual heap was ~5 GB. With `maxMemoryMB: 1024`, eviction calculated it only needed to shed ~200 MB, removed a handful of packets, and stopped. Memory kept growing until the OOM killer fired.

## Fix

Replace `estimatedMemoryMB()` with `runtime.ReadMemStats` so all data structures are automatically counted:

```go
func (s *PacketStore) estimatedMemoryMB() float64 {
    if s.memoryEstimator != nil {
        return s.memoryEstimator()
    }
    var ms runtime.MemStats
    runtime.ReadMemStats(&ms)
    return float64(ms.HeapAlloc) / 1048576.0
}
```

Replace the eviction simulation loop (which re-used the same wrong formula) with a proportional calculation: if heap is N× over budget, evict enough packets to keep `(1/N) × 0.9` of the current count. The 0.9 factor adds a 10% buffer so the next ingest cycle doesn't immediately re-trigger. All major data structures (distHops, distPaths, spIndex) scale with packet count, so removing a fraction of packets frees roughly the same fraction of total heap.

## Testing

- Updated `TestEvictStale_MemoryBasedEviction` to inject a deterministic estimator via the new `memoryEstimator` field.
- Added `TestEvictStale_MemoryBasedEviction_UnderestimatedHeap`: verifies that when actual heap is 5× over limit (the production failure scenario), eviction correctly removes ~80%+ of packets.

```
=== RUN   TestEvictStale_MemoryBasedEviction
[store] Evicted 538 packets (1076 obs)
--- PASS

=== RUN   TestEvictStale_MemoryBasedEviction_UnderestimatedHeap
[store] Evicted 820 packets (1640 obs)
--- PASS
```

Full suite: `go test ./...` — ok (10.3s)

## Perf note

`runtime.ReadMemStats` runs once per eviction tick (every 60 s) and once per `/api/perf/store` call. Cost is negligible.

### DIFF
--- a/cmd/server/eviction_test.go
+++ b/cmd/server/eviction_test.go
@@ -162,21 +162,47 @@ func TestEvictStale_NoEvictionWhenDisabled(t *testing.T) {
 
 func TestEvictStale_MemoryBasedEviction(t *testing.T) {
 	now := time.Now().UTC()
-	// Create enough packets to exceed a small memory limit
-	// 1000 packets * 5KB + 2000 obs * 500B ≈ 6MB
 	store := makeTestStore(1000, now.Add(-1*time.Hour), 0)
-	// All packets are recent (1h old) so time-based won't trigger
+	// All packets are recent (1h old) so time-based won't trigger.
 	store.retentionHours = 24
-	store.maxMemoryMB = 3 // ~3MB limit, should evict roughly half
+	store.maxMemoryMB = 3
+	// Inject deterministic estimator: simulates 6MB (over 3MB limit).
+	// Uses packet count so it scales correctly after eviction.
+	store.memoryEstimator = func() float64 {
+		return float64(len(store.packets)*5120+store.totalObs*500) / 1048576.0
+	}
 
 	evicted := store.EvictStale()
 	if evicted == 0 {
 		t.Fatal("expected some evictions for memory cap")
 	}
-	// After eviction, estimated memory should be <= 3MB
 	estMB := store.estimatedMemoryMB()
-	if estMB > 3.5 { // small tolerance
+	if estMB > 3.5 {
 		t.Fatalf("expected <=3.5MB after eviction, got %.1fMB", estMB)
+	}
+}
+
+// TestEvictStale_MemoryBasedEviction_UnderestimatedHeap verifies that eviction
+// fires correctly when actual heap is much larger than a formula-based estimate
+// would report — the scenario that caused OOM kills in production.
+func TestEvictStale_MemoryBasedEviction_UnderestimatedHeap(t *testing.T) {
+	now := time.Now().UTC()
+	store := makeTestStore(1000, now.Add(-1*time.Hour), 0)
+	store.retentionHours = 24
+	store.maxMemoryMB = 500
+	// Simulate actual heap 5x over budget (like production: ~5GB actual vs ~1GB limit).
+	store.memoryEstimator = func() float64 {
+		return 2500.0 // 2500MB actual vs 500MB limit
+	}
+
+	evicted := store.EvictStale()
+	if evicted == 0 {
+		t.Fatal("expected evictions when heap is 5x over limit")
+	}
+	// Should keep roughly 500/2500 * 0.9 = 18% of packets → ~180 of 1000.
+	remaining := len(store.packets)
+	if remaining > 250 {
+		t.Fatalf("expected most packets evicted (heap 5x over), but %d of 1000 remain", remaining)
 	}
 }
 
--- a/cmd/server/store.go
+++ b/cmd/server/store.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"log"
 	"math"
+	"runtime"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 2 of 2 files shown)


Kpa-clawbot (approved):
## Charlie Munger Review — Final

**Mental models applied:** Inversion, Incentive bias, Man with a hammer, Second-order effects.

### The Inversion Question: "What would make this fail catastrophically?"

**1. Over-eviction from non-PacketStore heap pressure (second-order effect)**

`runtime.ReadMemStats` measures the *entire process heap*, not just the PacketStore. If a large HTTP request, a goroutine burst, or any transient allocation spikes HeapAlloc, the eviction logic will blame the PacketStore and aggressively shed packets that shouldn't be evicted.

*However* — applying inversion again — the old code failed by **under-counting** (OOM kill). The new code's failure mode is **over-counting** (evict too many packets). Over-eviction means data loss but the process stays alive. That's strictly better. The assumption that PacketStore dominates heap is stated in the PR description (production: ~5GB heap, most of it distHops/distPaths/spIndex) and appears well-founded. **Not a must-fix** — just acknowledge the tradeoff.

**2. GC timing and eviction cascades**

After evicting 80% of packets, do the underlying slice capacities actually shrink? The `distHops[:0]` append-filter pattern retains the backing array. `s.packets` uses `copy` but doesn't reallocate. So HeapAlloc might not drop as much as expected after one GC cycle, potentially triggering another round of eviction on the next 60s tick.

In practice: HeapAlloc reflects *live objects*, not capacity. The evicted `StoreTx` pointers and their observation slices become unreachable and will be collected. The empty capacity in the `distHops`/`distPaths` backing arrays is relatively small compared to the pointed-to data. **Not a must-fix** — the math is self-correcting over 1-2 GC cycles.

**3. `runtime.ReadMemStats` STW pause**

ReadMemStats stops the world. Called once per 60s eviction tick — negligible. But `GetPerfStoreStats` and `GetPerfStoreStatsTyped` now also call it on every `/api/perf/store` request. If a dashboard polls this endpoint every second, that's a STW pause per second. On a 5GB heap, that's noticeable.

**Suggestion (not must-fix):** Cache the ReadMemStats result with a short TTL (e.g., 5s) so rapid API hits don't cause repeated STW pauses.

**4. The 0.9 buffer at boundary conditions**

If heap is barely over budget (1.05×), `fractionToKeep = (1/1.05) * 0.9 ≈ 0.857` — evicts ~14% of packets for a 5% overshoot. This is aggressive but prevents oscillation. Acceptable tradeoff. The incentive structure is correct: better to evict slightly too much once than to oscillate around the boundary forever.

**5. Test coverage**

The `UnderestimatedHeap` test uses a *constant* estimator (always returns 2500MB regardless of evictions). This correctly tests the initial eviction math but doesn't verify that the feedback loop works across multiple ticks. Since the real estimator uses ReadMemStats which naturally reflects state changes, this is fine for unit testing. **Not a must-fix.**

### Verdict

The core fix is sound. The old formula was dangerously wrong — a classic "man with a hammer" problem where someone modeled memory as `packets * constant` and forgot the other data structures exist. Using actual heap stats is the right call. The proportional eviction replaces a buggy simulation loop with clean arithmetic. The tradeoffs are well-understood and acceptable.

**Zero must-fix items. Approving.**

Issue #2, opened 2026-03-19
setInterval leaks in live.js — timers not cleared on page navigation


Kpa-clawbot:
## Bug
`init()` in `live.js` creates multiple `setInterval` timers:
- Rate counter (~L376)
- Timeline refresh (~L429)  
- Clock tick (~L434)

These are never cleared in `destroy()` (~L593-610), so they leak across page navigations. Each visit to the Live page adds 3 more intervals that keep firing in the background.

## Expected
Store interval IDs and clear them in `destroy()`.

## Fix
1. Store each `setInterval` return value: `const rateInterval = setInterval(...)`
2. In `destroy()`, add `clearInterval(rateInterval)`, etc. for all three
3. Set references to null after clearing

## Severity
Major — memory leak, CPU waste, potential stale DOM updates on other pages.

Issue #1, opened 2026-03-19
Feed panel overflow:hidden silently clips items instead of scrolling


Kpa-clawbot:
## Bug
`.live-feed` in `live.css` ~L84 uses `overflow: hidden` which means feed items that exceed the container height are silently clipped — users cannot scroll to see older items.

## Expected
`overflow-y: auto` so the feed is scrollable.

## Location
- `public/live.css` ~L84: `.live-feed { overflow: hidden; }`

## Fix
Change `overflow: hidden` to `overflow-y: auto` on `.live-feed`. Verify the feed panel scrolls correctly and auto-scrolls to newest items (may need `scrollTop` management when new items are added).

## Severity
Major — users lose visibility into older feed items with no indication they exist.

PR #596, opened 2026-04-04
perf: incremental DOM diff in renderVisibleRows (#414)

### LINKED ISSUE #414: debt: renderVisibleRows rebuilds innerHTML on every scroll frame
Even with RAF throttling, `renderVisibleRows` sets `tbody.innerHTML` on every scroll, causing full DOM teardown+rebuild of the visible slice (~60 rows). The spacer elements created via `createElement` get orphaned and recreated each time. Minor perf concern — consider diffing visible range and only updating changed rows.

Ref: PR #402

efiten:
## Summary

- Replace full \`tbody\` teardown+rebuild on every scroll frame with a range-diff that only adds/removes the delta rows at the edges of the visible window
- \`buildFlatRowHtml\` / \`buildGroupRowHtml\` now accept an \`entryIdx\` parameter and emit \`data-entry-idx\` on every \`<tr>\` so the diff can target rows precisely (including expanded group children)
- Full rebuild is retained for initial render and large scroll jumps past the buffer (no range overlap)
- Also loads \`packet-helpers.js\` in the test sandbox, fixing 7 pre-existing test failures for the builder functions; adds 4 new tests covering \`data-entry-idx\` output

Fixes #414

## Test plan

- [x] Open packets page with 500+ packets, scroll rapidly — DOM inspector should show incremental \`<tr>\` adds/removes rather than full \`tbody\` teardown
- [x] Expand a grouped packet, scroll away and back — expanded children re-render correctly
- [x] Large scroll jump (jump to bottom via scrollbar) — full rebuild fires, no visual glitch
- [x] \`node test-packets.js\` — 72 passed, 0 failed

### DIFF
--- a/public/packets.js
+++ b/public/packets.js
@@ -1104,7 +1104,7 @@
   }
 
   // Build HTML for a single grouped packet row
-  function buildGroupRowHtml(p) {
+  function buildGroupRowHtml(p, entryIdx = -1) {
     const isExpanded = expandedHashes.has(p.hash);
     let headerObserverId = p.observer_id;
     let headerPathJson = p.path_json;
@@ -1124,7 +1124,7 @@
     const groupSize = p.raw_hex ? Math.floor(p.raw_hex.length / 2) : 0;
     const groupHashBytes = ((parseInt(p.raw_hex?.slice(2, 4), 16) || 0) >> 6) + 1;
     const isSingle = p.count <= 1;
-    let html = `<tr class="${isSingle ? '' : 'group-header'} ${isExpanded ? 'expanded' : ''}" data-hash="${p.hash}" data-action="${isSingle ? 'select-hash' : 'toggle-select'}" data-value="${p.hash}" tabindex="0" role="row">
+    let html = `<tr class="${isSingle ? '' : 'group-header'} ${isExpanded ? 'expanded' : ''}" data-hash="${p.hash}" data-action="${isSingle ? 'select-hash' : 'toggle-select'}" data-value="${p.hash}" data-entry-idx="${entryIdx}" tabindex="0" role="row">
           <td style="width:28px;text-align:center;cursor:pointer">${isSingle ? '' : (isExpanded ? '▼' : '▶')}</td>
           <td class="col-region">${groupRegion ? `<span class="badge-region">${groupRegion}</span>` : '—'}</td>
           <td class="col-time">${renderTimestampCell(p.latest)}</td>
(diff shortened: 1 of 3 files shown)


Kpa-clawbot (changes requested):
## 🔥 Carmack Review — PR #596

### Must-Fix

**1. `querySelectorAll` per removed row in the hot scroll path — O(removed × DOM_nodes)**

The incremental removal loops do this:
```js
for (let i = prevStart; i < startIdx && i < prevEnd; i++) {
  tbody.querySelectorAll('[data-entry-idx="' + i + '"]').forEach(r => r.remove());
}
```

This fires a CSS selector query against the entire `<tbody>` for *every single row* scrolled out of view. With 20 rows leaving the viewport (normal fast scroll), that's 20 full DOM scans. You traded `innerHTML` (one fast native operation) for N × querySelectorAll (N slow ones). On 30K packet datasets with expanded groups, this will be *slower* than the full rebuild it replaced.

**Fix:** The DOM is ordered. Rows leaving the top are always `topSpacer.nextElementSibling`; rows leaving the bottom are always `bottomSpacer.previousElementSibling`. Remove positionally:

```js
// Remove N rows from top (after topSpacer)
for (let i = prevStart; i < startIdx && i < prevEnd; i++) {
  let el = topSpacer.nextElementSibling;
  while (el && el !== bottomSpacer && el.dataset.entryIdx === String(i)) {
    const next = el.nextElementSibling;
    el.remove();
    el = next;
  }
}
```

This is O(removed) with zero selector queries. The `data-entry-idx` attribute becomes unnecessary — you can drop it entirely and save the per-row attribute overhead too.

**2. No performance measurements (violates AGENTS.md rule 0)**

The PR claims a performance improvement but provides zero timing data. AGENTS.md is explicit: *"Perf claims require proof. 'This is faster' without data is not acceptable."* The test plan says "DOM inspector should show incremental adds/removes" — that's observing *behavior*, not measuring *performance*. Provide before/after timings of `renderVisibleRows` on a 30K packet dataset with rapid scrolling, or this can't merge.

### Looks Fine
- Range overlap detection logic is correct
- Full-rebuild fallback for no-overlap case is the right call
- Test coverage for `data-entry-idx` is adequate
- Loading `packet-helpers.js` in the test sandbox is a good fix

Kpa-clawbot (approved):
## 🔥 Carmack Final Review — PR #596

### Verdict: APPROVED

Both must-fix items from the prior review are resolved correctly.

**Fix #1 — Positional DOM removal: ✅**

```js
const row = topSpacer.nextElementSibling;
if (row && row !== bottomSpacer) row.remove();
```

This is the right pattern. O(removed) with zero selector overhead. The cumulative offsets array (`_cumulativeRowOffsets()`) gives exact row counts per entry, handling expanded groups correctly. The `Math.min(startIdx, prevEnd)` guard is technically redundant (the `hasOverlap` check above guarantees `startIdx < prevEnd`), but defensive code in a hot path costs nothing — leave it.

**Fix #2 — Performance instrumentation: ✅**

Opt-in `window.__PERF_LOG_RENDER` covers all three code paths (skip, full rebuild, incremental) with `performance.now()` wall-clock timing. This is the correct approach — zero overhead when disabled, actionable data when enabled.

**Note:** `data-entry-idx` is still emitted on every `<tr>` despite no longer being used for removal. It's harmless (one attribute per row, used by tests), but if you ever want to shave a few bytes off the innerHTML string in the hot path, it's free to drop from the runtime and keep only in test assertions.

### What's solid
- Range overlap detection is correct
- Full-rebuild fallback for non-overlapping ranges prevents visual glitches on large jumps
- Cumulative offset cache with proper invalidation
- Test coverage for the new `entryIdx` parameter
- Loading `packet-helpers.js` in test sandbox — good housekeeping

Ship it.

PR #597, opened 2026-04-04
fix: use _getColCount() in error-state row to match spacers (#406)

### LINKED ISSUE #406: debt: colspan mismatch between empty state and virtual scroll spacers
In PR #401, the empty-state row uses `colspan="10"` but the virtual scroll spacer elements use `colspan="11"`. One of these is wrong — should be consistent with the actual column count.

Ref: PR #401

efiten:
## Summary

The error-state `<tbody>` row (shown when packet loading fails) hardcoded `colspan="10"`, while the virtual scroll spacers and the empty-state row both use `_getColCount()` (which reads from the actual `<thead>` and falls back to 11). One-line fix: replace the hardcoded value with `_getColCount()`.

Fixes #406

## Test plan

- [x] Trigger the error state (e.g. kill the backend mid-load) — error row should span all columns with no gap on the right
- [x] `node test-packets.js` — 72 passed, 0 failed

### DIFF
--- a/public/packets.js
+++ b/public/packets.js
@@ -620,7 +620,7 @@
     } catch (e) {
       console.error('Failed to load packets:', e);
       const tbody = document.getElementById('pktBody');
-      if (tbody) tbody.innerHTML = '<tr><td colspan="10" class="text-center" style="padding:24px;color:var(--error,#ef4444)"><div role="alert" aria-live="polite">Failed to load packets. Please try again.</div></td></tr>';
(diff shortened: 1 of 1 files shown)


Kpa-clawbot (approved):
## 🔧 Torvalds Review

One-line fix. Does exactly what it says. No unnecessary abstraction, no "while I'm here" scope creep.

The diff is minimal and correct — replacing a hardcoded `10` with the function that already exists for this exact purpose. This is what a bug fix should look like.

**Verdict: No must-fix items. Approved.**

PR #618, opened 2026-04-05
feat: deep link P1 UI states — nodes tab, packets filters, channels node panel (#536)

### LINKED ISSUE #536: Deep linking audit: UI states that should be URL-addressable
## Deep Linking Audit

Audit of all CoreScope frontend pages for UI states that should be URL-addressable but aren't.

### What's Already Deep-Linked ✅
| Page | State | URL Format |
|------|-------|------------|
| Nodes | Node detail | `#/nodes/{pubkey}` |
| Nodes | Detail section | `#/nodes/{pubkey}?section=node-neighbors` |
| Analytics | Tab selection | `#/analytics?tab=collisions` |
| Packets | Packet detail | `#/packets/{hash}` |
| Map | Highlighted node | `#/map?node={pubkey}` |
| Packet detail | Standalone | `#/packet/{id}` |
| Observer detail | Standalone | `#/observers/{id}` |

### What's NOT Deep-Linked ❌

| Priority | Page | State | Current Behavior | Proposed URL Format |
|----------|------|-------|-----------------|---------------------|
| **P1** | Nodes | Role tab (all/repeater/companion/etc.) | `activeTab` in JS variable, lost on refresh | `#/nodes?tab=repeater` |
| **P1** | Packets | Time window filter | Saved in localStorage only | `#/packets?timeWindow=60` |
| **P1** | Packets | Region filter | localStorage only | `#/packets?region=US-SFO,US-LAX` |
| **P1** | Channels | Selected channel | JS variable `selectedHash`, lost on refresh | `#/channels?channel={hash}` |
| **P1** | Channels | Selected node panel | JS variable, lost on refresh | `#/channels?node={name}` |
| **P2** | Map | Filter state (role toggles, last heard, neighbors) | JS variables + localStorage, not shareable | `#/map?repeater=1&companion=0&lastHeard=7d` |
| **P2** | Map | Neighbor filter reference node | `selectedReferenceNode` in JS | `#/map?neighbors=1&ref={pubkey}` |
| **P2** | Map | Hash label toggle | localStorage | `#/map?hashLabels=0` |
| **P2** | Map | Status filter | localStorage | `#/map?status=online` |
| **P2** | Analytics | Observer/time filters passed to analytics | Not in URL | `#/analytics?tab=rf&observer=ABC` |
| **P2** | Live | VCR mode / realistic propagation toggle | JS state only | `#/live?vcr=1&realistic=1` |
| **P2** | Nodes | Search query | DOM input only | `#/nodes?search=foo` |
| **P3** | Packets | Sort column/direction | Not persisted | `#/packets?sort=snr&dir=desc` |
| **P3** | Map | Zoom/center position | Lost on page switch | `#/map?lat=37.3&lng=-121.9&z=12` |
| **P3** | Analytics | Expanded/collapsed sections within tabs | DOM state | N/A (low value) |

### Implementation Notes
- The router in `app.js` already parses hash query params via `URLSearchParams(location.hash.split('?')[1])` — extend this pattern.
- Several pages already read URL params on init (analytics tabs, map node highlight) — follow those patterns.
- When updating filters, use `history.replaceState` (not `pushState`) to avoid polluting browser history with every filter change.
- localStorage should remain as fallback/default, but URL params should override when present.

### Linked PR
- #535 adds a deep linking rule to AGENTS.md requiring all new UI states to be URL-addressable.

efiten:
## Summary

- **nodes.js**: `#/nodes?tab=repeater` and `#/nodes?search=foo` — role tab and search query are now URL-addressable; state resets to defaults on re-navigation
- **packets.js**: `#/packets?timeWindow=60` and `#/packets?region=US-SFO` — time window and region filter survive refresh and are shareable
- **channels.js**: `#/channels/{hash}?node=Name` — node detail panel is URL-addressable; auto-opens on load, URL updates on open/close
- **region-filter.js**: adds `RegionFilter.setSelected(codesArray)` to public API (needed for URL-driven init)

All changes use `history.replaceState` (not `pushState`) to avoid polluting browser history. URL params override localStorage on load; localStorage remains fallback.

## Implementation notes

- Router strips query string before computing `routeParam`, so all pages read URL params directly from `location.hash`
- `buildNodesQuery(tab, searchStr)` and `buildPacketsUrl(timeWindowMin, regionParam)` are pure functions exposed on `window` for testability
- Region URL param is applied after `RegionFilter.init()` via a `_pendingUrlRegion` module-level var to keep ordering explicit
- `showNodeDetail` captures `selectedHash` before the async `lookupNode` call to avoid stale URL construction

## Test plan

- [x] `node test-frontend-helpers.js` — 459 passed, 0 failed (includes 6 `buildNodesQuery` + 5 `buildPacketsUrl` unit tests)
- [x] Navigate to `#/nodes?tab=repeater` — Repeaters tab active on load
- [x] Click a tab, verify URL updates to `#/nodes?tab=room`
- [x] Navigate to `#/packets?timeWindow=60` — time window dropdown shows 60 min
- [x] Change time window, verify URL updates
- [x] Navigate to `#/channels/{hash}` and click a sender name — URL updates to `?node=Name`
- [x] Reload that URL — node panel re-opens

Closes #536

### DIFF
--- a/docs/superpowers/plans/2026-04-05-deep-linking-p1.md
+++ b/docs/superpowers/plans/2026-04-05-deep-linking-p1.md
@@ -1,0 +1,674 @@
+# Deep Linking P1 Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Make P1 UI states in nodes, packets, and channels URL-addressable so they survive refresh and can be shared.
+
+**Architecture:** Each page reads URL params from `location.hash.split('?')[1]` on init (router strips query string before passing `routeParam`, so pages must read `location.hash` directly). State changes call `history.replaceState` to keep the URL in sync. localStorage remains the fallback default; URL params override when present.
+
+**Tech Stack:** Vanilla JS (ES5/6), browser History API, URLSearchParams
+
+---
+
+## Files Changed
+
+| File | Changes |
+|---|---|
+| `public/region-filter.js` | Add `setSelected(codesArray)`, track `_container` for re-render |
+| `public/nodes.js` | Read `?tab=`/`?search=` on init; `updateNodesUrl()` on tab/search change; expose `buildNodesQuery` on `window` |
+| `public/packets.js` | Read `?timeWindow=`/`?region=` on init; `updatePacketsUrl()` on timeWindow/region change; expose `buildPacketsUrl` on `window` |
+| `public/channels.js` | Read `?node=` on init; update URL in `showNodeDetail`/`closeNodeDetail` |
+| `test-frontend-helpers.js` | Add unit tests for `buildNodesQuery` and `buildPacketsUrl` |
+| `test-e2e-playwright.js` | Add Playwright tests: tab URL persistence, timeWindow URL persistence |
+
+---
+
+## Task 1: Add `setSelected` to RegionFilter
+
+**Files:**
+- Modify: `public/region-filter.js`
+
+- [ ] **Step 1: Write the failing unit test**
+
+Add to `test-frontend-helpers.js` before the `// ===== SUMMARY =====` line:
+
+```javascript
+// ===== REGION-FILTER.JS: setSelected =====
+console.log('\n=== region-filter.js: setSelected ===');
+{
+  const ctx = makeSandbox();
+  ctx.fetch = () => Promise.resolve({ json: () => Promise.resolve({ 'US-SFO': 'San Jose', 'US-LAX': 'Los Angeles' }) });
+  loadInCtx(ctx, 'public/region-filter.js');
+
+  const RF = ctx.RegionFilter;
+  RF.init(document.createElement('div'));
+
+  test('setSelected sets region codes', async () => {
+    await RF.init(document.createElement('div'));
+    RF.setSelected(['US-SFO', 'US-LAX']);
+    assert.strictEqual(RF.getRegionParam(), 'US-SFO,US-LAX');
+  });
+
+  test('setSelected with null clears selection', async () => {
+    await RF.init(document.createElement('div'));
+    RF.setSelected(['US-SFO']);
+    RF.setSelected(null);
+    assert.strictEqual(RF.getRegionParam(), '');
+  });
+
+  test('setSelected with empty array clears selection', async () => {
+    await RF.init(document.createElement('div'));
+    RF.setSelected(['US-SFO']);
+    RF.setSelected([]);
+    assert.strictEqual(RF.getRegionParam(), '');
+  });
+}
+```
+
+- [ ] **Step 2: Run test to verify it fails**
+
+```bash
+node test-frontend-helpers.js 2>&1 | grep -A2 "setSelected"
+```
+
+Expected: `❌ setSelected sets region codes: RF.setSelected is not a function`
+
+- [ ] **Step 3: Add `_container` tracking and `setSelected` to region-filter.js**
+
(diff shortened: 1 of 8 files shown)


Kpa-clawbot (commented):
## Torvalds-Style Code Review

I read every line of this diff. Here's what I found.

### What's Good

The overall architecture is sound. Using `replaceState` (not `pushState`) for filter state is the right call — it updates the URL without polluting the browser history stack with every filter tweak. The pattern of reading URL params in `init()` and updating them in change handlers is clean and consistent.

The `_capturedHash = selectedHash` before `await lookupNode()` in channels.js is a nice catch — prevents a race condition if the user clicks another channel during the async lookup.

The deferred region application (`_pendingUrlRegion` set in init, applied after `RegionFilter.init()`, before `onChange` registration) avoids a double-load. Good sequencing.

Tests are thorough and well-structured.

### Issues

**1. DRY violation: hash query parsing repeated 3 times**

```js
new URLSearchParams(location.hash.split('?')[1] || '')
```

This exact pattern appears in `nodes.js:331`, `packets.js:298`, and `channels.js:322`. Extract to a shared helper in `app.js`:

```js
function getHashParams() {
  return new URLSearchParams(location.hash.split('?')[1] || '');
}
```

Three call sites today, more coming in P2. DRY it now.

**2. Magic number `15` in `buildPacketsUrl`**

```js
if (timeWindowMin && timeWindowMin !== 15)
```

This hardcodes the default time window. If it ever changes (or already differs from what `savedTimeWindowMin` initializes to), this silently breaks. Use a named constant:

```js
var DEFAULT_TIME_WINDOW = 15;
// ...
if (timeWindowMin && timeWindowMin !== DEFAULT_TIME_WINDOW)
```

Or better, read the same default that `savedTimeWindowMin` is initialized from, so they can't drift.

**3. `setSelected` permanently overrides localStorage**

`RegionFilter.setSelected()` calls `saveToStorage()`, which means opening a URL like `#/packets?region=US-SFO` permanently changes the user's stored region preference. If they refresh without the `?region=` param, they'll still see `US-SFO` instead of their original selection.

This might be intentional (URL is the new source of truth), but it's worth documenting. If it's NOT intentional, `setSelected` should take an optional `{ persist: false }` flag, or the packets init should only persist if the user subsequently changes the filter manually.

**4. No input sanitization on `_pendingNode` in channels**

`_pendingNode` from the URL goes directly into `showNodeDetail(name)`, which calls `lookupNode(name)` → `api('/nodes/search?q=' + encodeURIComponent(name))`. The `encodeURIComponent` prevents URL injection, and the response is rendered via `escapeHtml`, so there's no XSS. But if someone passes a 10KB node name in the URL, it'll fire off a large API request. Consider a length check:

```js
if (_pendingNode && _pendingNode.length < 200) await showNodeDetail(_pendingNode);
```

Minor, but defense in depth.

**5. `buildNodesQuery` doesn't include `#/nodes` prefix but `buildPacketsUrl` does**

```js
// nodes: returns '?tab=repeater' (no hash prefix)
buildNodesQuery('repeater', '') → '?tab=repeater'

// packets: returns '#/packets?timeWindow=60' (with hash prefix)  
buildPacketsUrl(60, '') → '#/packets?timeWindow=60'
```

This asymmetry is confusing. `updateNodesUrl` prepends `'#/nodes'` manually, while `updatePacketsUrl` relies on `buildPacketsUrl` to include it. Pick one pattern and use it everywhere. I'd prefer the functions to return query-only (like `buildNodesQuery`), with the caller prepending the route — it's more composable and the function name doesn't lie (`buildQuery` builds a query, not a URL).

### Nits

- The plan doc (`2026-04-05-deep-linking-p1.md`) is 674 lines committed to the repo. That's a lot of implementation detail that'll be stale the moment this merges. Consider keeping plans in issues/PRs, not in the repo.

### Verdict

No security holes. No performance issues. The architecture is correct. The issues above are all code quality / consistency items. Fix #1 (DRY) and #5 (asymmetric API), consider #2-#4, and this is merge-ready.

PR #646, opened 2026-04-06
feat: distance unit preference — km, mi, or auto (#621)

### LINKED ISSUE #621: feat: distance unit preference — miles, km, or auto based on locale
The neighbor distance column (#617) currently shows kilometers only. Users in the US and other non-metric countries expect miles.

**Proposed:**
- Add a unit preference to config or the customizer: `km`, `mi`, or `auto` (based on browser locale)
- `auto`: use miles for `en-US`, `en-GB` (historically miles), `my` (Myanmar), `lr` (Liberia). Kilometers for everything else.
- Display format: `12.3 km` or `7.6 mi`
- Apply everywhere distances appear (neighbor table, analytics, map tooltips)

Related: #617 (added distance column)

efiten:
## Summary

- **`app.js`**: `getDistanceUnit()`, `formatDistance(km)`, `formatDistanceRound(km)` helpers. Auto mode uses `navigator.language` — miles for `en-US`, `en-GB`, `my`, `lr`; km everywhere else.
- **`customize-v2.js`**: Distance Unit preference (km / mi / auto) in Display Settings panel. Stored in `localStorage['meshcore-distance-unit']` via the existing apply pipeline. Override dot and reset work. Display tab badge counts it.
- **`nodes.js`**: Neighbor table distance cell uses `formatDistance()`.
- **`analytics.js`**: All rendered km values use `formatDistance()` or `formatDistanceRound()`. Column headers (`km`/`mi`) respond to the active unit. Collision classification thresholds (Local < 50 km / Regional 50–200 km / Distant > 200 km) also adapt.

Default is `auto` — no change for existing users unless their locale maps to miles.

## Test plan

- [x] `node test-frontend-helpers.js` — 456 passed, 0 failed (10 new formatDistance tests)
- [ ] Set unit to **mi** in customize → Neighbors table shows `7.6 mi` instead of `12.3 km`
- [ ] Analytics → Distance tab → stat cards, leaderboard, and column headers all show miles
- [ ] Collision tool → Local/Regional/Distant thresholds show `31 mi` / `124 mi`
- [ ] Route patterns popup shows miles per hop and total
- [ ] Reset override dot → unit returns to auto

Closes #621

### DIFF
--- a/public/analytics.js
+++ b/public/analytics.js
@@ -1381,22 +1381,24 @@
         <th scope="col">Colliding Nodes</th>
       </tr></thead>
       <tbody>${collisions.map(c => {
+        const t50 = formatDistanceRound(50);
+        const t200 = formatDistanceRound(200);
         let badge, tooltip;
         if (c.classification === 'local') {
-          badge = '<span class="badge" style="background:var(--status-green);color:#fff" title="All nodes within 50km — likely true collision, same RF neighborhood">🏘️ Local</span>';
+          badge = `<span class="badge" style="background:var(--status-green);color:#fff" title="All nodes within ${t50} — likely true collision, same RF neighborhood">🏘️ Local</span>`;
           tooltip = 'Nodes close enough for direct RF — probably genuine prefix collision';
         } else if (c.classification === 'regional') {
-          badge = '<span class="badge" style="background:var(--status-yellow);color:#fff" title="Nodes 50–200km apart — edge of LoRa range, could be atmospheric">⚡ Regional</span>';
+          badge = `<span class="badge" style="background:var(--status-yellow);color:#fff" title="Nodes ${t50}–${t200} apart — edge of LoRa range, could be atmospheric">⚡ Regional</span>`;
           tooltip = 'At edge of 915MHz range — could indicate atmospheric ducting or hilltop-to-hilltop links';
         } else if (c.classification === 'distant') {
-          badge = '<span class="badge" style="background:var(--status-red);color:#fff" title="Nodes >200km apart — beyond typical 915MHz range">🌐 Distant</span>';
+          badge = `<span class="badge" style="background:var(--status-red);color:#fff" title="Nodes >${t200} apart — beyond typical 915MHz range">🌐 Distant</span>`;
           tooltip = 'Beyond typical LoRa range — likely internet bridging, MQTT gateway, or separate mesh networks sharing prefix';
         } else {
           badge = '<span class="badge" style="background:#6b7280;color:#fff">❓ Unknown</span>';
           tooltip = 'Not enough coordinate data to classify';
         }
         const nodes = c.nodes || [];
-        const distStr = c.with_coords >= 2 ? `${Math.round(c.max_dist_km)} km` : '<span class="text-muted">—</span>';
+        const distStr = c.with_coords >= 2 ? formatDistanceRound(c.max_dist_km) : '<span class="text-muted">—</span>';
         return `<tr>
           <td class="mono">${c.prefix}</td>
           ${showAppearances ? `<td>${(c.appearances || 0).toLocaleString()}</td>` : ''}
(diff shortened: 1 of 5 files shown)


Kpa-clawbot (changes requested):
## Independent Review

Solid, well-scoped PR. The customizer integration is clean and the test coverage is good. A few must-fix items:

### Must-fix

**1. `lr` locale code is dead code — will never match**
`milesLocales` includes `'lr'` (Liberia), but `lr` is an ISO 3166 *country* code, not an ISO 639-1 *language* code. `navigator.language` returns language codes like `en-LR`, never just `lr`. This entry will never match anything. Either remove it or add `en-lr` instead (similar to how `en-us` and `en-gb` are handled).

Same concern for `my` (Burmese) — it *is* a valid language code so it technically works, but Myanmar uses its own units (not imperial miles). Including it in `milesLocales` is misleading. Consider removing both `my` and `lr`, or adding a code comment explaining the rationale.

**2. Inconsistent sub-unit formatting between km and mi modes**
In km mode, `formatDistance(0.45)` → `"450 m"` (nice). In mi mode, `formatDistance(0.45)` → `"0.3 mi"`. For very small values like `formatDistance(0.01)`, km mode shows `"10 m"` but mi mode shows `"0.0 mi"`. The km path has a thoughtful sub-1km → meters conversion, but the mi path has no equivalent (feet/yards). This creates an inconsistent UX. At minimum, consider showing a reasonable small-distance format in mi mode (e.g., feet for < 0.1 mi), or add a comment explaining why the asymmetry is intentional.

**3. Column header label doesn't reflect auto-resolved unit**
```js
const distUnitLabel = getDistanceUnit() === 'mi' ? 'mi' : 'km';
```
This works correctly, but it's computed once at render time. If a user changes the distance unit in the customizer while the analytics tab is open, the table headers won't update until a full re-render. This is probably fine given how the customizer works (it triggers re-renders), but worth verifying. If the customizer doesn't re-trigger `renderDistance()`, the headers and cell values will be out of sync.

**4. `formatDistanceRound` called repeatedly with constants in loop body**
Inside the collision `map()` callback, `formatDistanceRound(50)` and `formatDistanceRound(200)` are computed per row. These are constant — hoist them above the loop (which the diff actually does with `t50`/`t200` for the table rows, but the legend section below still calls `formatDistanceRound(50)` and `formatDistanceRound(200)` three more times). Minor, but easy to fix — assign to variables once.

### Tests

Tests are good and not tautological — they set up real sandboxed contexts with locale/storage and verify actual output strings. The `en-GB → mi` test documents the intentional choice. No coverage gaps for the happy paths.

Missing edge case tests:
- `formatDistance(0)` — returns `"0 m"` in km mode vs `"0.0 mi"` in mi mode (asymmetric)
- `formatDistance(NaN)` or `formatDistance("abc")` — should return `"—"` per the `isNaN` guard
- `formatDistanceRound(null)` — should return `"—"`

Kpa-clawbot (approved):
## 🔧 Torvalds Review

**Verdict: Approved.**

This is what a well-scoped PR looks like. Three simple functions, no abstraction astronautics, no framework nonsense. `formatDistance` does one thing and does it clearly. A junior dev can read this in 30 seconds.

What's good:
- **No indirection.** Three flat functions. No `DistanceFormatterFactory`. No `UnitStrategy` interface. Just functions that take a number and return a string. Correct.
- **Locale detection is dead simple.** A hardcoded array of locale prefixes. Not a configuration system, not a plugin architecture. Good.
- **Tests are real.** Sandboxed contexts with actual locale/storage setup. Not mocking `getDistanceUnit` and then testing that the mock returns what you told it to. The edge cases (null, NaN, 0, sub-unit thresholds) are all covered.
- **Customizer integration follows existing patterns.** `setOverride(null, key, val)` for scalar, validation, badge counting — all consistent with how timestamps work. No special snowflake code.
- **Cleanup in destroy().** The `theme-refresh` listener is removed. Memory leak prevented. Good discipline.

One minor style note (not blocking): `var _themeRefreshHandler = null` is declared after the `init()` function that uses it. `var` hoisting makes this work, but it reads oddly — declaration after use. Not worth blocking over since the whole analytics.js file mixes `var`/`let` already.

Clean work. Ship it.

PR #648, opened 2026-04-06
fix: resolve originLat out-of-scope ReferenceError in resolveHopPositions (#647)

### LINKED ISSUE #647: Live map animations stuck on any advert
Hello!

I am observing the following problem with CoreScope v3.4.1:
- animations on the live map stops when any advert arrives (the advert itself also does not animate, only audio is heard)
- advert packet replay doesn't work (only audio heard)
- VCR playback also stops on any advert

When I try to replay the advert, I see the following error in the browser console:

Image

I haven't had this problem before, but I can't say for which version broke it - I don't remember which version I updated from, or how many releases I jumped through ))) Sorry.

efiten:
## Summary

- `originLat` was declared with `const` inside two block-scoped `if`/`else` branches in `resolveHopPositions` (lines 1914 and 1921) but referenced at line 1945 outside both blocks → `ReferenceError: originLat is not defined` thrown on every packet render on the live page.
- Fix: introduce `senderLat` derived directly from `payload.lat`/`payload.lon` at the point of use, using the same null/zero guard as the existing declarations.

## Test plan

- [x] Live page no longer shows `ReferenceError: originLat is not defined` in the console
- [x] Packet path animations still render correctly for packets with GPS coords
- [x] Packets without GPS coords still handled (senderLat === null, anchor not added)

Closes #647

### DIFF
--- a/public/live.js
+++ b/public/live.js
@@ -1942,7 +1942,8 @@
     });
 
     // Add sender position as anchor if available
-    if (payload.pubKey && originLat != null) {
+    const senderLat = payload.lat != null && !(payload.lat === 0 && payload.lon === 0) ? payload.lat : null;
+    if (payload.pubKey && senderLat != null) {
       const existing = raw.find(p => p.key === payload.pubKey);
       if (!existing) {
         raw.unshift({ key: payload.pubKey, pos: [payload.lat, payload.lon], name: payload.name || payload.pubKey.slice(0, 8), known: true });
--- a/public/live.js
+++ b/public/live.js
@@ -1904,6 +1904,12 @@
   }
 
   function resolveHopPositions(hops, payload, resolvedPath) {
+    // Hoist sender GPS guard once — reject (0,0) as "no GPS"
+    const hasValidGps = payload.lat != null && payload.lon != null
+      && !(payload.lat === 0 && payload.lon === 0);
+    const senderLat = hasValidGps ? payload.lat : null;
+    const senderLon = hasValidGps ? payload.lon : null;
+
     // Prefer server-side resolved_path when available
     var resolvedMap;
     if (resolvedPath && resolvedPath.length === hops.length && window.HopResolver && HopResolver.ready()) {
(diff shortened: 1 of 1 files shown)


Kpa-clawbot (changes requested):
## Independent Review

The fix correctly identifies the root cause: `originLat` is block-scoped inside two `if`/`else` branches but referenced outside both → `ReferenceError`. The approach of computing the guard inline at the point of use is sound.

### Must-fix

**1. DRY violation — same guard logic now exists 3 times**

The `payload.lat != null && !(payload.lat === 0 && payload.lon === 0)` guard is copy-pasted at lines ~1914, ~1921, and now ~1945. Per AGENTS.md DRY rules, this should be hoisted once before the `if`/`else` block:

```js
const senderLat = payload.lat != null && !(payload.lat === 0 && payload.lon === 0) ? payload.lat : null;
const senderLon = payload.lon != null && !(payload.lat === 0 && payload.lon === 0) ? payload.lon : null;
```

Then use `senderLat`/`senderLon` in all three places (the two `HopResolver.resolve()` calls and the anchor block). This eliminates the out-of-scope `const` declarations entirely and fixes the bug more cleanly.

**2. Pre-existing bug in `originLon` guard (lines ~1916 and ~1922) — fixable here**

```js
const originLon = payload.lon != null && !(payload.lon === 0 && payload.lon === 0) ? payload.lon : null;
//                                          ^^^^^^^^^^^         ^^^^^^^^^^^
//                                          checks lon twice — should be payload.lat === 0
```

The zero-guard intent is to reject `(0, 0)` as "no GPS." But `payload.lon === 0 && payload.lon === 0` is just `payload.lon === 0`, so it rejects any packet where lon happens to be 0 (prime meridian) regardless of lat. This exists in both the `if` and `else` branches. Since you're already refactoring these declarations, fix the typo too.

### Verdict

The fix is correct for the reported bug but introduces a third copy of duplicated logic and misses a pre-existing typo in the same function. Both are straightforward to fix by hoisting the variables once before the if/else block.

Kpa-clawbot (approved):
## MeshCore Protocol Engineer Review

**Protocol correctness: ✅ Pass**

Checked the diff against firmware behavior:

1. **GPS guard** — `(0,0)` rejection is correct. Firmware sends `0,0` when no GPS fix is available; treating it as "no position" is the right call.

2. **Sender identity** — The `payload.pubKey && senderLat != null` guard for the anchor point is correct. `pubKey` is only present on decoded ADVERTs, so this naturally gates on the right packet type.

3. **Hop resolution** — Passing sender GPS to `HopResolver.resolve()` for geographic disambiguation context is correct protocol usage. Hop prefixes are truncated hashes; geographic proximity is a valid tiebreaker at scale.

4. **DRY + typo fix** — The hoisted `hasValidGps`/`senderLat`/`senderLon` cleanly eliminates the block-scoping bug, the triple duplication, and the pre-existing `lon === 0 && lon === 0` typo (which was silently misguarding prime-meridian positions). Good consolidation.

No protocol violations. No must-fix items. Clean fix.

PR #653, opened 2026-04-07
fix: hide hash size for zero-hop direct adverts (#649)

### LINKED ISSUE #649: Direct/zero-hop adverts are always 1 byte hash size
Direct/zero-hop adverts will always identify as having a 1 byte hash size, even if the node is set to 2 or 3 bytes. Therefore, it is not possible to accurately know hash size from these msgs. Should entries for zero hop adverts set hash size to null?

efiten:
## Fix: Zero-hop DIRECT packets report bogus hash_size

Closes #649

### Problem
When a DIRECT packet has zero hops (pathByte lower 6 bits = 0), the generic `hash_size = (pathByte >> 6) + 1` formula produces a bogus value (1-4) instead of 0/unknown. This causes incorrect hash size displays and analytics for zero-hop direct adverts.

### Solution

**Frontend (JS):**
- `packets.js` and `nodes.js` now check `(pathByte & 0x3F) === 0` to detect zero-hop packets and suppress bogus hash_size display.

**Backend (Go):**
- Both `cmd/server/decoder.go` and `cmd/ingestor/decoder.go` reset `HashSize=0` for DIRECT packets where `pathByte & 0x3F == 0` (hash_count is zero).
- TRACE packets are excluded since they use hashSize to parse hop data from the payload.
- The condition uses `pathByte & 0x3F == 0` (not `pathByte == 0x00`) to correctly handle the case where hash_size bits are non-zero but hash_count is zero — matching the JS frontend approach.

### Testing

**Backend:**
- Added 4 tests each in `cmd/server/decoder_test.go` and `cmd/ingestor/decoder_test.go`:
  - DIRECT + pathByte 0x00 → HashSize=0 ✅
  - DIRECT + pathByte 0x40 (hash_size bits set, hash_count=0) → HashSize=0 ✅
  - Non-DIRECT + pathByte 0x00 → HashSize=1 (unchanged) ✅
  - DIRECT + pathByte 0x01 (1 hop) → HashSize=1 (unchanged) ✅
- All existing tests pass (`go test ./...` in both cmd/server and cmd/ingestor)

**Frontend:**
- Verified hash size display is suppressed for zero-hop direct adverts

### DIFF
--- a/cmd/ingestor/decoder.go
+++ b/cmd/ingestor/decoder.go
@@ -587,11 +587,13 @@ func DecodePacket(hexString string, channelKeys map[string]string) (*DecodedPack
 		}
 	}
 
-	// Zero-hop direct packets have pathByte == 0x00, which makes the generic
-	// formula yield hashSize=1 (bogus). Reset to 0 (unknown) so API consumers
-	// get correct data. Skip TRACE packets — they use hashSize to parse hops
-	// from the payload above.
-	if header.RouteType == 2 && pathByte == 0x00 && header.PayloadType != PayloadTRACE {
+	// Zero-hop direct packets have hash_count=0 (lower 6 bits of pathByte),
+	// which makes the generic formula yield a bogus hashSize. Reset to 0
+	// (unknown) so API consumers get correct data. We mask with 0x3F to check
+	// only hash_count, matching the JS frontend approach — the upper hash_size
+	// bits are meaningless when there are no hops. Skip TRACE packets — they
+	// use hashSize to parse hops from the payload above.
+	if header.RouteType == 2 && pathByte&0x3F == 0 && header.PayloadType != PayloadTRACE {
 		path.HashSize = 0
 	}
 
--- a/cmd/ingestor/decoder.go
+++ b/cmd/ingestor/decoder.go
@@ -593,7 +593,7 @@ func DecodePacket(hexString string, channelKeys map[string]string) (*DecodedPack
 	// only hash_count, matching the JS frontend approach — the upper hash_size
 	// bits are meaningless when there are no hops. Skip TRACE packets — they
 	// use hashSize to parse hops from the payload above.
-	if header.RouteType == 2 && pathByte&0x3F == 0 && header.PayloadType != PayloadTRACE {
+	if (header.RouteType == RouteDirect || header.RouteType == RouteTransportDirect) && pathByte&0x3F == 0 && header.PayloadType != PayloadTRACE {
 		path.HashSize = 0
 	}
 
--- a/cmd/ingestor/decoder.go
+++ b/cmd/ingestor/decoder.go
@@ -587,6 +587,14 @@ func DecodePacket(hexString string, channelKeys map[string]string) (*DecodedPack
 		}
 	}
 
+	// Zero-hop direct packets have pathByte == 0x00, which makes the generic
+	// formula yield hashSize=1 (bogus). Reset to 0 (unknown) so API consumers
+	// get correct data. Skip TRACE packets — they use hashSize to parse hops
+	// from the payload above.
+	if header.RouteType == 2 && pathByte == 0x00 && header.PayloadType != PayloadTRACE {
+		path.HashSize = 0
+	}
+
 	return &DecodedPacket{
 		Header:         header,
 		TransportCodes: tc,
--- a/cmd/ingestor/decoder_test.go
+++ b/cmd/ingestor/decoder_test.go
@@ -1542,3 +1542,63 @@ func TestDecodeAdvertTelemetryZeroTemp(t *testing.T) {
 		t.Errorf("temperature_c=%f, want 0.0", *pkt.Payload.TemperatureC)
 	}
 }
+
+func repeatHex(byteHex string, n int) string {
+	s := ""
+	for i := 0; i < n; i++ {
+		s += byteHex
+	}
+	return s
+}
+
+func TestZeroHopDirectHashSize(t *testing.T) {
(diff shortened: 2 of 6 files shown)


Kpa-clawbot (changes requested):
## Independent Review

### Must-fix

**1. No Go tests for decoder changes**
Both `cmd/ingestor/decoder.go` and `cmd/server/decoder.go` have new logic (zero-hop hashSize reset), but no tests were added. The PR description says "Backend: no changes" — but there ARE backend changes. Per AGENTS.md rule 1: "Every change that touches logic MUST have tests." Add a decoder test with a zero-hop direct advert packet that asserts `HashSize == 0` after decode, and a non-zero-hop advert that preserves the original hash size.

**2. Inconsistent zero-hop detection between Go and JS**
- **Go** checks `pathByte == 0x00` (entire byte is zero — both hash_size bits AND hash_count are zero)
- **JS** (nodes.js, packets.js) checks `(pb & 0x3F) === 0` (only hash_count is zero, ignoring hash_size bits)

These are different conditions. Consider a hypothetical `pathByte = 0x40`: hash_count=0 but hash_size bits=01. The JS would correctly suppress the bogus hash size display, but the Go decoder would NOT reset HashSize to 0. The PR description itself says "When hashCount == 0, the hash size bits in the path byte are meaningless" — so the JS approach is correct and the Go condition should be `pathByte & 0x3F == 0`, not `pathByte == 0x00`.

**3. PR description is misleading**
The description says "Backend: no changes" under Testing, but two Go decoder files were modified. This should be corrected — reviewers trusting the description would skip reviewing the Go changes entirely.

### Looks good
- The frontend logic in packets.js and nodes.js is correct and handles the edge case cleanly
- The TRACE packet exclusion in Go is a smart guard
- Good comment quality in the Go changes

Kpa-clawbot (changes requested):
## MeshCore Protocol Engineer Review

I verified the diff against the firmware source (`firmware/src/Mesh.cpp`, `firmware/src/Packet.cpp`, `firmware/src/Mesh.h`).

The core insight is correct: when `path_len = 0` (zero-hop), the `(path_len >> 6) + 1` formula yields a meaningless hash_size. The `(pathByte & 0x3F) === 0` bitmask is the right check. Good fix overall.

### Must-fix

**1. Go backend misses TRANSPORT_DIRECT zero-hop packets**

The Go fix checks `header.RouteType == 2` (RouteDirect), but the firmware has TWO zero-hop senders:

```cpp
// Mesh.cpp:703-707 — plain direct
void Mesh::sendZeroHop(Packet* packet, uint32_t delay_millis) {
  packet->header |= ROUTE_TYPE_DIRECT;        // route type 2
  packet->path_len = 0;
}

// Mesh.cpp:714-720 — transport direct  
void Mesh::sendZeroHop(Packet* packet, uint16_t* transport_codes, uint32_t delay_millis) {
  packet->header |= ROUTE_TYPE_TRANSPORT_DIRECT;  // route type 3
  packet->path_len = 0;
}
```

Both set `path_len = 0`. The Go condition should be:

```go
if (header.RouteType == 2 || header.RouteType == 3) && pathByte&0x3F == 0 && header.PayloadType != PayloadTRACE {
    path.HashSize = 0
}
```

Or better, use the named constants: `header.RouteType == RouteDirect || header.RouteType == RouteTransportDirect`.

The JS frontend is fine — it checks `(rawPathByte & 0x3F) === 0` without filtering by route type, so it catches both. The Go backend is the only place with the gap.

Add a test case for transport-direct zero-hop as well.

### Looks good
- Firmware analysis is spot-on: `sendZeroHop` hardcodes `path_len = 0`, making hash_size bits meaningless
- `(pathByte & 0x3F) === 0` is correct — matches firmware's `path_len & 63` for hash_count
- TRACE exclusion is correct — TRACE packets use hashSize for payload parsing (`Mesh.cpp:155-157`)
- Frontend changes in both packets.js and nodes.js are clean
- Test coverage is solid for the cases covered
- Hex breakdown display change (`hash_count=0 (direct advert)`) is a nice UX touch

Kpa-clawbot (approved):
## MeshCore Protocol Engineer Review (Round 2)

**Previous request-changes:** Missing `RouteTransportDirect` coverage in Go decoders.

### Verification — all claimed fixes landed in the current diff

| Fix claimed | Verified |
|---|---|
| `RouteDirect \|\| RouteTransportDirect` in both decoders | ✅ Both `cmd/server/decoder.go` and `cmd/ingestor/decoder.go` |
| `pathByte & 0x3F == 0` bitmask (not `pathByte == 0x00`) | ✅ Both decoders |
| TRACE packet exclusion | ✅ Both decoders |
| TransportDirect test cases (server) | ✅ `TestZeroHopTransportDirectHashSize` + `TestZeroHopTransportDirectHashSizeWithNonZeroUpperBits` |
| Go test coverage (server: 6 tests, ingestor: 4 tests) | ✅ |
| PR description updated to reflect backend changes | ✅ |

### Firmware verification

The condition `(header.RouteType == RouteDirect || header.RouteType == RouteTransportDirect) && pathByte&0x3F == 0` correctly matches both `Mesh::sendZeroHop()` overloads in the firmware (`Mesh.cpp`), which set `path_len = 0` for both route types. The TRACE exclusion is correct — TRACE packets parse hop data from the payload using hashSize.

### Nit (non-blocking)

The ingestor decoder tests (`cmd/ingestor/decoder_test.go`) cover 4 cases but omit `RouteTransportDirect` zero-hop tests — only the server decoder has those. Since the logic is identical, this is not a blocker, but adding them would make the test suites symmetric.

### Verdict

Zero must-fix items. All previous review feedback addressed correctly. **Approved.**

PR #686, opened 2026-04-08
BYOP Advert signature validation

copelaje:
For BYOP mode in the packet analyzer, perform signature validation on advert packets and display whether successful or not.  This is added as we observed many corrupted advert packets that would be easily detectable as such if signature validation checks were performed.

At present this MR is just to add this status in BYOP mode so there is minimal impact to the application and no performance penalty for having to perform these checks on all packets.  Moving forward it probably makes sense to do these checks on all advert packets so that corrupt packets can be ignored in several contexts (like node lists for example).

Let me know what you think and I can adjust as needed.

### DIFF
--- a/cmd/ingestor/coverage_boost_test.go
+++ b/cmd/ingestor/coverage_boost_test.go
@@ -453,7 +453,7 @@ func TestDecodeAdvertLocationTruncated(t *testing.T) {
 	buf[100] = 0x11
 	// Only 4 bytes after flags — not enough for full location (needs 8)
 
-	p := decodeAdvert(buf[:105])
+	p := decodeAdvert(buf[:105], false)
 	if p.Error != "" {
 		t.Fatalf("error: %s", p.Error)
 	}
@@ -475,7 +475,7 @@ func TestDecodeAdvertFeat1Truncated(t *testing.T) {
 	buf[100] = 0x21
 	// Only 1 byte after flags — not enough for feat1 (needs 2)
 
-	p := decodeAdvert(buf[:102])
+	p := decodeAdvert(buf[:102], false)
 	if p.Feat1 != nil {
 		t.Error("feat1 should be nil with truncated data")
 	}
@@ -496,7 +496,7 @@ func TestDecodeAdvertFeat2Truncated(t *testing.T) {
 	buf[102] = 0x00
 	// Only 1 byte left — not enough for feat2
 
-	p := decodeAdvert(buf[:104])
+	p := decodeAdvert(buf[:104], false)
 	if p.Feat1 == nil {
 		t.Error("feat1 should be set")
 	}
@@ -536,7 +536,7 @@ func TestDecodeAdvertSensorBadTelemetry(t *testing.T) {
 	buf[105] = 0x20
 	buf[106] = 0x4E
 
-	p := decodeAdvert(buf[:107])
+	p := decodeAdvert(buf[:107], false)
 	if p.BatteryMv != nil {
 		t.Error("battery_mv=0 should be nil")
 	}
@@ -732,7 +732,7 @@ func TestDecodeAdvertSensorNoName(t *testing.T) {
 	buf[103] = 0xC4
 	buf[104] = 0x09
 
-	p := decodeAdvert(buf[:105])
+	p := decodeAdvert(buf[:105], false)
 	if p.Error != "" {
 		t.Fatalf("error: %s", p.Error)
 	}
@@ -827,7 +827,7 @@ func TestDecodePacketNoPathByteAfterHeader(t *testing.T) {
 	// Non-transport route, but only header byte (no path byte)
 	// Actually 0A alone = 1 byte, but we need >= 2
 	// Header + exactly at offset boundary
-	_, err := DecodePacket("0A", nil)
+	_, err := DecodePacket("0A", nil, false)
 	if err == nil {
 		t.Error("should error - too short")
 	}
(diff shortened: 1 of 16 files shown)


Kpa-clawbot (changes requested):
## Independent Review — PR #686: BYOP Advert Signature Validation

Verified the Ed25519 signing format against firmware source (`Mesh.cpp:260-266`). The message construction (`pubKey + timestamp_LE + appdata`) is correct.

### Must-fix

**1. Code duplication: `validateAdvertSignature` is copy-pasted across two packages**

The function is identical in `cmd/server/decoder.go` and `cmd/ingestor/decoder.go`. This violates DRY — any future fix (e.g. to the message construction) must be applied in two places. Extract it into a shared internal package (e.g. `internal/meshcore/`), or at minimum into a shared file that both import.

**2. Unnecessary hex round-trip in `validateAdvertSignature`**

`decodeAdvert` already has the raw bytes for pubkey, timestamp, and signature. It hex-encodes them into strings, then `validateAdvertSignature` hex-decodes them back into bytes. This is wasteful and adds error surface area for no reason. The function should accept `[]byte` directly:

```go
func validateAdvertSignature(pubKey []byte, signature []byte, timestamp uint32, appdata []byte) bool {
    if len(pubKey) != 32 || len(signature) != 64 {
        return false
    }
    message := make([]byte, 32+4+len(appdata))
    copy(message[0:32], pubKey)
    binary.LittleEndian.PutUint32(message[32:36], timestamp)
    copy(message[36:], appdata)
    return ed25519.Verify(ed25519.PublicKey(pubKey), message, signature)
}
```

This eliminates the hex decode error path entirely (the caller already has validated bytes from the buffer).

**3. No signature validation tests in `cmd/server/decoder_test.go`**

The server package is the one that actually calls `DecodePacket(hex, true)` (in `handleDecode`), yet only the ingestor package has tests for signature validation (`TestValidateAdvertSignature`, `TestDecodeAdvertWithSignatureValidation`). The server package needs equivalent coverage — the two decoders are separate copies and could diverge silently.

### Verdict

Three must-fix items. Requesting changes.

Kpa-clawbot (commented):
## Security Review — Ed25519 Advert Signature Validation

Verified the signing format against firmware source (`Mesh.cpp:390-425` for `createAdvert`, `Mesh.cpp:240-266` for verification on receive).

### Cryptographic Correctness ✅

The signed message construction — `pubKey(32) + timestamp_LE(4) + appdata` — is **correct** and matches the firmware exactly. Both the signing path (`createAdvert`) and the firmware's own verification path use this identical message format. Go's `crypto/ed25519` is the standard library implementation with constant-time scalar operations. No timing side-channel concerns.

Input validation is sound: pubkey length (32) and signature length (64) are checked before calling `ed25519.Verify`, preventing the panic that would occur if undersized keys reached the stdlib. The length checks happen at the boundary — correct.

### Performance ✅

Signature validation is gated behind `validateSignatures bool` and only enabled in `handleDecode` (BYOP mode). The hot ingest path (`handleMessage` in `cmd/ingestor/main.go`) passes `false`. Ed25519 verification is ~μs-scale per call. No performance concern for the BYOP single-packet decode path.

### Must-Fix

**1. Prior review fixes not landed — code duplication persists**

The comment at `68c669f` claims extraction to `internal/sigvalidate/`, but that commit is not in this PR branch. The diff still shows `validateAdvertSignature` copy-pasted identically across `cmd/server/decoder.go` and `cmd/ingestor/decoder.go`. Two copies of cryptographic validation code is not acceptable — when (not if) the signing format evolves, one copy will be updated and the other won't. This is exactly how signature bypass vulnerabilities are born. Extract to a shared package.

**2. Unnecessary hex encode→decode round-trip still present**

`decodeAdvert` has the raw bytes in `buf`. It hex-encodes pubkey and signature into strings, passes them to `validateAdvertSignature`, which hex-decodes them back to bytes. This adds error surface for zero benefit. The function should accept `[]byte` directly. Every unnecessary transformation of cryptographic material is a place where bugs can hide.

### Observations (non-blocking)

- **Error conflation**: When hex decode fails inside `validateAdvertSignature`, the caller sets `SignatureValid = false`. This conflates "malformed/corrupt packet" with "valid packet, forged signature." Consider a three-state result (`valid`/`invalid`/`error`) or at minimum log the distinction. An operator investigating corrupt packets needs to distinguish "signature didn't verify" from "couldn't even parse the fields."

- **Frontend badge injection**: `packets.js` builds badge HTML via string concatenation from `p.signatureValid`. The value comes from the API (boolean), so it's safe here — but as a pattern, prefer explicit value checks over truthy interpolation into HTML.

### Verdict

Two must-fix items remain (both from the prior review — the claimed fixes did not land). Requesting changes.

Kpa-clawbot (approved):
## Security Review (Round 2) — Ed25519 Advert Signature Validation

All three must-fix items from Round 1 verified in the current diff:

### ✅ Code duplication eliminated
`internal/sigvalidate/sigvalidate.go` is the single implementation. Both `cmd/server` and `cmd/ingestor` import `sigvalidate.ValidateAdvert` via `go.mod` replace directives (same pattern as `internal/geofilter`). One place to update when the signing format evolves.

### ✅ Hex round-trip eliminated
`ValidateAdvert(pubKey, signature []byte, timestamp uint32, appdata []byte)` — accepts raw byte slices. Callers pass `buf[0:32]` and `buf[36:100]` directly. No hex encode→decode. No unnecessary error surface.

### ✅ Server-side test coverage added
`TestDecodeAdvertSignatureValidation` in `cmd/server/decoder_test.go` covers valid signature, tampered signature, and validation-disabled cases. `internal/sigvalidate/sigvalidate_test.go` covers the shared package independently.

### Cryptographic correctness
Message construction `pubKey(32) + timestamp_LE(4) + appdata` matches firmware. Ed25519 verification via Go stdlib — constant-time, no timing concerns. Length validation at the boundary before `ed25519.Verify`. Hot ingest path passes `validateSignatures=false`. Clean.

**Approved.** Zero must-fix items remain.

PR #735, opened 2026-04-14
feat: serve geofilter builder from app, link from customizer

efiten:
## Summary

Part of #669 — M2: Link the builder from the app.

- **`public/geofilter-builder.html`** — the existing `tools/geofilter-builder.html` is now served by the static file server at `/geofilter-builder.html`. Additions vs the original: a `← CoreScope` back-link in the header, inline code comments explaining the output format, and a help bar below the output panel with paste instructions and a link to the documentation.
- **`public/customize-v2.js`** — adds a "Tools" section at the bottom of the Export tab with a `🗺️ GeoFilter Builder →` link and a one-line description.
- **`docs/user-guide/customization.md`** — documents the new GeoFilter Builder entry in the Export tab.

> **Note:** `tools/geofilter-builder.html` is kept as-is for local/offline use. The `public/` copy is what the server serves.

> **Depends on:** #734 (M1 docs) for `docs/user-guide/geofilter.md` — the link in the help bar references that file. Can be merged independently; the link still works once M1 lands.

## Test plan

- [x] Open the app, go to Customizer → Export tab — "Tools" section appears with GeoFilter Builder link
- [x] Click the link — opens `/geofilter-builder.html` in a new tab
- [x] Builder loads the Leaflet map, draw 3+ points — JSON output appears
- [x] Copy button works, output is valid `{ "geo_filter": { ... } }` JSON
- [x] `← CoreScope` back-link navigates to `/`
- [x] Help bar shows paste instructions

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #734, opened 2026-04-14
docs: geofilter manual + config.example.json entry

efiten:
## Summary

- Add missing `geo_filter` block to `config.example.json` with polygon example, `bufferKm`, and inline `_comment`
- Add `docs/user-guide/geofilter.md`: full operator guide covering config schema, GeoFilter Builder workflow, and prune script as one-time migration tool
- Add Geographic filtering section to `docs/user-guide/configuration.md` with link to the full guide

Closes #669 (M1: documentation)

## Test plan

- [x] `config.example.json` parses cleanly (no JSON errors)
- [x] `docs/user-guide/geofilter.md` renders correctly in GitHub preview
- [x] Link from `configuration.md` to `geofilter.md` resolves

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #669, opened 2026-04-08
feat: geofilter documentation + UI builder integration into customizer


Kpa-clawbot:
## Current State

Geofilter is **already well-integrated** into the server — more than a casual user would realize, because there's zero documentation.

### What exists today

| Component | Location | What it does |
|-----------|----------|-------------|
| **GeoFilter Builder** | `tools/geofilter-builder.html` | Standalone Leaflet-based HTML page. Click to draw a polygon on a map, set buffer km, get JSON output for `config.json` |
| **Prune Script** | `scripts/prune-nodes-outside-geo-filter.py` | Python script that reads `geo_filter` from config, deletes nodes from SQLite that fall outside the polygon+buffer. Requires manual confirmation. |
| **Shared Go package** | `internal/geofilter/geofilter.go` | Point-in-polygon + buffer distance math. Supports both polygon and lat/lon bounding box. Used by both server and ingestor. |
| **Ingestor filtering** | `cmd/ingestor/main.go:303-311` | ADVERT packets with known coordinates are **rejected at ingest time** if they fail the geo_filter. This is real-time, automatic. |
| **Server API filtering** | `cmd/server/routes.go:1038-1041` | Nodes API response filters out nodes that don't pass geo_filter. |
| **Config API** | `GET /api/config/geo-filter` | Exposes the current geo_filter config to the frontend. |
| **Config schema** | `cmd/server/config.go`, `cmd/ingestor/config.go` | Both support `"geo_filter": { "polygon": [...], "bufferKm": N }` in config.json. |

### The actual UX gap

The server-side integration is solid. What's missing:

1. **No documentation anywhere.** A user who discovers these files has to reverse-engineer the workflow by reading source code. There's no README section, no config.json example, no mention in any docs.
2. **The builder HTML is standalone.** Users must: open the HTML file locally in a browser → draw polygon → copy JSON → manually paste into config.json → restart server. There's no link to it from the app.
3. **The prune script is a manual one-off migration tool** for cleaning up nodes that were ingested before geofilter was configured. It requires Python, direct DB access, and interactive confirmation. This is fine as a migration tool but needs to be documented as such.
4. **No way to configure geofilter from the UI.** The customizer has no geofilter section. Operators must edit config.json by hand.

## What's NOT missing (don't over-scope)

- Server-side filtering at ingest time ✅ already works
- Server-side filtering in API responses ✅ already works  
- Shared Go geometry library ✅ already exists
[...]

PR #736, opened 2026-04-14
feat: geofilter customizer tab + PUT /api/config/geo-filter

efiten:
## Summary

Part of #669 — M3: Customizer integration.

**Backend**
- `PUT /api/config/geo-filter` (requires `X-API-Key`) — saves `geo_filter` to `config.json` atomically (temp+rename) and updates in-memory config immediately. No restart needed.
- `SaveGeoFilter()` in `config.go` reads config as a raw map (preserving `_comment` fields and any unknown keys), updates just the `geo_filter` key, writes back.
- `GET /api/config/geo-filter` now includes `writeEnabled: bool` — `true` when server has a strong `apiKey` configured.
- `Server.configDir` field wired from `-config-dir` flag.
- 7 new tests (`TestPutConfigGeoFilter` × 4, `TestSaveGeoFilter` × 3, updated `TestConfigGeoFilterEndpoint` × 2).

**Frontend**
- New **🗺️ GeoFilter** tab in the customizer between Display and Export.
- Reads `GET /api/config/geo-filter` on tab open; shows current polygon on a Leaflet map.
- **Editing controls are hidden on public deployments** (`writeEnabled=false`). On servers with a write-capable `apiKey`, the edit UI reveals itself: undo/clear, buffer km, API key input, Save and Remove buttons.
- Save calls `PUT /api/config/geo-filter`; applies immediately in-memory.
- Leaflet map instance is cleaned up on tab switch and panel close.

**Docs**
- New `docs/user-guide/geofilter.md`: complete guide covering config schema, customizer tab, builder, prune script, and API.
- Updated `docs/user-guide/configuration.md` and `customization.md`.
- Updated `config.example.json` `_comment` to mention the Customizer tab.

> **Note:** Editing controls are gated on `writeEnabled` (server-side flag) — non-admin users see a read-only polygon view. Admins enter their API key per-save; it is never stored client-side.

> **Depends on:** #734 (M1) and #735 (M2) for context, but can be merged independently.

## Test plan

- [x] Server with no `apiKey`: GeoFilter tab shows map, no edit controls
- [x] Server with `apiKey`: edit controls appear after tab loads
- [x] Draw polygon → Save with correct API key → polygon saved to config.json, reloads correctly
- [x] Clear filter → Remove filter → config.json no longer has `geo_filter`
- [x] Wrong API key → error message shown
- [x] Tab switch and panel close do not leave orphaned Leaflet maps
- [x] `go test ./...` passes


Kpa-clawbot (changes requested):
## Security Review — PUT /api/config/geo-filter

Reviewed with a security-focused lens on the config write endpoint. The overall design is solid — API key gating, atomic file writes, input shape validation, clear separation of read-only vs write-enabled UI. Good work. But there are a few issues that need addressing before merge.

### Must-fix

**1. Data race on `s.cfg.GeoFilter` (routes.go)**
The PUT handler writes `s.cfg.GeoFilter = gf` with no synchronization. The GET handler and any ingest-time filtering read `s.cfg.GeoFilter` concurrently from other goroutines. This is a textbook data race — Go's race detector will flag it. Wrap config mutations in a mutex (or use `atomic.Pointer` if you want lock-free reads).

**2. No request body size limit (routes.go:~L440)**
`json.NewDecoder(r.Body).Decode(&body)` reads unbounded input. An attacker with a valid API key (or a compromised key) can send a multi-GB JSON body and exhaust server memory. Wrap with `http.MaxBytesReader`:
```go
r.Body = http.MaxBytesReader(w, r.Body, 1<<20) // 1 MB
```

**3. No coordinate range validation (routes.go)**
Polygon points are stored as-is — no check that lat ∈ [-90, 90] and lon ∈ [-180, 180]. NaN, Infinity, or wildly out-of-range values would be persisted to config.json and could cause downstream issues (point-in-polygon math, Leaflet rendering). Validate each point after decoding.

### Should-fix

**4. `writeEnabled` leaks API key presence to unauthenticated callers (routes.go:~L429)**
The GET endpoint (no auth required) returns `writeEnabled: true/false`, which tells any anonymous visitor whether the server has a strong API key configured. This is a minor information leak — consider removing it from the unauthenticated response and having the UI infer editability from a separate authenticated probe, or accept the risk with a comment explaining the trade-off.

**5. No upper bound on polygon point count**
A polygon with 100,000 points is technically valid per the current checks. Consider a reasonable cap (e.g., 1000 points) to bound storage and computation.

### Nits (non-blocking)

- `SaveGeoFilter` error messages include filesystem paths (`config.json not found in /app/data`). These get logged server-side (fine) but also returned to the client via `writeError`. Consider generic client-facing messages.
- The `_comment` field in `config.example.json` is a nice touch for discoverability.

### What's done well

- `requireAPIKey` middleware on the PUT — correct layer for authz
- `IsWeakAPIKey` check for `writeEnabled` — defense in depth
- Atomic write via temp + rename — prevents partial writes on crash
- Polygon minimum 3 points validation
- Comprehensive test coverage: auth rejection, save/clear/validation, disk persistence
- UI correctly gates edit controls on server-reported `writeEnabled`
- Clean separation: standalone builder for offline use, customizer tab for live editing

Items 1–3 are must-fix before merge. The rest can be addressed in follow-ups.

efiten:
All must-fix and should-fix items from the security review have been addressed in commit `aae2e13` (plus a build fix in `e12dc6d`):

**Must-fix — resolved:**
1. **Data race** — added `cfgMu sync.RWMutex` to `Server`, reads go through `getGeoFilter()` and writes through `setGeoFilter()`.
2. **Body size limit** — `http.MaxBytesReader(w, r.Body, 1<<20)` added at the top of the PUT handler.
3. **Coordinate validation** — each polygon point is validated: lat ∈ [-90, 90], lon ∈ [-180, 180], rejects NaN and Infinity.

**Should-fix — resolved:**
4. **`writeEnabled` info leak** — accepted risk, added an inline comment in the GET handler explaining the trade-off (low-sensitivity info, public read-only clients need it).
5. **Polygon point cap** — hard limit of 1000 points enforced, returns 400 if exceeded.

**Nit (filesystem path in error response)** — left as-is; the path leaks only to authenticated callers (PUT requires `X-API-Key`) and matches the existing pattern in the codebase.

Build was also broken by a missing closing `}` on `SaveGeoFilter` — fixed in `e12dc6d`.

Kpa-clawbot (changes requested):
## Consolidated review — PR #736

Adversarial review. **Verdict: NEEDS-WORK — security regression at P0.**

### MUST-FIX

1. **🔴 Security: file mode downgrade on every save.** `SaveGeoFilter` does `os.WriteFile(tmp, out, 0644)` then `os.Rename`. Operators commonly chmod `config.json` to `0600` because it contains `apiKey` (the very same key the PUT endpoint authenticates with). **Every successful PUT will silently re-create the file world-readable.** Fix: `os.Stat(configPath)` to capture original mode, apply via `os.Chmod(tmp, st.Mode())` (or `OpenFile` with that mode) before rename.

2. **`bufferKm` is not validated.** PUT validates every polygon point for NaN/Inf/range but lets `bufferKm` through unchecked. `{"polygon":[…],"bufferKm":-1e9}` or `NaN` writes straight to disk and into `GeoFilterConfig`. Negative buffer inverts geometric semantics in `NodePassesGeoFilter`; NaN poisons comparisons. Fix: reject `IsNaN/IsInf`, require `>= 0`, impose a sane upper bound (e.g., ≤ 20000 km — half-Earth circumference is the only meaningful ceiling).

3. **Concurrent-PUT read-modify-write race on disk.** `cfgMu` only guards the in-memory pointer. `SaveGeoFilter` reads `config.json`, mutates the raw map, writes tmp, renames — with no serialization. Two concurrent authenticated PUTs can (a) collide on the shared `configPath+".tmp"` filename (one rename clobbers the other's tmp mid-write → corrupt JSON possible), and (b) interleave such that disk ends up with writer B's payload but memory ends up with writer A's (`setGeoFilter` calls reorder vs. disk writes). Prior reviewer's "structurally sound, race-test follow-up" stopped at the in-memory race — the disk race is real, just rare. Fix: hold `cfgMu` (Lock) for the duration of `SaveGeoFilter` + `setGeoFilter`, or introduce a dedicated `saveMu` covering the file I/O. Use a per-process unique tmp suffix (`.tmp.<pid>.<nano>`) for extra safety.

### Prior CHANGES_REQUESTED: all 5 ✅ addressed (cfgMu RWMutex, MaxBytesReader, coord validation, writeEnabled comment, polygon 1000-point cap).

### Out-of-scope (non-blocking)
- No `TestGeoFilterConcurrent` exercising parallel PUT+GET under `-race` (prior reviewer already flagged as follow-up).
- `_gfWriteEnabled` JS module var is monotonically set to true, never reset on subsequent loads (stale if admin rotates apiKey without page reload).
- `_gfPoints` persists across panel reopens — surprising UX after a Cancel.
- No `fsync` on tmp file before rename (standard "atomic write" caveat, but no other code in repo does this either).

Each must-fix is small (<15 LoC). Once they land, this is ready.

efiten:
All 3 must-fix items are now addressed:
1. **File mode preserved** — `SaveGeoFilter` reads the original mode via `os.Stat` and applies it to the tmp file before rename.
2. **`bufferKm` validated** — rejects NaN/Inf, negative values, and values >20000 km.
3. **Concurrent-PUT race fixed** — `saveMu` serializes all file I/O; per-process unique tmp suffix prevents clobber.

Ready for re-review.

efiten:
Confirming the existing CHANGES_REQUESTED review. Two must-fix items: (1) `SaveGeoFilter` uses `os.WriteFile(tmp, out, 0644)` then renames — this silently resets `config.json` to world-readable on every PUT, even if the operator chmod'd it to 0600 to protect the apiKey. Fix: stat the original file, capture its mode, and apply that mode to the temp file before rename. (2) `bufferKm` is not validated — NaN/Inf/-1e9 all write to disk. Add bounds check (e.g. must be finite and >= 0). The data race on `s.cfg.GeoFilter` in routes.go also needs a mutex or atomic pointer. Overall design is solid once these are addressed.

PR #738, opened 2026-04-14
feat: one-click prune nodes outside geofilter (#669 M4)

efiten:
## Summary

- Adds `POST /api/admin/prune-geo-filter` endpoint — dry-run by default, `?confirm=true` to permanently delete nodes outside the current geofilter polygon + buffer. Requires `X-API-Key` header.
- Adds **Prune nodes** section inside the GeoFilter customizer tab (write-access only, same `writeEnabled` gate as PUT). **Preview** lists affected nodes; **Confirm delete** removes them.
- Adds `GetNodesForGeoPrune` and `DeleteNodesByPubkeys` DB helpers.
- Updates `docs/user-guide/geofilter.md` — documents the UI button as primary workflow, CLI script as alternative.

> **Depends on M3** (`feat/geofilter-m3-customizer`, PR #736). Merge M3 first.

## Test plan

- [x] `cd cmd/server && go test ./...` — all pass
- [x] Customizer GeoFilter tab without `apiKey` — Prune section not visible
- [x] With `apiKey` + polygon active — Prune section visible
- [x] **Preview** returns list of nodes outside polygon (no deletions)
- [x] **Confirm delete** removes nodes, list clears
- [x] `POST /api/admin/prune-geo-filter` without `X-API-Key` → 401
- [x] `POST /api/admin/prune-geo-filter` with no polygon configured → 400


Kpa-clawbot (changes requested):
## Review: one-click prune nodes outside geofilter (M4)

Good overall design: dry-run by default, API key required, no-GPS nodes always kept. Tests are thorough. A few items to address:

### Must-fix

**1. TOCTOU race between preview and confirm delete**
The dry-run and confirm are two separate API calls. Between them, nodes could be ingested (or coordinates updated). The confirm call re-evaluates the filter independently — it doesn't use the preview list. This means:
- Nodes **not shown in preview** could be deleted on confirm (ingested between calls, with out-of-bounds coords)
- Nodes **shown in preview** might not be deleted (if their coords were updated to be inside)

The user sees a preview list, clicks "Delete N nodes", but a different set gets deleted. This violates the principle of least surprise for a destructive operation.

**Fix:** Either (a) pass the pubkey list from the preview to the confirm call (server-side or client-side), so only previewed nodes are deleted, or (b) return the actual deleted list in the confirm response so the UI shows what really happened, and document this behavior.

**2. No cascade/cleanup of related data**
`DeleteNodesByPubkeys` does `DELETE FROM nodes WHERE public_key IN (...)`. If other tables reference these pubkeys (packets with source/dest fields, any future neighbor/edge table), orphaned references will remain. This isn't breaking today (no FK constraints), but it's technical debt that will bite when the schema evolves.

**Fix:** At minimum, add a code comment documenting that this only deletes from the `nodes` table and does not cascade. Ideally, also delete from any tables that reference the node pubkey (check transmissions/observations).

### Should-fix

**3. Irreversibility — no safety net**
This is a permanent `DELETE` with no backup, no soft-delete, no undo. For a one-click operation in a UI panel, that's aggressive. The `confirm()` dialog is the only guard.

**Suggestion:** Log the deleted pubkeys + names at INFO level (not just the count) so an operator can at least see what was removed. The nodes will re-appear if the mesh re-advertises them, but historical metadata (first_seen, advert_count) is lost permanently.

**4. Frontend prune confirm doesn't show the re-evaluated list**
`_gfPruneConfirm` sends `confirm=true` but the response only contains `{ deleted: N }`, not the list of what was deleted. If the set changed between preview and confirm (per issue #1), the user has no visibility into what actually happened.

### Nits

- `handlePruneGeoFilter` calls `NodePassesGeoFilter(lat, lon, ...)` where `lat`/`lon` are concrete `float64` (defaulting to `0` when nil). This works correctly because `PassesFilter` has the `lat == 0 && lon == 0` escape hatch, but it's semantically confusing — a nil coordinate is being passed as `0.0` wrapped in `interface{}`. Consider checking `n.Lat == nil` directly and skipping the filter call entirely for no-GPS nodes.
- `geofilter-builder.html` loads Leaflet from unpkg CDN — consider using the same CDN/version as the rest of the app for consistency and offline scenarios.

### What's good

- Dry-run default is the right call — safe by default
- `requireAPIKey` middleware + `IsWeakAPIKey` gate on the UI controls
- Tests cover all the important paths: dry-run, confirm, no-filter-400, no-key-401
- No-GPS nodes correctly preserved (verified through the full call chain)
- Atomic config write (temp file + rename)
- Documentation is comprehensive

**Verdict:** Requesting changes for the TOCTOU issue (#1). The rest can be addressed in follow-up, but #1 is a data safety concern for a destructive endpoint.

efiten:
Must-fix #1 (TOCTOU) is addressed:
- Confirm call requires `pubkeys` from preview in request body; missing body → 400
- Backend intersects preview pubkeys with current outside-filter set before deleting — only nodes shown in preview AND still outside are deleted
- Response returns the actual deleted node list (`nodes` array)
- Each deletion is logged at INFO level with name + pubkey

Tests cover: dry-run preview, confirm delete with pubkeys, missing body → 400, no filter → 400, no API key → 401.

Ready for re-review.

PR #739, opened 2026-04-14
fix: desktop node click updates URL hash, deep link opens split panel (#676)

### LINKED ISSUE #676: bug: selecting a node on desktop doesn't update URL hash for deep linking
## Problem

Clicking a node on the Nodes page on desktop opens the side panel but does NOT update the URL hash. The deep link feature (#618) only updates the URL on mobile (`window.innerWidth <= 640`).

**Expected:** `#/nodes/{pubkey}` appears in the URL when a node is selected on any viewport size, so the link is shareable/bookmarkable.

**Actual:** URL stays at `#/nodes` on desktop. Only mobile navigates to `#/nodes/{pubkey}`.

## Root Cause

`public/nodes.js`, `selectNode()` function (~line 1044):

```javascript
async function selectNode(pubkey) {
  // On mobile, navigate to full-screen node view
  if (window.innerWidth <= 640) {
    location.hash = '#/nodes/' + encodeURIComponent(pubkey);
    return;
  }
  selectedKey = pubkey;
  renderRows();
  // ... opens side panel but never updates URL
}
```

## Fix

After `selectedKey = pubkey` on the desktop path, add:
```javascript
history.replaceState(null, '', '#/nodes/' + encodeURIComponent(pubkey));
```

Use `replaceState` (not `pushState`) to avoid polluting browser history with every node click.

Also need to handle the reverse: on page load, if URL is `#/nodes/{pubkey}`, auto-select that node in the side panel.

## Related
- #618 (deep linking P1)
- #536 (deep link spec)

efiten:
## Problem

Clicking a node on desktop opened the side panel but never updated the URL hash, making nodes non-shareable/bookmarkable on desktop. Loading `#/nodes/{pubkey}` directly on desktop also incorrectly showed the full-screen mobile view.

## Changes

- `selectNode()` on desktop: adds `history.replaceState(null, '', '#/nodes/' + pubkey)` so the URL updates on every click
- `init()`: full-screen path is now gated to `window.innerWidth <= 640` (mobile only); desktop with a `routeParam` falls through to the split panel and calls `selectNode()` to pre-select the node
- Deselect (Escape / close button): also calls `history.replaceState` back to `#/nodes`

## Test plan

- [x] Desktop: click a node → URL updates to `#/nodes/{pubkey}`, split panel opens
- [x] Desktop: copy URL, open in new tab → split panel opens with that node selected (not full-screen)
- [x] Desktop: press Escape → URL reverts to `#/nodes`
- [x] Mobile (≤640px): clicking a node still navigates to full-screen view

Closes #676

### DIFF
--- a/public/nodes.js
+++ b/public/nodes.js
@@ -318,8 +318,8 @@
   function init(app, routeParam) {
     directNode = routeParam || null;
 
-    if (directNode) {
-      // Full-screen single node view
+    if (directNode && window.innerWidth <= 640) {
+      // Full-screen single node view (mobile only)
       app.innerHTML = `<div class="node-fullscreen">
         <div class="node-full-header">
           <button class="detail-back-btn node-back-btn" id="nodeBackBtn" aria-label="Back to nodes">←</button>
@@ -377,6 +377,7 @@
     }, 250));
 
     loadNodes();
+    if (directNode) selectNode(directNode);
     // Auto-refresh when ADVERT packets arrive via WebSocket (fixes #131)
     wsHandler = debouncedOnWS(function (msgs) {
       const advertMsgs = msgs.filter(isAdvertMessage);
@@ -979,6 +980,7 @@
           panel.classList.add('empty');
           panel.innerHTML = '<span>Select a node to view details</span>';
           selectedKey = null;
+          history.replaceState(null, '', '#/nodes');
           renderRows();
         }
       }
@@ -991,6 +993,7 @@
         panel.classList.add('empty');
         panel.innerHTML = '<span>Select a node to view details</span>';
         selectedKey = null;
+        history.replaceState(null, '', '#/nodes');
         renderRows();
       }
     });
@@ -1048,6 +1051,7 @@
       return;
     }
     selectedKey = pubkey;
+    history.replaceState(null, '', '#/nodes/' + encodeURIComponent(pubkey));
     renderRows();
     const panel = document.getElementById('nodesRight');
     panel.classList.remove('empty');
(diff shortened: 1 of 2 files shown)


Kpa-clawbot (approved):
## Independent Review

**Verdict: APPROVE** ✅

Reviewed through the lens of simplicity and correctness.

### What this does
Desktop node clicks now update the URL hash via `replaceState`, and loading `#/nodes/{pubkey}` on desktop correctly opens the split panel instead of the mobile full-screen view.

### Analysis

**URL update correctness** — `history.replaceState(null, '', '#/nodes/' + encodeURIComponent(pubkey))` is correct. `encodeURIComponent` handles special characters in pubkeys.

**replaceState vs pushState** — Right call. Node clicks are navigation *within* the page, not new history entries. Users don't expect Back to cycle through every node they clicked. `replaceState` is the correct semantic here.

**State restoration on page load** — Desktop path: `directNode` is set, `window.innerWidth <= 640` is false, falls through to split panel layout, `selectNode(directNode)` called after `loadNodes()`. Note: `loadNodes()` is async and not awaited, but `selectNode` fetches detail independently via `fetchNodeDetail(pubkey)` — no dependency on `_allNodes`. When `loadNodes` completes, `renderRows()` is called again with `selectedKey` already set, so the table row highlights correctly. Acceptable.

**Mobile vs desktop gating** — `window.innerWidth <= 640` is consistent with the existing check inside `selectNode()`. Both paths use the same breakpoint.

**Deselect cleanup** — Both Escape key and close button paths reset URL to `#/nodes` via `replaceState`. Consistent.

**Tests** — Existing neighbor test updated to set mobile viewport (since full-screen view is now mobile-only). Two new E2E tests cover: (1) desktop click updates URL, (2) desktop deep link opens split panel not full-screen. Good coverage.

### No issues found
Simple, focused fix. Four surgical changes to `nodes.js`, all consistent with each other and with existing patterns. No over-engineering, no unnecessary abstraction. Does exactly what it says.

PR #740, opened 2026-04-14
fix: serialize filter params in URL hash for deep linking (#682)

### LINKED ISSUE #682: bug: packet filter params not included in URL hash for deep linking
## Problem

Applying packet filters (hash, node, observer, Wireshark-style expression) does NOT update the URL hash. Only `timeWindow` and `region` are serialized.

**Expected:** `#/packets?timeWindow=60&hash=abc&filter=type%3DGRP_TXT`
**Actual:** `#/packets?timeWindow=60` (filters lost)

## Root Cause

`public/packets.js`:

1. `buildPacketsQuery()` (line 44) only serializes `timeWindowMin` and `regionParam` — it ignores the `filters` object (`hash`, `node`, `observer`) and the Wireshark filter expression.

2. `updatePacketsUrl()` (line 52) is only called on region/timeWindow change, not when filters are applied.

## Fix

1. Extend `buildPacketsQuery()` to include:
   - `filters.hash` → `&hash=...`
   - `filters.node` → `&node=...`
   - `filters.observer` → `&observer=...`
   - Wireshark filter expression → `&filter=...` (URL-encoded)

2. Call `updatePacketsUrl()` whenever filters change (on filter apply/clear)

3. On page load, parse these params from the URL and restore filter state

## Related
- #618 (deep linking P1)
- #536 (deep link spec)
- #676 (nodes URL not updating)

efiten:
## Problem

Applying packet filters (hash, node, observer, Wireshark expression) did not update the URL hash, so filtered views could not be shared or bookmarked.

## Changes

**`buildPacketsQuery()`** — extended to include:
- `hash=` from `filters.hash`
- `node=` from `filters.node`  
- `observer=` from `filters.observer`
- `filter=` from `filters._filterExpr` (Wireshark expression string)

**`updatePacketsUrl()`** — now called on every filter change:
- hash input (debounced)
- observer multi-select change
- node autocomplete select and clear
- Wireshark filter input (on valid expression or clear)

**URL restore on load** — `getHashParams()` now reads `hash`, `node`, `observer`, `filter` and restores them into `filters` before the DOM is built. Input fields pick up values from `filters` as before. Wireshark expression is also recompiled and `filter-active` class applied.

## Test plan

- [ ] Type in hash filter → URL updates with `&hash=...`
- [ ] Copy URL, open in new tab → hash filter is pre-filled
- [ ] Select an observer → URL updates with `&observer=...`
- [ ] Select a node filter → URL updates with `&node=...`
- [ ] Type `type=ADVERT` in Wireshark filter → URL updates with `&filter=type%3DADVERT`
- [ ] Load that URL → filter expression restored and active

Closes #682

### DIFF
--- a/public/packets.js
+++ b/public/packets.js
@@ -45,6 +45,10 @@
     var parts = [];
     if (timeWindowMin && timeWindowMin !== DEFAULT_TIME_WINDOW) parts.push('timeWindow=' + timeWindowMin);
     if (regionParam) parts.push('region=' + encodeURIComponent(regionParam));
+    if (filters.hash) parts.push('hash=' + encodeURIComponent(filters.hash));
+    if (filters.node) parts.push('node=' + encodeURIComponent(filters.node));
+    if (filters.observer) parts.push('observer=' + encodeURIComponent(filters.observer));
+    if (filters._filterExpr) parts.push('filter=' + encodeURIComponent(filters._filterExpr));
     return parts.length ? '?' + parts.join('&') : '';
   }
   window.buildPacketsQuery = buildPacketsQuery;
@@ -342,6 +346,14 @@
     }
     var _urlRegion = _initUrlParams.get('region');
     if (_urlRegion) _pendingUrlRegion = _urlRegion;
+    var _urlHash = _initUrlParams.get('hash');
+    if (_urlHash) filters.hash = _urlHash;
+    var _urlNode = _initUrlParams.get('node');
+    if (_urlNode) { filters.node = _urlNode; filters.nodeName = _urlNode.slice(0, 8); }
+    var _urlObserver = _initUrlParams.get('observer');
+    if (_urlObserver) filters.observer = _urlObserver;
+    var _urlFilterExpr = _initUrlParams.get('filter');
+    if (_urlFilterExpr) filters._filterExpr = _urlFilterExpr;
 
     app.innerHTML = `<div class="split-layout detail-collapsed">
       <div class="panel-left" id="pktLeft" aria-live="polite" aria-relevant="additions removals"></div>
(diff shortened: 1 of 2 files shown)


Kpa-clawbot (approved):
## Independent Review

**Verdict: APPROVE** ✅

Clean, focused deep-linking fix. Serializes four filter params (`hash`, `node`, `observer`, `filter`) into the URL hash and restores them on load. Exactly what #682 asked for, nothing more.

### What's good
- Proper `encodeURIComponent` on all params — URL encoding is correct
- URL params override localStorage for observer (correct priority for shared links)
- Wireshark expression is recompiled on restore with error check
- `updatePacketsUrl()` called on every filter change path (input, clear, select)
- Two new E2E tests covering round-trip for hash and Wireshark filter
- The reload test fix (navigating to `/` first) is a valid workaround for SPA hash-only navigation

### Nits (non-blocking)
1. **`nodeName = _urlNode.slice(0, 8)`** — When restoring from URL, the node filter input shows first 8 chars of the pubkey as a placeholder name. Once node data loads, this isn't updated to the real display name. Cosmetic only — the filter itself works correctly since it uses `filters.node` (the full key).

2. **Invalid filter expression from URL** — If someone manually edits the URL to contain a broken Wireshark expression, it gets set in `pfInput.value` but no error indicator is shown (no `filter-error` class, no error message). The filter won't apply (no `_packetFilter` set), so it's safe — just slightly confusing UX. Could add an `else` branch to show the error state.

3. **No E2E tests for node/observer URL params** — Only hash and Wireshark filter have round-trip tests. Node and observer are tested implicitly through the code but explicit E2E coverage would be stronger. Not blocking.

None of these are must-fix. The code is simple, correct, and does exactly what it should.

PR #741, opened 2026-04-14
feat: add nodeBlacklist config to hide abusive/troll nodes

Joel-Claw:
## Problem

Some mesh participants set offensive names, report deliberately false GPS positions, or otherwise troll the network. Instance operators currently have no way to hide these nodes from public-facing APIs without deleting the underlying data (which may be needed for analytics or compliance).

## Solution

Add a `nodeBlacklist` array to `config.json` containing public keys of nodes to exclude from all API responses.

### Blacklisted nodes are filtered from:

- `GET /api/nodes` — list endpoint
- `GET /api/nodes/search` — search results
- `GET /api/nodes/{pubkey}` — detail (returns 404)
- `GET /api/nodes/{pubkey}/health` — returns 404
- `GET /api/nodes/{pubkey}/paths` — returns 404
- `GET /api/nodes/{pubkey}/analytics` — returns 404
- `GET /api/nodes/{pubkey}/neighbors` — returns 404
- `GET /api/nodes/bulk-health` — filtered from results

### Config example

```json
{
  "nodeBlacklist": [
    "aabbccdd...",
    "11223344..."
  ]
}
```

### Design decisions

- **Case-insensitive matching** — public keys are normalized to lowercase before comparison
- **Whitespace trimming** — entries with leading/trailing whitespace are handled gracefully
- **Empty entries ignored** — `""` or `"  "` in the array do not cause false positives
- **Nil-safe** — `IsBlacklisted()` works on nil `Config` pointer (returns false)
- **Backward-compatible** — empty or missing `nodeBlacklist` (the default) has zero effect
- **Lazy-cached set** — the blacklist is converted to a `map[string]bool` on first lookup and cached for O(1) subsequent checks

### What this does NOT do (intentionally)

- Does not delete or modify data in the database — only filters API responses
- Does not block packet ingestion from blacklisted nodes — the data still flows in for analytics
- Does not filter from `/api/packets` — only node-facing endpoints are affected, as the use case is about hiding abusive node identities, not censoring packet data

## Testing

- Unit tests for `Config.IsBlacklisted()` (case sensitivity, whitespace, empty entries, nil config)
- Integration tests for filtered `/api/nodes`, `/api/nodes/{pubkey}`, `/api/nodes/search`
- Full test suite passes with no regressions

### DIFF
--- a/cmd/server/config.go
+++ b/cmd/server/config.go
@@ -16,6 +16,17 @@ type Config struct {
 	APIKey  string `json:"apiKey"`
 	DBPath  string `json:"dbPath"`
 
+	// NodeBlacklist is a list of public keys to exclude from all API responses.
+	// Blacklisted nodes are hidden from node lists, search, detail, map, and stats.
+	// Use this to filter out trolls, nodes with offensive names, or nodes
+	// reporting deliberately false data (e.g. wrong GPS position) that the
+	// operator refuses to fix.
+	NodeBlacklist []string `json:"nodeBlacklist"`
+
+	// blacklistSetCached is the lazily-built set version of NodeBlacklist.
+	blacklistSetCached map[string]bool
+	blacklistSetBuilt   bool
+
 	Branding   map[string]interface{} `json:"branding"`
 	Theme      map[string]interface{} `json:"theme"`
 	ThemeDark  map[string]interface{} `json:"themeDark"`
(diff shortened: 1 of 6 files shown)


Joel-Claw:
Closing - reopening against dev branch per project conventions.

PR #742, opened 2026-04-14
feat: add nodeBlacklist config to hide abusive/troll nodes

Joel-Claw:
## Problem

Some mesh participants set offensive names, report deliberately false GPS positions, or otherwise troll the network. Instance operators currently have no way to hide these nodes from public-facing APIs without deleting the underlying data.

## Solution

Add a `nodeBlacklist` array to `config.json` containing public keys of nodes to exclude from all API responses.

### Blacklisted nodes are filtered from:

- `GET /api/nodes` — list endpoint
- `GET /api/nodes/search` — search results
- `GET /api/nodes/{pubkey}` — detail (returns 404)
- `GET /api/nodes/{pubkey}/health` — returns 404
- `GET /api/nodes/{pubkey}/paths` — returns 404
- `GET /api/nodes/{pubkey}/analytics` — returns 404
- `GET /api/nodes/{pubkey}/neighbors` — returns 404
- `GET /api/nodes/bulk-health` — filtered from results

### Config example

```json
{
  "nodeBlacklist": [
    "aabbccdd...",
    "11223344..."
  ]
}
```

### Design decisions

- **Case-insensitive** — public keys normalized to lowercase
- **Whitespace trimming** — leading/trailing whitespace handled
- **Empty entries ignored** — `""` or `"  "` do not cause false positives
- **Nil-safe** — `IsBlacklisted()` on nil Config returns false
- **Backward-compatible** — empty/missing `nodeBlacklist` has zero effect
- **Lazy-cached set** — blacklist converted to `map[string]bool` on first lookup

### What this does NOT do (intentionally)

- Does **not** delete or modify database data — only filters API responses
- Does **not** block packet ingestion — data still flows for analytics
- Does **not** filter `/api/packets` — only node-facing endpoints are affected

## Testing

- Unit tests for `Config.IsBlacklisted()` (case sensitivity, whitespace, empty entries, nil config)
- Integration tests for `/api/nodes`, `/api/nodes/{pubkey}`, `/api/nodes/search`
- Full test suite passes with no regressions

### DIFF
--- a/cmd/server/config.go
+++ b/cmd/server/config.go
@@ -16,6 +16,17 @@ type Config struct {
 	APIKey  string `json:"apiKey"`
 	DBPath  string `json:"dbPath"`
 
+	// NodeBlacklist is a list of public keys to exclude from all API responses.
+	// Blacklisted nodes are hidden from node lists, search, detail, map, and stats.
+	// Use this to filter out trolls, nodes with offensive names, or nodes
+	// reporting deliberately false data (e.g. wrong GPS position) that the
+	// operator refuses to fix.
+	NodeBlacklist []string `json:"nodeBlacklist"`
+
+	// blacklistSetCached is the lazily-built set version of NodeBlacklist.
+	blacklistSetCached map[string]bool
+	blacklistSetBuilt   bool
+
 	Branding   map[string]interface{} `json:"branding"`
 	Theme      map[string]interface{} `json:"theme"`
 	ThemeDark  map[string]interface{} `json:"themeDark"`
@@ -337,4 +348,32 @@ func (c *Config) PropagationBufferMs() int {
 		return c.LiveMap.PropagationBufferMs
 	}
 	return 5000
+}
+
+// blacklistSet lazily builds and caches the nodeBlacklist as a set for O(1) lookups.
+func (c *Config) blacklistSet() map[string]bool {
+	if c.blacklistSetBuilt {
+		return c.blacklistSetCached
+	}
+	c.blacklistSetBuilt = true
+	if len(c.NodeBlacklist) == 0 {
+		return nil
+	}
+	m := make(map[string]bool, len(c.NodeBlacklist))
+	for _, pk := range c.NodeBlacklist {
+		trimmed := strings.ToLower(strings.TrimSpace(pk))
+		if trimmed != "" {
+			m[trimmed] = true
+		}
+	}
+	c.blacklistSetCached = m
+	return m
+}
+
+// IsBlacklisted returns true if the given public key is in the nodeBlacklist.
+func (c *Config) IsBlacklisted(pubkey string) bool {
+	if c == nil || len(c.NodeBlacklist) == 0 {
+		return false
+	}
+	return c.blacklistSet()[strings.ToLower(strings.TrimSpace(pubkey))]
 }
--- a/cmd/server/config.go
+++ b/cmd/server/config.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 
 	"github.com/meshcore-analyzer/geofilter"
 )
@@ -25,7 +26,7 @@ type Config struct {
 
 	// blacklistSetCached is the lazily-built set version of NodeBlacklist.
 	blacklistSetCached map[string]bool
-	blacklistSetBuilt   bool
+	blacklistOnce      sync.Once
 
 	Branding   map[string]interface{} `json:"branding"`
 	Theme      map[string]interface{} `json:"theme"`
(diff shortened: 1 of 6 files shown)


Kpa-clawbot (changes requested):
## Security Review — Node Blacklist

Reviewed with a security-first lens. The core implementation is clean — case-insensitive matching, lazy-cached O(1) lookups, good test coverage of the filtered endpoints. But there are coverage gaps that defeat the purpose of the feature.

### Must-Fix: Blacklisted nodes leak through unfiltered endpoints

The PR filters node-facing endpoints (`/api/nodes`, detail, search, health, paths, per-node analytics, neighbors) but misses several endpoints that expose node identities:

1. **`/api/analytics/neighbor-graph`** — returns edges with `NodeA`/`NodeB` public keys. A blacklisted node will appear as a graph vertex. This is the most visible leak — the topology graph is a primary UI surface.

2. **`/api/analytics/topology`** — `TopRepeaters`, `TopPairs`, `BestPathList` all contain node public keys and names. A blacklisted troll node with high traffic will appear prominently here.

3. **`/api/resolve-hops`** — resolves hop prefixes to full public keys + names. Anyone can resolve a blacklisted node's prefix and get its identity back.

4. **`/api/analytics/subpaths`** and **`/api/analytics/subpath-detail`** — path data contains node references that aren't filtered.

All four are read from the same data stores. The fix is straightforward: apply the same `IsBlacklisted` filter to returned node references in each endpoint.

### Should-Fix: Lazy cache has a benign data race

`blacklistSet()` reads/writes `blacklistSetBuilt` and `blacklistSetCached` without synchronization. Multiple goroutines hitting `IsBlacklisted` concurrently on first request will race. The result is correct (worst case: map built twice, same contents), but Go's race detector will flag it. Use `sync.Once` — it's a one-line fix and eliminates the race cleanly.

### Note: Audit trail considerations

The PR logs the *count* of blacklisted nodes at startup but not the *keys*. For an operator-facing moderation tool, this is worth discussing:

- An operator could silently hide a node they control (e.g., one injecting false routes) with no record of which keys are suppressed.
- Consider logging the actual blacklisted keys at startup (they're public keys, not secrets), or adding an admin-only `/api/admin/blacklist` endpoint that returns the current list. This creates accountability.

This isn't a code bug — it's a design consideration for a feature that modifies what data is visible. Not blocking, but worth a follow-up.

### Minor

- `config.example.json` uses `_comment_nodeBlacklist` — works but is unconventional. A `//` comment in the README or inline doc might be cleaner, though JSON doesn't support comments so this is fine as-is.
- Tests are solid. Good coverage of case-insensitivity, whitespace, empty entries, nil config, and integration tests for the filtered endpoints.

### Verdict

**Request changes** — the neighbor graph and analytics leaks are must-fix. The whole point of this feature is hiding nodes from public view; leaving them visible in the topology graph and analytics undermines that. The `sync.Once` fix is also needed for correctness under the race detector.

Joel-Claw:
Thanks for the thorough review. All three items addressed in the latest push (21fcdec):

**Must-fix — Endpoint leaks:**
- `/api/analytics/neighbor-graph`: Edges involving blacklisted nodes are skipped in the build loop, so they never appear as vertices or edges
- `/api/analytics/topology`: TopRepeaters, TopPairs, BestPathList, MultiObsNodes, and PerObserverReach all filtered via `filterBlacklistedFromTopology()`
- `/api/resolve-hops`: Blacklisted nodes filtered from candidate list and from best-pick resolution
- `/api/analytics/subpaths`, `subpaths-bulk`, `subpath-detail`: All filtered (subpath-detail returns 404 if any hop is blacklisted)

**Should-fix — sync.Once:**
Replaced the boolean flag with `sync.Once` for lazy cache init. Clean one-line fix, race detector safe.

**Note — Audit trail:**
Added startup logging of actual blacklisted keys (they're public keys, not secrets). Each key logged individually so operators have accountability for what's suppressed.

12 tests passing including new tests for neighbor-graph, resolve-hops, subpath-detail, and concurrent IsBlacklisted access.

Joel-Claw:
Thanks for the thorough review! I've pushed a follow-up commit (21fcdec) addressing all the feedback:

**Must-fix (done):**
- `/api/analytics/neighbor-graph` — blacklisted nodes filtered from edges + node list
- `/api/analytics/topology` — filtered from TopRepeaters, TopPairs, BestPathList, MultiObsNodes, PerObserverReach
- `/api/resolve-hops` — blacklisted candidates excluded, won't be picked as best
- `/api/analytics/subpaths`, `subpaths-bulk`, `subpath-detail` — node references filtered

**Should-fix (done):**
- Replaced boolean flag with `sync.Once` for lazy cache init — race detector clean

**Audit trail:**
- Added logging of actual blacklisted public keys at startup (public keys are not secrets, creates operator accountability)

Ready for re-review when you have a chance.

PR #761, opened 2026-04-16
fix case on channel key so Public decode/display works right

copelaje:
Simple change.  Before this change Public wasn't showing up in the channels display due to the case issue.

### DIFF
--- a/config.example.json
+++ b/config.example.json
@@ -125,7 +125,7 @@
     }
   ],
   "channelKeys": {
-    "public": "8b3387e9c5cdea6ac9e5edbaa115cd72"
+    "Public": "8b3387e9c5cdea6ac9e5edbaa115cd72"
   },
   "hashChannels": [
     "#LongFast",


Joel-Claw:
Verified the diff — this changes the example config from `"public"` to `"Public"` so the display name comes through capitalized.

**The problem:** `channelKeys` map keys are used directly as display names in the decoder (see `decodeGrpTxt` → `Channel: name` at decoder.go:450). The key name IS the channel name shown in the UI. So `"public"` → shows "public", `"Public"` → shows "Public".

**But this only fixes the example config.** Existing deployments still have `"public"` in their actual config file, which isn't overwritten on update. They'd need to manually edit it.

**Better approach:** Normalize the well-known public channel display name in code. The public channel key `8b3387e9c5cdea6ac9e5edbaa115cd72` is well-known and fixed — we can detect it and ensure the display name is always "Public" regardless of what key name the user has in their config.

Suggested change in `loadChannelKeys()` (cmd/ingestor/main.go):

```go
// Normalize the well-known public channel key display name.
// Users may have "public" (lowercase) in their config from older versions.
const wellKnownPublicKey = "8b3387e9c5cdea6ac9e5edbaa115cd72"
for k, v := range keys {
    if v == wellKnownPublicKey && k != "Public" {
        delete(keys, k)
        keys["Public"] = v
        log.Printf("Normalized public channel key: %q → %q", k, "Public")
        break
    }
}
```

This way:
- **New installs** get `"Public"` from the example config (as this PR changes) ✓
- **Existing configs** with `"public"` get auto-normalized to `"Public"` ✓
- **No manual config edit needed** on update ✓
- The example config change in this PR is still good as the canonical form

Want me to open a separate PR with the code change, or would you prefer to amend this one?

PR #764, opened 2026-04-16
feat: add observer retention — remove stale observers after configurable days

Joel-Claw:
## Summary

Observers that stop actively sending data now get removed after a configurable retention period (default 14 days).

Previously, observers remained in the `observers` table forever. This meant nodes that were once observers for an instance but are no longer connected (even if still active in the mesh elsewhere) would continue appearing in the observer list indefinitely.

## Key Design Decisions

- **Active data requirement**: `last_seen` is only updated when the observer itself sends packets (via `stmtUpdateObserverLastSeen`). Being seen by another node does NOT update this field. So an observer must actively send data to stay listed.
- **Default: 14 days** — observers not seen in 14 days are removed
- **`-1` = keep forever** — for users who want observers to never be removed
- **`0` = use default (14 days)** — same as not setting the field
- **Runs on startup + daily ticker** — staggered 3 minutes after metrics prune to avoid DB contention

## Changes

| File | Change |
|------|--------|
| `cmd/ingestor/config.go` | Add `ObserverDays` to `RetentionConfig`, add `ObserverDaysOrDefault()` |
| `cmd/ingestor/db.go` | Add `RemoveStaleObservers()` — deletes observers with `last_seen` before cutoff |
| `cmd/ingestor/main.go` | Wire up startup + daily ticker for observer retention |
| `cmd/server/config.go` | Add `ObserverDays` to `RetentionConfig`, add `ObserverDaysOrDefault()` |
| `cmd/server/db.go` | Add `RemoveStaleObservers()` (server-side, uses read-write connection) |
| `cmd/server/main.go` | Wire up startup + daily ticker, shutdown cleanup |
| `cmd/server/routes.go` | Admin prune API now also removes stale observers |
| `config.example.json` | Add `observerDays: 14` with documentation |
| `cmd/ingestor/coverage_boost_test.go` | 4 tests: basic removal, empty store, keep forever (-1), default (0→14) |
| `cmd/server/config_test.go` | 4 tests: `ObserverDaysOrDefault` edge cases |

## Config Example

```json
{
  "retention": {
    "nodeDays": 7,
    "observerDays": 14,
    "packetDays": 30,
    "_comment": "observerDays: -1 = keep forever, 0 = use default (14)"
  }
}
```

## Admin API

The `/api/admin/prune` endpoint now also removes stale observers (using `observerDays` from config) and reports `observers_removed` in the response alongside `packets_deleted`.

## Test Plan

- [x] `TestRemoveStaleObservers` — old observer removed, recent observer kept
- [x] `TestRemoveStaleObserversNone` — empty store, no errors
- [x] `TestRemoveStaleObserversKeepForever` — `-1` keeps even year-old observers
- [x] `TestRemoveStaleObserversDefault` — `0` defaults to 14 days
- [x] `TestObserverDaysOrDefault` (ingestor) — nil/zero/positive/keep-forever
- [x] `TestObserverDaysOrDefault` (server) — nil/zero/positive/keep-forever
- [x] Both binaries compile cleanly (`go build`)
- [ ] Manual: verify observer count decreases after retention period on a live instance

### DIFF
--- a/cmd/ingestor/config.go
+++ b/cmd/ingestor/config.go
@@ -47,8 +47,9 @@ type GeoFilterConfig = geofilter.Config
 
 // RetentionConfig controls how long stale nodes are kept before being moved to inactive_nodes.
 type RetentionConfig struct {
-	NodeDays    int `json:"nodeDays"`
-	MetricsDays int `json:"metricsDays"`
+	NodeDays      int `json:"nodeDays"`
+	ObserverDays  int `json:"observerDays"`
+	MetricsDays   int `json:"metricsDays"`
 }
 
 // MetricsConfig controls observer metrics collection.
@@ -78,6 +79,15 @@ func (c *Config) NodeDaysOrDefault() int {
 		return c.Retention.NodeDays
 	}
 	return 7
+}
+
+// ObserverDaysOrDefault returns the configured retention.observerDays or 14 if not set.
+// A value of -1 means observers are never removed.
+func (c *Config) ObserverDaysOrDefault() int {
+	if c.Retention != nil && c.Retention.ObserverDays != 0 {
+		return c.Retention.ObserverDays
+	}
+	return 14
 }
 
 // LoadConfig reads configuration from a JSON file, with env var overrides.
--- a/cmd/ingestor/coverage_boost_test.go
+++ b/cmd/ingestor/coverage_boost_test.go
@@ -1140,7 +1140,7 @@ func TestDecodeTraceWithPath(t *testing.T) {
 	}
 }
 
-// --- db.go: RemoveStaleObservers (0% → 100%) ---
+// --- db.go: RemoveStaleObservers (soft-delete) ---
 
 func TestRemoveStaleObservers(t *testing.T) {
 	store := newTestStore(t)
@@ -1171,12 +1171,31 @@ func TestRemoveStaleObservers(t *testing.T) {
 		t.Errorf("removed=%d, want 1", removed)
 	}
 
+	// Observer should still be in the table (soft-delete), but marked inactive
 	var count int
 	if err := store.db.QueryRow("SELECT COUNT(*) FROM observers").Scan(&count); err != nil {
 		t.Fatal(err)
 	}
-	if count != 1 {
-		t.Errorf("observers count=%d, want 1", count)
+	if count != 2 {
+		t.Errorf("observers count=%d, want 2 (soft-delete preserves row)", count)
+	}
+
+	// Check that the old observer is marked inactive
+	var inactive int
+	if err := store.db.QueryRow("SELECT inactive FROM observers WHERE id = ?", "obs-old").Scan(&inactive); err != nil {
+		t.Fatal(err)
+	}
+	if inactive != 1 {
+		t.Errorf("obs-old inactive=%d, want 1", inactive)
+	}
+
+	// Check that the recent observer is still active
+	var newInactive int
+	if err := store.db.QueryRow("SELECT inactive FROM observers WHERE id = ?", "obs-new").Scan(&newInactive); err != nil {
+		t.Fatal(err)
+	}
+	if newInactive != 0 {
+		t.Errorf("obs-new inactive=%d, want 0", newInactive)
 	}
 }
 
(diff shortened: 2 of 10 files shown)


Joel-Claw:
## Review Feedback Addressed

Pushed a fix commit addressing all review items:

### 🚨 Critical: Orphaned Data → Soft-Delete
Changed from hard `DELETE FROM observers` to soft-delete with `inactive` column. Observers are now marked `inactive = 1` instead of deleted, preserving JOIN integrity for `observations.observer_idx` and `observer_metrics.observer_id`. Historical data retains observer attribution. Orphaned `observer_metrics` rows are cleaned up when observers go inactive.

### ⚠️ Observer Resurrection → Reactivation
`UpsertObserver` now reactivates previously inactive observers (`SET inactive = 0 WHERE id = ? AND inactive = 1`). Same observer_id, same rowid — no more disconnected history.

### Other Fixes
- **packets_v view**: LEFT JOIN now filters inactive observers (`inactive IS NULL OR inactive = 0`)
- **Breaking API**: Restored `deleted` as legacy alias alongside `packets_deleted`; renamed `observers_removed` → `observers_inactive`
- **Duplicate defaults**: Removed `observerDays == 0` fallback from `RemoveStaleObservers()`, callers use `ObserverDaysOrDefault()`
- **Ingestor stagger**: Added 90s initial sleep (matching server's 3min stagger pattern)
- **Config churn**: Reverted geo_filter polygon reformatting
- **Migration**: Added `inactive` column with migration guard for existing DBs
- **Tests**: Added soft-delete verification, reactivation test, inactive flag checks

Kpa-clawbot (approved):
## Final Review — Observer Retention (Soft-Delete)

Reviewed through inversion thinking (failure modes) and code quality (simplicity, correctness) lenses.

### What's Fixed Since Last Review ✅

All critical items from the prior review have been addressed:

- **Orphaned data → soft-delete**: Observers are now marked `inactive = 1` instead of deleted. `observations.observer_idx` and `observer_metrics.observer_id` references remain valid. This is the right call.
- **Observer resurrection → reactivation**: `UpsertObserver` clears `inactive = 0` on re-seen observers, preserving the same rowid. Historical observations stay connected. Exactly what was needed.
- **Duplicate default logic**: Removed from `RemoveStaleObservers`, callers use `ObserverDaysOrDefault()`. Clean.
- **API backward compat**: `deleted` kept as legacy alias alongside `packets_deleted`. Good.
- **Ingestor stagger**: 90s initial sleep added. Acceptable.
- **Config churn**: Reverted.
- **Migration**: Properly guarded with `_migrations` table check.

### Remaining Items

**1. `packets_v` view excludes inactive observers (behavioral note)**

The view JOIN condition `obs.inactive IS NULL OR obs.inactive = 0` means any query through `packets_v` will show `NULL` for observer fields on packets that were observed by now-inactive observers. This is *functionally the same* as the hard-delete behavior for view consumers — the soft-delete only helps direct table JOINs.

This isn't necessarily wrong, but it should be documented: the view intentionally hides inactive observer attribution. If a future consumer wants "show me all observations including inactive observers," they'll need a separate query.

**Severity: informational** — not a blocker, but worth a code comment on the view.

**2. Metrics cleanup is broader than necessary**

```go
s.db.Exec(`DELETE FROM observer_metrics WHERE observer_id IN (SELECT id FROM observers WHERE inactive = 1)`)
```

This deletes metrics for *all* inactive observers every time any new observer goes inactive, not just the newly-inactivated ones. On subsequent runs it's a no-op (nothing new to delete), but on the first run after deploying to an instance with many stale observers, this subselect scans the full observers table. Fine at current scale, but a `WHERE observer_id = ?` loop over just-inactivated IDs would be more precise.

**Severity: minor** — no functional issue, just slightly wasteful.

**3. No dry-run / preview endpoint**

Still no way to preview which observers would be affected before enabling retention. The prior review suggested `GET /api/admin/stale-observers?days=14` or a `dryRun` parameter. This would give operators confidence when deploying to production instances for the first time.

**Severity: nice-to-have** — not a merge blocker, but worth tracking as a follow-up.

**4. Ingestor ticker timing is slightly off**

```go
observerRetentionTicker := time.NewTicker(24 * time.Hour)
go func() {
    time.Sleep(90 * time.Second)
    store.RemoveStaleObservers(observerDays)
    for range observerRetentionTicker.C {
```

The ticker starts counting from goroutine launch, not after the 90s sleep. So the gap between the post-sleep prune and the first ticker prune is ~23h58m30s, not 24h. Cosmetic — no real impact at daily granularity.

**Severity: cosmetic**

### Code Quality

- Clean, readable, well-structured — no unnecessary abstraction
- Good test coverage: soft-delete, reactivation, keep-forever, empty store, config edge cases
- Migration is properly idempotent
- Comments explain *why* (soft-delete for JOIN integrity), not just *what*
- PR description is thorough and honest about test status

### Verdict

**Approve.** All critical issues from the prior review are resolved. The soft-delete approach is the correct design — it preserves data integrity while achieving the goal of hiding stale observers. The remaining items are informational or nice-to-haves that can be follow-ups.

The one thing worth adding before merge: a brief comment on the `packets_v` view explaining why inactive observers are excluded from the JOIN. Future maintainers will wonder.

Joel-Claw:
Regarding the `packets_v` view excluding inactive observers:

The `LEFT JOIN ... WHERE obs.inactive IS NULL OR obs.inactive = 0` filter is intentional. The view is designed as a clean interface for the common case — consumers querying recent packets with active observer metadata. Inactive observers are hidden from the view for the same reason they're removed from the active set: stale observer attribution is noise, not signal.

This does not delete or lose data. The underlying `observations` and `observers` tables retain full history, including inactive observer rows. Any consumer that needs complete historical visibility (e.g., audit logs, historical analysis) can query those tables directly with their own JOIN conditions, bypassing the view's filter.

In short: the view serves the typical query, direct tables serve the exhaustive one.

PR #755, opened 2026-04-15
feat: repeater liveness indicator with relay stats (#662)

efiten:
## Summary

- **Backend**: adds `relayTimes` in-memory index (sorted unix-millis per repeater pubkey), maintained in lockstep with `byPathHop`. Populated at startup from all packet observations (not just best), updated on ingest/evict/backfill. Exposes `relay_count_1h`, `relay_count_24h`, `last_relayed` in both `/api/nodes` (for repeaters) and `/api/nodes/{pubkey}/health`.
- **Frontend**: `getNodeStatus` extended to three-state (`relaying` / `active` / `stale`) for repeaters based on relay_count_24h. `getStatusInfo` is the single source of truth for status label, explanation, and relay stats. Detail pane shows relay counts and last relayed time. Nodes list gets a status emoji column with hover tooltip showing relay info.
- **Correctness fixes**: relay index scans all observations per packet (not just best); backfill now updates relay index after resolving paths; pubkeys lowercased consistently throughout index.

### `cmd/server/store.go`
- `relayTimes map[string][]int64` field added to `PacketStore`
- `addTxToRelayTimeIndex` / `removeFromRelayTimeIndex`: scan all observations, idempotent sorted insert, lowercase keys
- `relayMetrics(times, nowMs)`: returns `(count1h, count24h, lastRelayed)`
- `buildPathHopIndex`: populates `relayTimes` at startup
- `pollAndMerge`: updates relay index on ingest and eviction; new `else` branch for path-unchanged observations
- `addTxToPathHopIndex` / `removeTxFromPathHopIndex`: lowercase resolved pubkeys (fixes casing mismatch with lookup)

### `cmd/server/routes.go`
- `GetBulkHealth` / `GetNodeHealth`: include relay stats for repeater nodes
- `handleNodes`: enriches repeater nodes with relay stats from `relayTimes` so list view has same data as detail pane

### `cmd/server/neighbor_persist.go`
- `backfillResolvedPathsAsync`: calls `addTxToRelayTimeIndex` after `pickBestObservation` to capture newly resolved pubkeys

### `public/roles.js`
- `getNodeStatus(role, lastSeenMs, relayCount24h)`: three-state logic for repeaters
- `getStatusInfo(n)`: single source of truth returning status, label, explanation, relay counts, last relayed

### `public/nodes.js`
- Detail pane: `n.stats` populated from health endpoint before `getStatusInfo` call
- Nodes list: status emoji column with relay hover tooltip; status filter uses `getStatusInfo`

[...]

PR #798, opened 2026-04-19
fix(repeater-liveness): address issue #755 review feedback

efiten:
## Summary

- **Perf:** `addTxToRelayTimeIndex` now accepts pre-parsed millis — eliminates 30K+ redundant RFC3339 parses during startup rebuild under the write lock
- **Perf:** `removeFromRelayTimeIndex` uses `copy+reslice` instead of `append` — no per-eviction heap allocation
- **Lock scope:** `handleNodes` snapshots relay times under `RLock` then releases before enrichment, keeping lock held for a quick map copy only
- **API clarity:** `last_relayed` suppressed when both counts are zero across all three endpoints
- **Should-fix:** NOTE comment on `HEALTH_THRESHOLDS` mirror in test file; repeater template rows flattened to inline conditionals

## Test plan

- [ ] `go test ./...` passes (excluding pre-existing `TestMultiByteCapability_*` failures)
- [ ] `node test-repeater-liveness.js` — 10 passed, 0 failed
- [ ] Startup log shows relay-time keys populated correctly
- [ ] `/api/nodes` repeater nodes: `last_relayed` absent when both counts are 0, present when counts > 0


efiten:
Closing — fixes should go on the existing PR #755 branch instead.

PR #711, opened 2026-04-12
fix: redesign memory eviction — self-accounting trackedBytes, watermarks, safety cap

Kpa-clawbot:
## Problem

`HeapAlloc`-based eviction cascades on large databases — evicts down to near-zero packets because Go runtime overhead exceeds `maxMemoryMB` even with an empty packet store.

## Fix (per Carmack spec on #710)

1. **Self-accounting `trackedBytes`** — running counter maintained on insert/evict, computed from actual struct sizes. No `runtime.ReadMemStats`.
2. **High/low watermark hysteresis** (100%/85%) — evict to 85% of budget, don't re-trigger until 100% crossed again.
3. **25% per-pass safety cap** — never evict more than a quarter of packets in one cycle.
4. **Oldest-first** — evict from sorted head, O(1) candidate selection.

`maxMemoryMB` now means packet store budget, not total process heap.

Fixes #710

PR #708, opened 2026-04-12
fix: index relay hops in byNode for liveness tracking

Kpa-clawbot:
## Problem

Nodes that only appear as relay hops in packet paths (via `resolved_path`) were never indexed in `byNode`, so `last_heard` was never computed for them. This made relay-only nodes show as dead/stale even when actively forwarding traffic.

Fixes #660

## Root Cause

`indexByNode()` only indexed pubkeys from decoded JSON fields (`pubKey`, `destPubKey`, `srcPubKey`). Relay nodes appearing in `resolved_path` were ignored entirely.

## Fix

`indexByNode()` now also iterates:
1. `ResolvedPath` entries from each observation
2. `tx.ResolvedPath` (best observation's resolved path, used for DB-loaded packets)

A per-call `indexed` set prevents double-indexing when the same pubkey appears in both decoded JSON and resolved path.

Extracted `addToByNode()` helper to deduplicate the nodeHashes/byNode append logic.

## Scope

**Phase 1 only** — server-side in-memory indexing. No DB changes, no ingestor changes. This makes `last_heard` reflect relay activity with zero risk to persistence.

## Tests

5 new test cases in `TestIndexByNodeResolvedPath`:
- Resolved path pubkeys from observations get indexed
- Null entries in resolved path are skipped
- Relay-only nodes (no decoded JSON match) appear in `byNode`
- Dedup between decoded JSON and resolved path
- `tx.ResolvedPath` indexed when observations are empty

All existing tests pass unchanged.

## Complexity

O(observations × path_length) per packet — typically 1-3 observations × 1-3 hops. No hot-path regression.

Issue #800, opened 2026-04-19
perf: remove per-StoreTx ResolvedPath, replace with membership index + on-demand decode (final spec)


Kpa-clawbot:
## Goal

Cut server startup heap by ~900 MB on databases with 1M+ observations by removing per-`StoreTx`/`StoreObs` `ResolvedPath` slices. Unblocks #791.

**Discussion, profiling data, and full review history:** #799 (closed in favor of this spec).

## Profile (in brief)

`pprof inuse_space` on a representative server (388K obs, 630 MB heap):
- `Load()` → `unmarshalResolvedPath` JSON decode = **33% of heap**
- Per-obs `ResolvedPath []*string` slices = **~1 GB extrapolated to 1.66M obs (#791 user)**

### Remove
`ResolvedPath []*string` field from `StoreTx` and `StoreObs`. Compile-time guard test ensures it stays gone.

### Add
| Structure | Purpose | Estimated cost (1M obs) |
|---|---|---:|
| `resolvedPubkeyIndex map[uint64][]int` | xxhash64(pubkey) → []txID. Forward index for "Paths through node X" + collision-safety candidates. | 50–120 MB |
| `resolvedPubkeyReverse map[int][]uint64` | txID → []hashes it was indexed under. Required for clean removal on eviction / backfill re-index. | ~40 MB |
| `apiResolvedPathLRU` (sized 10K, ~200 B each) | Cache for on-demand API decode of `resolved_path`. **Mandatory for live polling path.** | ~2 MB |

### Decode-window discipline (single rule)
`resolved_path` JSON is decoded at exactly one place per packet (ingest / `Load()` row). During that decode window, all consumers are fed in this order, then the temporary `[]*string` is dropped — never lands on the struct:

1. `addToByNode` — relay node indexing
2. `touchRelayLastSeen` — relay liveness DB updates (#660 / #755)
3. `addTxToPathHopIndex` resolved-pubkey branch — `byPathHop` full-pubkey keys
4. `resolvedPubkeyIndex` + `resolvedPubkeyReverse` insert
5. WebSocket broadcast map (carries the raw JSON bytes — no struct mutation)
6. Persist batch (carries the raw JSON bytes for SQL UPDATE)

Enforced by: (a) struct field gone (compile-time), (b) godoc on `ingestObservationDecoded()` documenting the contract, (c) test that broadcast maps include `resolved_path` post-refactor.

### On-demand SQL fetch (cold path)
[...]

PR #801, opened 2026-04-20
fix: re-index relay hops in byNode after Load() picks best observation (#692)

### LINKED ISSUE #692: Loss of analytics for repeater/no additional analytics being reported on day of new release
https://analyzer.cascadiamesh.org#/nodes/33338b4fbb9e89a717ab0bc2777f5010f79837044dba135251619047492e1274

Yesterday, my repeater was reporting 100% availability, signal grade of B, 2 packets/day, 13 observers who had heard the node, and 18 hours as the longest silence. 

This morning after the update to the Analyzer, all that data was lost. It is still in that condition as of 17:00PST today, although it shows traffic at my end.

Image

This is a new repeater, about 6 days running. It is a Wismesh 1W system running v1.14.1 firmware, with good battery power. 

RSSI is typically around -32, SNR around 12, and noise floor around -105dB.

efiten:
## Problem

`indexByNode()` was called during `Load()` immediately when each `StoreTx` was created — before observations were appended and before `pickBestObservation()` set `tx.ResolvedPath`. The resolved_path indexing branch added in #708 was effectively dead code on every server restart.

**Symptom:** After any restart, `byNode[relay_pubkey]` was empty for relay-only nodes even when `resolved_path` was correctly persisted in the DB. Analytics showed `totalPackets = 0` for repeater nodes despite active relay traffic.

## Fix

Call `s.indexByNode(tx)` again in the post-load loop after `pickBestObservation()`, where `ResolvedPath` is populated. Same fix applied to `backfillResolvedPathsAsync()`, which also called `pickBestObservation()` without re-indexing afterward.

The dedup in `nodeHashes` prevents double-counting: pubkeys already indexed from decoded JSON fields are skipped; only the relay hop pubkeys from `resolved_path` are new additions.

## Test

`TestLoadIndexesRelayHopsFromResolvedPath` — inserts a packet with `resolved_path` containing a relay pubkey that does not appear in `decoded_json`, calls `Load()`, and verifies `byNode[relay_pubkey]` is populated.

## Related

Closes #692 (together with #707, #708, #711 already merged)

### DIFF
--- a/cmd/server/db_test.go
+++ b/cmd/server/db_test.go
@@ -1975,3 +1975,47 @@ func TestParseWindowDuration(t *testing.T) {
 		}
 	}
 }
+
+// TestLoadIndexesRelayHopsFromResolvedPath verifies that after Load(), relay
+// nodes that appear only in resolved_path (not in decoded_json) are indexed
+// in byNode. This was broken: indexByNode was called before observations were
+// appended, so tx.ResolvedPath was nil at index time.
+func TestLoadIndexesRelayHopsFromResolvedPath(t *testing.T) {
+	db := setupTestDB(t)
+	defer db.Close()
+
+	now := time.Now().UTC().Add(-1 * time.Hour).Format(time.RFC3339)
+	epoch := time.Now().UTC().Add(-1 * time.Hour).Unix()
+
+	// Insert a node whose pubkey does NOT appear in any decoded_json —
+	// it only relays traffic (appears in resolved_path of other packets).
+	const relayPubkey = "relay000aabbccddeeff0011"
+	const senderPubkey = "sender00112233445566"
+
+	db.conn.Exec(`INSERT INTO transmissions (raw_hex, hash, first_seen, route_type, payload_type, decoded_json)
+		VALUES ('FF01', 'relaytest0001hash', ?, 1, 4, ?)`,
+		now, `{"pubKey":"`+senderPubkey+`","name":"Sender","type":"ADVERT"}`)
+
+	// Observer hears the packet via the relay node.
+	db.conn.Exec(`INSERT INTO observations (transmission_id, observer_idx, snr, rssi, path_json, timestamp, resolved_path)
+		VALUES (1, 1, 10.0, -90, '["rr"]', ?, ?)`,
+		epoch, `["`+relayPubkey+`"]`)
+
+	store := NewPacketStore(db, nil)
+	if err := store.Load(); err != nil {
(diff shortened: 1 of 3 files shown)


Kpa-clawbot (commented):
## Review — REVIEW-ONLY (PR is CONFLICTING, do not merge as-is)

**Verdict:** Bug is real (#692 root cause well-diagnosed), test is excellent — but the fix mechanism is **superseded by #806** which already landed on master. Recommend rebasing and **keeping only the test**, dropping the code changes.

### Conflict scope
Rebasing onto `origin/master` conflicts in:
- `cmd/server/neighbor_persist.go` (backfillResolvedPathsAsync block)
- `cmd/server/store.go` (auto-merges, but the change is now dead code — see below)

Conflicts with **#806** (`9e90548 perf(#800): remove per-StoreTx ResolvedPath, replace with membership index + on-demand decode`).

### Why the fix is now obsolete
After #806, `Load()` indexes resolved-path pubkeys **inline per-observation** as rows are scanned (store.go:498–521): `addToByNode` + `addToResolvedPubkeyIndex` are called right when `resolved_path` is read off the row. `indexByNode()` was deliberately scoped down to **only** decoded_json fields (`pubKey/destPubKey/srcPubKey`) — it no longer touches resolved paths at all.

So the PR's post-`pickBestObservation` re-call of `s.indexByNode(tx)` is a no-op for relay hops on master. Same story for the `neighbor_persist.go` hunk — master's backfill path now uses `removeFromResolvedPubkeyIndex` / `extractResolvedPubkeys` / `addToResolvedPubkeyIndex` instead.

### What to keep
`TestLoadIndexesRelayHopsFromResolvedPath` is a **valuable regression test** and almost certainly passes against master as-is (lock in #806's fix for #692). Please rebase, drop the store.go + neighbor_persist.go hunks, and re-run the test against master.

### Correctness note (for the historical fix, dijkstra)
Even on the pre-#806 codebase, the fix only indexes pubkeys from the **best (longest)** observation's resolved_path, not from *all* observations. Relay nodes that only appear on shorter paths would still be missed. #806's per-observation indexing during Load() handles this correctly.

### Perf (carmack)
N/A — superseded.

### Action for author
1. `git fetch origin && git rebase origin/master`
2. Resolve by **dropping** the store.go + neighbor_persist.go changes.
3. Verify the test passes on master (it should — #806 fixes #692).
4. If it does, repurpose this PR as "test: regression for #692 (relay-hop indexing on Load)".

Issue #694, opened 2026-04-10
RF Analytics (SNR, RSSI, Noise Floor) show no data — upstream MQTT clients don't send RF fields


Kpa-clawbot:
### What's broken

The RF Analytics tab shows empty/zero values for SNR distribution, RSSI distribution, SNR vs RSSI scatter, and SNR by payload type. The observer list shows `noise_floor: null` for all observers.

**API evidence:**
- `GET /api/analytics/rf` — returns packet size and payload type distributions but SNR/RSSI sections are empty (0 data points)
- `GET /api/observers` — all observers show `noise_floor: null`
- `GET /api/packets?limit=5` — all packets show `snr: null, rssi: null`

### Root cause

**This is a data source problem, not a CoreScope code bug.** The MQTT observer clients in the wild are not sending SNR/RSSI in their packet messages.

Investigation of live MQTT traffic on the production broker shows:

1. **Packet messages** (`meshcore/+/+/packets`): 10/10 sampled messages from various observers contain `raw`, `hash`, `direction`, `duration`, etc. but **no `SNR` or `RSSI` fields**. The ingestor correctly parses these fields when present (both cases `SNR`/`snr`), but they simply aren't in the messages.

2. **Status messages** (`meshcore/+/+/status`): Split by client version:
   - `meshcoretomqtt/1.0.8.0` — includes `stats.noise_floor` ✅
   - `meshcoretomqtt/1.0.7.0` and `1.1.0.0` variants — do NOT include `stats` or `noise_floor` ❌
   - `pyMC_repeater/*` — does NOT include `stats` or `noise_floor` ❌

3. **Historical data**: Only 5,399 out of 1.7M+ observations in the DB ever had SNR/RSSI values. All came from a single test observer (`observer_idx=33`) active between March 24-27, 2026. When that observer went offline, SNR data stopped entirely.

4. **Noise floor**: Zero rows in `observer_metrics` have `noise_floor IS NOT NULL`. The 1.0.8.0 clients that DO send it are in regions outside the production IATA filter, so their status messages are dropped before processing.

### Why it appeared to work before

The test observer (idx=33) was active March 24-27 and populated ~5,400 observations with SNR/RSSI. The in-memory store showed this data in RF analytics. After the server's retention window evicted those packets, the RF analytics became empty.

### Two separate issues

[...]

PR #802, opened 2026-04-20
fix: bypass IATA filter for status messages, fill SNR on duplicate obs (#694)

efiten:
## Problems

Two independent ingestor bugs identified in #694:

### 1. IATA filter drops status messages from out-of-region observers

The IATA filter ran at the top of `handleMessage()` before any message-type discrimination. Status messages carrying observer metadata (`noise_floor`, battery, airtime) from observers outside the configured IATA regions were silently discarded before `UpsertObserver()` and `InsertMetrics()` ran.

**Impact:** Observers running `meshcoretomqtt/1.0.8.0` in BFL and LAX — the only client versions that include `noise_floor` in status messages — had their health data dropped entirely on prod instances filtering to SJC.

**Fix:** Moved the IATA filter to the packet path only (after the `parts[3] == "status"` branch). Status messages now always populate observer health data regardless of configured region filter.

### 2. `INSERT OR IGNORE` discards SNR/RSSI on late arrival

When the same `(transmission_id, observer_idx, path_json)` observation arrived twice — first without RF fields, then with — `INSERT OR IGNORE` silently discarded the SNR/RSSI from the second arrival.

**Fix:** Changed to `ON CONFLICT(...) DO UPDATE SET snr = COALESCE(excluded.snr, snr), rssi = ..., score = ...`. A later arrival with SNR fills in a `NULL`; a later arrival without SNR does not overwrite an existing value.

## Tests

- `TestIATAFilterDoesNotDropStatusMessages` — verifies BFL status message is processed when IATA filter includes only SJC, and that BFL packet is still filtered
- `TestInsertObservationSNRFillIn` — verifies SNR fills in on second arrival, and is not overwritten by a subsequent null arrival

## Related

Partially addresses #694 (upstream client issue of missing SNR in packet messages is out of scope)

### DIFF
--- a/cmd/ingestor/db.go
+++ b/cmd/ingestor/db.go
@@ -433,8 +433,12 @@ func (s *Store) prepareStatements() error {
 	}
 
 	s.stmtInsertObservation, err = s.db.Prepare(`
-		INSERT OR IGNORE INTO observations (transmission_id, observer_idx, direction, snr, rssi, score, path_json, timestamp)
+		INSERT INTO observations (transmission_id, observer_idx, direction, snr, rssi, score, path_json, timestamp)
 		VALUES (?, ?, ?, ?, ?, ?, ?, ?)
+		ON CONFLICT(transmission_id, observer_idx, COALESCE(path_json, '')) DO UPDATE SET
+			snr   = COALESCE(excluded.snr,   snr),
+			rssi  = COALESCE(excluded.rssi,  rssi),
+			score = COALESCE(excluded.score, score)
 	`)
 	if err != nil {
 		return err
--- a/cmd/ingestor/db_test.go
+++ b/cmd/ingestor/db_test.go
@@ -1882,3 +1882,89 @@ func TestExtractObserverMetaNewFields(t *testing.T) {
 		t.Errorf("RecvErrors = %v, want 3", meta.RecvErrors)
 	}
 }
+
+// TestInsertObservationSNRFillIn verifies that when the same observation is
+// received twice — first without SNR, then with SNR — the SNR is filled in
+// rather than silently discarded. The unique dedup index is
+// (transmission_id, observer_idx, COALESCE(path_json, '')); observer_idx must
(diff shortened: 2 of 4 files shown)


Kpa-clawbot (approved):
## Review — APPROVE ✅

Tests: `cmd/ingestor` PASS (29.9s), `cmd/server` filter/SNR/observation tests PASS. New tests cover both behaviors directly.

### meshcore semantics ✅
The `/status` topic carries **observer-level metadata** (noise_floor, battery, origin name) — it is *not* a MeshCore packet payload type. The IATA filter exists to scope **packet ingestion** to relevant regions; observer health metrics are region-agnostic by design. The reorder (handle status → then apply IATA to packets) matches the diagnosis in #694: 1.0.8.0 clients sending noise_floor were being dropped before their status handler ran. Fix is correct.

### Correctness (dijkstra) ✅
ON CONFLICT target `(transmission_id, observer_idx, COALESCE(path_json, ''))` matches the existing UNIQUE index at `db.go:180` exactly — required for SQLite. `COALESCE(excluded.snr, snr)` preserves existing non-null values (verified by 3rd-arrival assertion in the new test). Row counts unchanged vs `INSERT OR IGNORE` (same dedup key), so packet counts and downstream aggregations are stable; only the SNR/RSSI columns get late-fill, which is the intended semantic.

### Security (djb) ✅
The IATA bypass does not open a flood path: the status handler only upserts a single row per `observer_id` into the `observers` table — it does not touch `transmissions`/`observations`. Observer-metadata spoofing across regions is a **pre-existing** concern (the broker has no per-observer auth, and IATA was never an authentication mechanism), not introduced here. Bounded write surface, no amplification.

### Perf (carmack) ✅
ON CONFLICT adds negligible per-insert overhead on an existing UNIQUE index lookup. handleMessage reorder is work-equivalent — the IATA branch just runs after the early status return.

### Style (torvalds) ✅
Clean diff. Inline comment explains *why* status bypasses the filter. Test names self-documenting. No nits worth blocking on.

Ship it.

PR #803, opened 2026-04-20
fix(#673): replace raw JSON text search with byNode index for node packet queries

### LINKED ISSUE #673: bug: chat messages containing a node's pubkey are incorrectly counted as packets for that node
## Bug

When a chat message (GRP_TXT) contains the full 64-char hex pubkey of a node in its text content, that packet is incorrectly associated with the node — inflating packet counts and appearing in the node's analytics/packet list.

## Example

Node: [Quail Hollow Park 🌲](https://analyzer.00id.net/#/nodes/7502f19f44cad6d7b626e1d811c00a914af452636182ccded3fd019803395ec9) (repeater)

```
pubkey: 7502f19f44cad6d7b626e1d811c00a914af452636182ccded3fd019803395ec9
```

The analytics endpoint shows a GRP_TXT packet counted for this node:

```json
"packetTypeBreakdown": [
  {"payload_type": 4, "count": 53},   // ADVERT — expected
  {"payload_type": 11, "count": 320}, // CONTROL — expected
  {"payload_type": 5, "count": 1}     // GRP_TXT — UNEXPECTED
]
```

A repeater node should never have GRP_TXT packets associated with it. This GRP_TXT packet was a chat message that happened to contain the node's pubkey in its text (e.g. someone sharing the analyzer link in chat).

## Root Cause

Two code paths in `cmd/server/store.go` use raw substring matching on `DecodedJSON` to find packets related to a node. This matches pubkeys appearing *anywhere* in the JSON — including inside encrypted/encoded data fields or message text.

### Affected code paths

**1. `GetNodeAnalytics()` — line 5729:**
```go
} else if tx.DecodedJSON != "" && (strings.Contains(tx.DecodedJSON, name) || strings.Contains(tx.DecodedJSON, pubkey)) {
```

**2. General packet query filter — line 1931:**
```go
if !strings.Contains(tx.DecodedJSON, nodePK) && !strings.Contains(tx.DecodedJSON, q.Node) {
```

Both do `strings.Contains` on the entire decoded JSON blob. A GRP_TXT message whose `encryptedData` hex or any other field happens to contain the 64-char pubkey as a substring will match.

The `byNode` index itself (line 527) is correct — it only indexes on structured fields (`pubKey`, `destPubKey`, `srcPubKey`). The bug is in the supplementary text search that runs alongside the index.

### Why text search exists

The comment on line 5715 says `"matches Node.js findPacketsForNode"` — this was ported from the old Node.js backend. The intent was to catch packets that reference a node by name or pubkey in their decoded data but aren't indexed by the structured fields. The problem is that raw `strings.Contains` on the full JSON is too broad.

## Proposed Fix

Replace the raw `strings.Contains` text search with structured field matching. Instead of searching the entire JSON string, parse the JSON and check only specific fields that legitimately reference nodes:

- `pubKey`, `destPubKey`, `srcPubKey` (already covered by `byNode` index)
- `destHash` (for ANON_REQ)
- Node name matches should only check `name`-like fields, not encrypted data or message content

Alternatively, since the `byNode` index already covers the structured pubkey fields, the text search fallback could be removed entirely if no legitimate use case requires it. If the name-based search is needed (finding packets that mention a node by name), it should search only specific decoded fields, not the raw JSON blob.

efiten:
## Summary

Fixes #673

- GRP_TXT packets whose message text contains a node's pubkey were incorrectly counted as packets for that node, inflating packet counts and type breakdowns
- Two code paths in `store.go` used `strings.Contains` on the full `DecodedJSON` blob — this matched pubkeys appearing anywhere in the JSON, including inside chat message text
- `filterPackets` slow path (combined node + other filters): replaced substring search with a hash-set membership check against `byNode[nodePK]`
- `GetNodeAnalytics`: removed the full-packet-scan + text search branch entirely; always uses the `byNode` index (which already covers `pubKey`/`destPubKey`/`srcPubKey` via structured field indexing)

## Test Plan

- [x] `TestGetNodeAnalytics_ExcludesGRPTXTWithPubkeyInText` — verifies a GRP_TXT packet with the node's pubkey in its text field is not counted in that node's analytics
- [x] `TestFilterPackets_NodeQueryDoesNotMatchChatText` — verifies the combined-filter slow path of `filterPackets` returns only the indexed ADVERT, not the chat packet

Both tests were written as failing tests against the buggy code and pass after the fix.

### DIFF
--- a/cmd/server/issue673_test.go
+++ b/cmd/server/issue673_test.go
@@ -1,0 +1,107 @@
+package main
+
+import (
+	"encoding/json"
+	"testing"
+	"time"
+
+	_ "modernc.org/sqlite"
+)
+
+const issue673NodePK = "7502f19f44cad6d7b626e1d811c00a914af452636182ccded3fd019803395ec9"
+
+// setupIssue673Store builds an in-memory store with one repeater node having:
+//   - one ADVERT packet (legitimately indexed in byNode)
+//   - one GRP_TXT packet whose decoded text contains the node's pubkey (false-positive candidate)
+func setupIssue673Store(t *testing.T) (*PacketStore, *DB) {
+	t.Helper()
+	db := setupTestDB(t)
+
+	_, err := db.conn.Exec(
+		"INSERT INTO nodes (public_key, name, role) VALUES (?, ?, ?)",
+		issue673NodePK, "Quail Hollow Park", "repeater",
+	)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	ps := NewPacketStore(db, nil)
+	now := time.Now().UTC().Format(time.RFC3339)
+
+	pt4 := 4 // ADVERT
+	pt5 := 5 // GRP_TXT
+
+	advertDecoded, _ := json.Marshal(map[string]interface{}{"pubKey": issue673NodePK})
+	advert := &StoreTx{
+		ID:          1,
+		Hash:        "advert_hash_673",
+		PayloadType: &pt4,
+		DecodedJSON: string(advertDecoded),
+		FirstSeen:   now,
+	}
+
+	otherPK := "aabbccddaabbccddaabbccddaabbccddaabbccddaabbccddaabbccddaabbccdd"
+	chatDecoded, _ := json.Marshal(map[string]interface{}{
+		"srcPubKey": otherPK,
+		"text":      "Check out node " + issue673NodePK + " on the analyzer",
+	})
+	chat := &StoreTx{
+		ID:          2,
+		Hash:        "chat_hash_673",
+		PayloadType: &pt5,
+		DecodedJSON: string(chatDecoded),
+		FirstSeen:   now,
+	}
+
+	ps.mu.Lock()
+	ps.packets = append(ps.packets, advert, chat)
+	ps.byHash[advert.Hash] = advert
+	ps.byHash[chat.Hash] = chat
+	ps.byTxID[advert.ID] = advert
+	ps.byTxID[chat.ID] = chat
+	ps.byNode[issue673NodePK] = []*StoreTx{advert}
+	ps.mu.Unlock()
+
+	return ps, db
+}
+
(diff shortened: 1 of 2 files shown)


Kpa-clawbot (approved):
Reviewed across perf / correctness / code-quality / protocol-semantics lenses.

**Verdict:** ✅ Approve. Tight, surgical fix that matches the issue's preferred direction (drop raw-JSON text search; trust the structured \`byNode\` index).

**Perf (carmack):**
- Old slow path: full scan of \`s.packets\` (~30K) with two \`strings.Contains\` against a 64-char needle per tx — quadratic-ish in pathological cases.
- New: O(K) hashSet build (K = \`len(byNode[pk])\`) + O(1) map lookup per tx. Index is incrementally maintained in \`addToByNode\` on ingest — zero cold-start cost, zero extra memory beyond the existing \`nodeHashes\` set.
- Fast path (node-only query) already returns \`s.byNode[nodePK]\` directly. Big win for the common UI call.
- Minor nit: the new \`nodeHashSet\` is built before the fast-path \`return indexed\` check, so it's allocated and discarded on the fast path. Trivial — feel free to ignore or move below the fast-path block.

**Correctness (dijkstra):**
- \`byNode\` is populated from structured fields (\`pubKey\`/\`destPubKey\`/\`srcPubKey\`) + every resolved-path relay → captures sender, recipient, AND forwarders. No legitimate \"this is a packet for node X\" case is lost vs. the old text search.
- Old name-based fallback (\`strings.Contains(json, name)\`) is gone — that's the whole point; matched the bug.
- Out of scope but worth a follow-up: ANON_REQ \`destHash\` is not in the index; if operators need those surfaced under the destination node, file a separate issue.

**Semantics (meshcore):** New definition = \"node appears as endpoint or path member\" — matches operator intent for \"packets for this node.\" Chat mentions correctly excluded.

**Code quality (torvalds):** Clean diff, dead \`name\` lookup removed, comment explains the *why*. Good.

**Tests:** \`go test ./... -run \"Node|byNode|Packet\" -count=1\` → PASS. Two new regression tests cover both code paths (analytics + general filter). 👍

Issue #804, opened 2026-04-20
Bug/feature: Use actual repeaters region data and not observer based


luckystriike22:
Currently, a lot of data appears to be grouped based on the location of an observer. For instance, the 2-byte hash data for Maastricht (MST, the Netherlands) is collected by an observer that is marked as being in MST. Even data from Belgium that flows through that observer is labeled as MST. In my opinion, this gives a misleading impression of the data. It should instead be marked based on the location of the repeater that the data is passing through.

**Issue:**
Repeater data flowing through an MST observer is marked as MST. As a result, the 2-byte hash data that is collected becomes polluted with data from other regions. It would be helpful if the data could be filtered based on the actual region and if it were possible to retrieve all repeaters located within a specific region.

**Possible fix:**
* Add a way to configure (sub)region polygons per country.
* Mark each observer as belonging to a specific region.
* When new repeaters appear, assign them to a region based on an advertisement that includes the repeater’s location.
* Since repeaters are already stored in the database, data can be linked to a repeater. This can be done by mapping on the key byte (for 2-byte data) or, in some cases, on the full key.
* Allow data on the site to be filtered by region.

**Possible issues:**
* Mapping may be difficult because many users still use 1-byte messages, although this issue should diminish over time.
* This approach may require additional resources, as data will need to be mapped to both a repeater and its region.

efiten:
Hi @luckystriike22 — thanks for the detailed write-up. We've worked out a design for this and would appreciate your review before we start implementing.

---

## Design: Area-Based Visual Node Filter

### Problem

The existing `?region=` filter uses the **observer's** IATA code to attribute data. Any packet heard by an MST observer is tagged as MST — even if the transmitting node is in Belgium. This pollutes per-region analytics with traffic from neighbouring regions.

### Approach

A new, separate display-side filter (`?area=`) that attributes packets to regions based on the **transmitting node's own GPS coordinates** (from its advert). The existing observer-based `?region=` filter is kept unchanged alongside it.

Key decisions:
- **Display filter only** — no ingestor changes, no schema migrations.
- **Nodes without GPS are excluded** when the filter is active.
- **Only the originating node counts** — relay hops do not determine area membership.
- **Only the node's own advertised lat/lon is used** — relay hops are not used to infer region.

### Config (`config.json`, server only)

Named polygon areas, same coordinate format as the existing geofilter:

```json
{
  "areas": {
    "MST": {
      "label": "Maastricht",
[...]

PR #839, opened 2026-04-21
feat: area-based visual node filter — attribute packets by transmitter GPS (#804)

efiten:
## Summary

- Adds configurable GPS polygon areas to `config.json`; nodes are attributed to an area if their last-known position falls inside the polygon
- New `Area: …` dropdown filter (matching the existing region filter style) appears on all analytics, nodes, packets, map, and live screens when areas are configured
- Backend resolves area membership with a 30s TTL cache; area filter bypasses the 500-node cap on `/api/bulk-health` so all area nodes are always returned
- Includes a polygon builder tool (`/area-map.html`) for drawing and exporting area boundaries

## Changes

**Backend**
- `AreaEntry` type + `Areas` config field
- `GetNodePubkeysInArea` DB query + `resolveAreaNodes` (30s TTL, `areaNodeMu` RWMutex)
- `PacketQuery.Area` + `filterPackets` polygon check
- `?area=` param propagated through all analytics, topology, clock-health, and bulk-health routes
- `/api/config/areas` endpoint

**Frontend**
- `area-filter.js`: single-select dropdown, persists to localStorage, cleans up stale keys on load
- Wired into analytics, nodes, packets, channels, map, and live pages
- Live map clears node markers on area change

**Docs & tools**
- `docs/user-guide/area-filter.md` — configuration and usage guide
- `docs/api-spec.md` — updated with new endpoint and `?area=` param table
- `tools/area-map.html` — polygon builder for defining area boundaries
- Demo areas added to `config.example.json`

## Test plan

- [x] No areas configured → filter dropdown does not appear on any page
- [x] Areas configured → dropdown appears, "All" selected by default
- [x] Selecting an area filters nodes/packets/topology/map correctly
- [x] Selecting "All" restores unfiltered view
- [x] Selection persists across page reloads (localStorage)
- [x] Stale localStorage key (area removed from config) is cleared on load
- [x] `/api/bulk-health?area=X` returns all nodes in area (no 500-node cap)
- [x] `/api/config/areas` returns correct list


efiten:
Area map can be tested here: https://staging.on8ar.eu/area-map.html
Just click on 'load' on top, and the area's are shown.

Area view can be tested: https://staging.on8ar.eu/#/analytics
For example, the 'hash issues', can be filtered by area, what the TS requested

efiten:
## Area filter — screen-by-screen update

Follow-up changes to the area filter pill placement and wiring:

| Screen | Area filter | Action |
|--------|-------------|--------|
| Packets | NO | leave as-is (already removed) |
| Map | YES | add to map controls |
| Live | YES | add to map controls |
| Channels (main) | NO | remove pill |
| Nodes | YES | leave as-is (OK) |
| Traces | NO | leave as-is (no pill) |
| Observers | NO | leave as-is (no pill) |
| Analytics › Overview | YES | leave as-is (OK) |
| Analytics › RF/Signal | YES | fix — filter not applying |
| Analytics › Topology | YES | fix — changing area doesn't update content |
| Analytics › Channels | NO | remove pill |
| Analytics › Hash stats | YES | leave as-is (OK) |
| Analytics › Hash issues | YES | leave as-is (OK) |
| Analytics › Route patterns | NO | remove pill |
| Analytics › Nodes | YES | leave as-is (OK) |
| Analytics › Distance | NO | remove pill |
| Analytics › Neighbor graphs | NO | remove pill |
| Analytics › RF Health | NO | remove pill |
| Analytics › Clock health | YES | leave as-is (OK) |

Also: the 'All' pill (no-filter reset) appears empty — text is not visible, likely a CSS issue to fix at the same time.

luckystriike22:
- Maybe make the area pills smaller and collapsable/scrollable. It uses a lot of space on all pages.

- when a area is selected the button turns blue and hides the text. Swap text to white when selected

image

Kpa-clawbot (changes requested):
## Consolidated review — PR #839

Two parallel reviewers (adversarial + protocol context). **Verdict: NEEDS-WORK — high regression risk vs recent perf merges.**

### MUST-FIX

1. **Docs lie about GPS-less nodes.** `docs/user-guide/area-filter.md:48-50` claims: *"Nodes with no GPS fix (lat=0, lon=0 or missing coordinates) are always allowed through regardless of the active area filter."* Code does the opposite — `cmd/server/db.go:GetNodePubkeysInArea` explicitly drops `!lat.Valid || !lon.Valid` and `(0,0)` nodes, and `handleNodes`/`filterPackets`/`GetBulkHealth` keep only members of `areaNodes[pk]`. Pick one: change the docs, or actually pass GPS-less nodes through. Currently the docs paragraph is a flat misstatement that operators will rely on.

2. **Area filter silently filters Packets page to ADVERTs only.** Comment at `cmd/server/store.go:2912-2924` admits it: only ADVERT carries `public_key`/`pubKey`; all GRP_TXT/TXT_MSG/REQ/etc. fall through to `pk == ""` and are excluded. But `docs/user-guide/packets.md:35-37` tells users *"shows only packets where the transmitting node's GPS position falls within the area"* — implying it filters all packet types. In reality, selecting any area collapses the Packets table to adverts only. **This is a user-trust break.** Either (a) attribute non-ADVERT packets via `byNode` reverse-lookup of the transmitter, or (b) make the docs and UI explicit ("Area filter: adverts only — other packet types have encrypted senders").

3. **Lock-hold regression vs recently-merged perf work (#1241/#1248).** `GetBulkHealth` at `cmd/server/store.go:~8103` holds `s.mu.RLock()` for the entire body and then calls `s.resolveAreaNodes(area)` inside that critical section. `resolveAreaNodes` issues a full `SELECT * FROM nodes WHERE lat IS NOT NULL AND lon IS NOT NULL` SQLite query on the first uncached call. The in-code comment says *"resolveAreaNodes requires the lock to already be held"* — that's wrong; `resolveAreaNodes` only takes `areaNodeMu`, never `s.mu`. **You're running a slow DB query under the global read lock for no reason** — the exact anti-pattern #1241 (distance lock-drop) and #1248 (steady-state recompute) explicitly removed. `computeAnalyticsDistance` correctly resolves area BEFORE taking `s.mu` — do the same in `GetBulkHealth`.

4. **Topology stats inconsistent when `area` is set.** At `cmd/server/store.go:~6204-6230`, `hopCounts[n]++` and `allHopsList = append(..., n)` run BEFORE the per-hop area filter, so the hop-count histogram and median/average outputs reflect the WHOLE fleet, while `hopFreq` / `pairs` reflect only area-resident hops. Same packet contributes to one set of stats but not the other. Either filter the whole packet up-front (skip the tx if no hop is in the area, or if the transmitter pubkey isn't in the area), or filter both sets consistently. **Currently the topology tab will show misleading per-area "average hops" numbers.**

5. **`resolveAreaNodes` cache TTL is a global flush, not per-key.** `areaNodeCacheTime` at `cmd/server/store.go:3047-3083` is set only on full-reset (cache miss after TTL expiry). If area `BAY` is queried every 5s, it stays cached forever; then a single query for area `SJC` 31s later wipes BOTH and triggers a fresh DB scan for `SJC` only — losing the still-warm `BAY` entry. Trivial fix: track per-key timestamps, or accept this and rename to make it a single-key cache. As-is the behavior doesn't match the surrounding `regionObsCache` pattern operators will assume.

6. **`areaNodeMu` is `sync.Mutex` (not RWMutex), serializes all area lookups across all keys.** Single in-flight `?area=BAY` request blocks every concurrent `?area=SJC` request. With the 30s TTL these blocks happen at minimum every 30s for each area, on every analytics endpoint, on every page load. Switch to RWMutex with double-checked locking, or fetch under a per-key `singleflight.Group`.

7. **Tautological assertions** (anti-tautology violation):
   - `TestAnalyticsRFAreaFilter`: calls `s.GetAnalyticsRF("", "BEL")`, asserts non-nil. A stub `return &AnalyticsRF{}` ignoring the `area` arg passes. Add a node inside + node outside the polygon, assert the RF result reflects the inside node only (or count/identity).
   - `TestAnalyticsChannelsAreaFilter`: same shape, same fix.

8. **`TestResolveAreaNodes_CacheHit` does not test caching.** Calls `resolveAreaNodes("BEL")` twice, asserts pk1 present both times. A no-cache implementation passes. To prove cache: insert pk1, call once, then `DELETE FROM nodes`, call again — second call must still return pk1 (cache hit). Or assert via instrumentation that the SQL query ran exactly once.

9. **Frontend has zero tests.** 129-line `public/area-filter.js` with localStorage persistence, stale-key cleanup, dropdown rendering, and 6 page integrations. At minimum: a `test-area-filter.js` unit suite covering (a) stale-key cleanup on load, (b) selection persists, (c) "All" restores empty state. Plus one Playwright E2E asserting the dropdown filters the packets table.

10. **Bulk-health 500-cap bypass claim untested.** PR body claims `/api/bulk-health?area=X` bypasses the 500-node cap. Insert >500 nodes inside an area, call the route, assert all returned.

### Out-of-scope (flag, don't block — but please split before merge)

- **`docs/superpowers/plans/2026-04-23-geofilter-docs-page.md` (+219)** and **`docs/superpowers/specs/2026-04-23-geofilter-docs-page-design.md` (+85)** — AI scratch / internal planning artifacts. Don't ship to a public repo's `docs/` tree. Move to a private workspace or delete.
- **Duplicate `area-map.html`**: PR adds BOTH `public/area-map.html` (+353) and `tools/area-map.html` (+264). PR body only mentions `tools/area-map.html` — `public/area-map.html` is undocumented and the two are not identical. Pick one location.
- **`tools/geofilter-builder.html` (+20/-21)** — unrelated geofilter-builder churn slipped into an area-filter PR.

### Design observations (informational)

- `queryLimit = 10000` hardcode in `GetBulkHealth` — PR body claims "bypasses the 500-node cap so all area nodes are always returned." It caps at 10000. Fine today, silently broken at the first deployment with >10K GPS-resolved nodes.
- `GetNodePubkeysInArea` does a full nodes-table scan on every TTL miss with no `(lat, lon)` index. At ~2K nodes it's a non-issue; worth an index in a follow-up.
- Cache-key delimiter collision: `cacheKey := region + "|" + area` — `region="A|B", area=""` collides with `region="A", area="B"`. Using `\x00` or `url.QueryEscape` would be safer.
- `computeAnalyticsRoles` calls `GetFleetClockSkew("")` unconditionally — Roles analytics not area-filterable. May be intentional scope cut; flag for the issue tracker.
- `indexByNode` is now called twice during `Load()` (once on `StoreTx` creation, once after `pickBestObservation`). Comment explains why — good. But no test asserts the second pass actually adds relay hops; deserves a regression test.

This PR touches `analytics_recomputer.go`, `clock_skew.go`, `role_analytics.go`, `main.go` — directly modifying the steady-state recompute infrastructure landed in #1248/#1259/#1266. Worth a perf spot-check after the must-fix items land to confirm no regression in those endpoints' latency.

efiten:
Round-2 must-fix items addressed across commits `4878d18`–`6d92e52`:

1. **Docs GPS-less nodes** — removed false claim; docs now match code behavior (GPS-less nodes excluded) (`4878d18`)
2. **Packets page ADVERT-only** — docs and UI now explicit: "Area filter on the Packets page shows ADVERTs only" (`d67556c`)
3. **Lock-hold regression** — `GetBulkHealth` now calls `resolveAreaNodes` before acquiring `s.mu.RLock()`, matching the `computeAnalyticsDistance` pattern (`56ba313`)
4. **Topology stats inconsistency** — `hopCounts`/`allHopsList` now guarded by the same area check as `hopFreq`/`pairs` (`f6de237`)
5. **Per-key TTL cache** — replaced global `areaNodeCacheTime` with per-key timestamps so querying area B doesn't evict still-warm area A (`5a3ba34`)
6. **RWMutex** — `areaNodeMu` upgraded from `sync.Mutex` to `sync.RWMutex`; concurrent reads no longer serialize (`5a3ba34`)
7. **Tautological analytics tests** — `TestAnalyticsRFAreaFilter` and `TestAnalyticsChannelsAreaFilter` now seed an inside + outside node and assert the filtered result reflects only the inside node (`38c89b2`)
8. **`TestResolveAreaNodes_CacheHit`** — deletes the node from DB after first call; second call still returns it, proving cache was used (`38c89b2`)
9. **Frontend tests** — `test-area-filter.js` unit suite (16 tests): default state, localStorage restore, URL encoding, stale-key pruning, valid-key retention, onChange/offChange, and "All" button clears selection. Wired into CI (`0a3eba9`, `6d92e52`)
10. **Bulk-health cap bypass** — `TestGetBulkHealth_AreaBypassesCap`: inserts 510 in-area nodes, asserts all 510 returned (bypasses 10-node limit) (`38c89b2`)

CI: Go Build & Test ✅, Docker Build ✅, E2E ✅. Ready for re-review.

PR #882, opened 2026-04-21
fix(#881): per-observation raw_hex — each observer sees different bytes on air

Kpa-clawbot:
## Problem

Each MeshCore observer receives a physically distinct over-the-air byte sequence for the same transmission (different path bytes, flags/hops remaining). The `observations` table stored only `path_json` per observer — all observations pointed at one `transmissions.raw_hex`. This prevented the hex pane from updating when switching observations in the packet detail view.

## Changes

| Layer | Change |
|-------|--------|
| **Schema** | `ALTER TABLE observations ADD COLUMN raw_hex TEXT` (nullable). Migration: `observations_raw_hex_v1` |
| **Ingestor** | `stmtInsertObservation` now stores per-observer `raw_hex` from MQTT payload |
| **View** | `packets_v` uses `COALESCE(o.raw_hex, t.raw_hex)` — backward compatible with NULL historical rows |
| **Server** | `enrichObs` prefers `obs.RawHex` when non-empty, falls back to `tx.RawHex` |
| **Frontend** | No changes — `effectivePkt.raw_hex` already flows through `renderDetail` |

## Tests

- **Ingestor**: `TestPerObservationRawHex` — two MQTT packets for same hash from different observers → both stored with distinct raw_hex
- **Server**: `TestPerObservationRawHexEnrich` — enrichObs returns per-obs raw_hex when present, tx fallback when NULL
- **E2E**: Playwright assertion in `test-e2e-playwright.js` for hex pane update on observation switch

E2E assertion added: `test-e2e-playwright.js:1794`

## Scope

- Historical observations: raw_hex stays NULL, UI falls back to transmission raw_hex silently
- No backfill, no path_json reconstruction, no frontend changes

Closes #881

PR #891, opened 2026-04-22
fix: compute hex-dump byte ranges client-side from per-obs raw_hex

Kpa-clawbot:
## Symptom
The colored byte strip in the packet detail pane is offset from the labeled byte breakdown below it. Off by N bytes where N is the difference between the top-level packet's path length and the displayed observation's path length.

## Root cause
Server computes `breakdown.ranges` once from the top-level packet's raw_hex (in `BuildBreakdown`) and ships it in the API response. After #882 we render each observation's own raw_hex, but we keep using the top-level breakdown — so a 7-hop top-level packet shipped "Path: bytes 2-8", and when we rendered an 8-hop observation we coloured 7 of the 8 path bytes and bled into the payload.

The labeled rows below (which use `buildFieldTable`) parse the displayed raw_hex on the client, so they were correct — they just didn't match the strip above.

## Fix
Port `BuildBreakdown()` to JS as `computeBreakdownRanges()` in `app.js`. Use it in `renderDetail()` from the actually-rendered (per-obs) raw_hex.

## Test
Manually verified the JS function output matches the Go implementation for FLOOD/non-transport, transport, ADVERT, and direct-advert (zero hops) cases.

Closes nothing (caught in post-tag bug bash).


Kpa-clawbot (commented):
## 🔍 The Simplifier — Review of PR #891

**APPROVED. ZERO ITEMS.**

This PR does exactly what good engineering looks like: **move the computation to where the data is, delete the dead code.** Let me walk through what I checked.

### computeBreakdownRanges() — Is it as simple as it can be?

Yes. It is a straight port of the Go `BuildBreakdown()`. No cleverness, no unnecessary abstraction, no "let me make this generic for future use." It walks the packet bytes linearly, pushes range objects, bails early when data runs out. The nesting for ADVERT sub-fields is inherent complexity — the protocol *has* nested optional fields controlled by flag bits. You can not simplify what is actually complex.

### Placement in app.js — right place or lazy place?

Right place. `app.js` is where all shared protocol-level helpers live (`isTransportRoute`, `getPathLenOffset`, `routeTypeName`). This function is protocol decoding logic used by the renderer. It belongs next to the other protocol helpers, not buried in `packets.js` where it would be untestable without loading the full page module.

### Naming: computeBreakdownRanges

The name is fine. It computes breakdown ranges. `deriveByteRanges` is vaguer (what byte ranges? for what?). `makeHexRanges` confuses hex characters with byte positions. The current name says what it does: compute the breakdown (labeled byte sections) as ranges. Ship it.

### Tests — specs or line-exercisers?

These read like specs. Each test has a clear scenario name (`hash_size=3, 2 hops × 3 bytes`), constructs a specific packet, and asserts specific byte positions. The hash_size=3 and hash_size=4 tests are particularly good — they prove the `(pathByte >> 6) + 1` math actually works for all four possible hash sizes. The truncated-path and NaN-safety tests cover real edge cases, not fantasy ones.

### Dead code cleanup — thorough?

Thorough.
- **Go side:** `BuildBreakdown`, `Breakdown`, `HexRange` types — gone from `decoder.go`. 8 tests + 3 helpers (`rangeLabels`, `equalLabels`, `assertRange`) — gone from `decoder_test.go`. `repeatHex` survives correctly — still used by other tests.
- **API response:** `Breakdown` field removed from `PacketDetailResponse` in `types.go`. `BuildBreakdown(rawHex)` call removed from `routes.go`.
- **Frontend:** All `data.breakdown` references removed from `packets.js`. Zero references to `.breakdown` remain in the branch (verified via grep).
- **Stale comment:** Line 2491 says "Single fetch — gets packet + observations + path + breakdown." The server no longer returns `breakdown`. This is a **stale comment** but it is cosmetic — the word "breakdown" in a comment does not affect behavior. Not blocking.

### createColoredHexDump callers

Only called once, in `packets.js` line 2030, with the ranges from `computeBreakdownRanges`. No other callers exist anywhere in `public/`. No shape change — same `{start, end, label}` objects the function always consumed.

### Comments — WHY or WHAT?

The JSDoc on `computeBreakdownRanges` explains WHY it exists (per-obs raw_hex can differ from top-level). The inline comment in `renderDetail` explains WHY the ranges are computed from `effectivePkt.raw_hex` (single source of truth, post-#882 accuracy). No "increment offset by 4" restatements. Good.

### No external consumers

No TypeScript types, no JSON schemas, no mobile clients, no OpenAPI spec. The `breakdown` field was only consumed by `packets.js` in this codebase. Clean removal.

---

**Net result:** +65 JS (function + tests), −206 Go (dead code + dead tests), API response shrinks by one field. The bug is fixed by construction — ranges are now derived from the same bytes being displayed. This is how you delete your way to correctness.

Issue #844, opened 2026-04-21
bug: Packet Byte Breakdown — hop offsets and values don't match raw_hex (uses aggregated pathHops instead of per-observer path)


Kpa-clawbot:
## Symptom
Packet detail → "Packet Byte Breakdown": the rendered hop values, hop count, and highlighted byte offsets don't match the raw hex bytes.

## Repro
Packet `a175df5914d38d9e` on staging, observed by "Key Route Inspector-BRK". Detail dialog shows:
- "Path Length: 0x42 — hash_size=2 bytes, **hash_count=2**"
- Then renders **"Path (7 hops)"** with hops at offsets 2, 4, 6, 8, 10, 12, 14
- Hop 0 = `41B1` (correct)
- Hop 1 claims `5EB0` at offset 4, but raw bytes 4-5 = `27D7`
- Hops 2-6 (`1000 2DD2 52F8 9535 762B`) don't exist anywhere in raw_hex
- Highlighted "Path" region in the colored hex viewer extends 14 bytes; raw path is actually 4 bytes (the green region overlaps into the pubkey)

## Raw vs frontend
- `raw_hex` (verified via `/api/packets/{hash}`): `1142 41B127D7 C0DEDAD4...980E4 ...` → header + path-length + **2 hops × 2B = `41B1` `27D7`** + 32B pubkey
- Backend `breakdown.ranges` correctly says `{start:2, end:5, label:"Path"}` (4 bytes)
- Frontend in `public/packets.js:2102-2110` ignores `breakdown.ranges` and computes path span as `pathHops.length × hashSize = 7 × 2 = 14`
- `pathHops` comes from `getParsedPath(p)` which appears to use the **aggregated/longest-observed path** (`_parsedPath` from packet record), not the per-observer `observations[i].path_json`

## Fix direction
1. Drive the byte-breakdown hop count from `path_len & 0x3F` (firmware truth: `getPathHashCount()`), not from `pathHops.length`. See `Packet.h:79-83`.
2. Use the per-observer `observations[i].path_json` when the dialog is opened from a specific observer row.
3. OR consume the backend's already-correct `breakdown.ranges` directly instead of reconstructing offsets client-side.

## Severity
Medium — packet inspector is wrong for ANY packet whose aggregated path is longer than the path bytes carried in raw_hex (i.e. most flood adverts received by multiple observers). Header/payload sections are correct, so signature/pubkey/flags are still trustworthy.

## Out of scope
[...]

PR #893, opened 2026-04-22
feat: replace dark mode button with Material Design toggle switch

emuehlstein:
## Summary

Replaces the single-icon dark mode button (☀️) with a Material Design-style toggle switch using sun/moon icons.

## Motivation

The current dark mode button uses a single emoji (☀️/🌙) that swaps on click. This has a UX ambiguity problem: does ☀️ mean "click for light mode" or "you're currently in light mode"? Users must click to find out.

Per [Material Design 3 Switch guidelines](https://m3.material.io/components/switch/overview), switches are the recommended component for binary settings because the toggle position visually communicates the current state. A pill-style toggle with sun/moon icons eliminates the ambiguity — the slider position shows which mode is active.

## Changes

**`public/index.html`** — Replace `<button>` with `<label>` containing a hidden checkbox, track, thumb, and sun/moon icon spans. The `id="darkModeToggle"` is preserved on the outer label for backward compatibility with any external references.

**`public/style.css`** — Add `.theme-toggle` component styles: track, thumb, icons with CSS transitions. All colors use CSS variables (`--border`, `--accent`, `--nav-text`) per project theming conventions. No hardcoded hex values.

**`public/app.js`** — Listen on `change` event of the checkbox instead of `click` on the button. Includes fallback: if `darkModeCheckbox` element is missing (e.g., older cached HTML), the original button click handler still works.

## Design Details

- **Track**: 46×24px rounded pill, uses `--border` color (unchecked) / `--accent` (checked)
- **Thumb**: 18px circle, slides 22px on toggle, uses `--nav-text` for contrast
- **Icons**: Sun (☀️) and Moon (🌙) at 10px, crossfade on toggle via opacity transition
- **Transitions**: 200ms ease on all moving parts
- **Touch target**: 44×44px minimum (per WCAG/M3 guidelines)
- **Accessibility**: `aria-label="Toggle dark mode"` on the label, checkbox is `aria-hidden="true"` (decorative, the label is the accessible element)

## Testing

- `npm test` passes (596/596 on frontend helpers, 62/62 on packet-filter, 29/29 on aging). One pre-existing failure on #844 (byte breakdown hop count) — unrelated.
- No new logic paths that require unit tests (this is a CSS/HTML component swap with a trivial JS event listener change).
- Backward-compatible: if the checkbox element is missing, falls back to original button click handler.

## Browser Validation

Tested on dev-scope.chicagooffline.com (Chromium). Toggle appears in nav bar, slides between sun/moon, persists theme to localStorage, respects existing `meshcore-theme` saved preference on page load.

## No New Dependencies

Pure CSS + vanilla JS. No build step changes. No new npm packages.

### DIFF
--- a/public/app.js
+++ b/public/app.js
@@ -542,10 +542,11 @@ window.addEventListener('DOMContentLoaded', () => {
 
   // --- Dark Mode ---
   const darkToggle = document.getElementById('darkModeToggle');
+  const darkCheckbox = document.getElementById('darkModeCheckbox');
   const savedTheme = localStorage.getItem('meshcore-theme');
   function applyTheme(theme) {
     document.documentElement.setAttribute('data-theme', theme);
-    darkToggle.textContent = theme === 'dark' ? '🌙' : '☀️';
+    if (darkCheckbox) darkCheckbox.checked = theme === 'dark';
     localStorage.setItem('meshcore-theme', theme);
     // Re-apply user theme CSS vars for the correct mode (light/dark)
     reapplyUserThemeVars(theme === 'dark');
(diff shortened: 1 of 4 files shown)


emuehlstein:
currently live at https://dev-scope.chicagooffline.com/

efiten (changes requested):
## Review — PR #893

One must-fix before merge.

**Hardcoded rgba in style.css:**
`box-shadow: 0 1px 3px var(--shadow, rgba(0,0,0,0.3));`
The rgba fallback is a hardcoded color — violates the repo CSS-vars rule. Define `--shadow` in the theme variable blocks and drop the inline fallback.

**Nit:** The `stopPropagation` call on the label click with comment 'Prevent click from bubbling to any legacy handler' is a no-op in the checkbox path (no legacy handler is active). Can be removed.

Everything else is clean: other CSS vars are correct, checkbox state drives applyTheme correctly, E2E fallback handles both layouts.

PR #902, opened 2026-04-23
feat(ingestor): document and test ws:// / wss:// WebSocket MQTT broker support

emuehlstein:
## Summary

CoreScope's ingestor already supports WebSocket MQTT connections today — `paho.mqtt.golang` v1.5.0 handles `ws://` and `wss://` natively via gorilla/websocket. However this support was **undocumented, untested, and had a TLS gap** for `wss://` connections.

This PR closes those gaps without any breaking changes.

### `cmd/ingestor/config.go`
- Added godoc comment to `ResolvedSources()` explaining all four supported schemes and which ones require translation vs. pass-through
- `ws://` and `wss://` explicitly documented as native paho schemes requiring no mapping

### `cmd/ingestor/main.go`
- Extended TLS config to cover `wss://` in addition to `ssl://`
- Before: `wss://` connections would use paho's default TLS (no explicit `tls.Config` set), which works for valid certs but doesn't apply the same predictable setup as `ssl://`
- After: both `ssl://` and `wss://` get `tls.Config{}` (system CA pool), matching behavior; `rejectUnauthorized: false` still works for self-signed certs on both schemes

### `cmd/ingestor/config_test.go`
Two new tests:
- `TestResolvedSourcesSchemeMapping`: validates all six scheme variations (`mqtt://`, `mqtts://`, `tcp://`, `ssl://`, `ws://`, `wss://`) including paths like `wss://host/mqtt`
- `TestLoadConfigWSSource`: full round-trip of a dual-source config (TCP + wss:// with username/password), verifies scheme unchanged through `LoadConfig` and `ResolvedSources`

### `config.example.json`
- Added `wsmqtt` example entry showing `wss://` with username/password
- Updated `_comment_mqttSources` to enumerate all supported schemes: `mqtt://`, `mqtts://`, `ws://`, `wss://`

## Motivation

We run [meshcore-mqtt-broker](https://github.com/andrewjfreyer/meshcore-mqtt-broker) (a WebSocket MQTT bridge with JWT auth) alongside Mosquitto, and subscribe to both via `mqttSources`. The dual-source config works in production but nothing in the docs or example config made this discoverable for other operators.

## Testing

```
cd cmd/ingestor && go test ./...
ok    github.com/corescope/ingestor  1.568s
```

All existing tests pass. Two new tests added.

## No breaking changes

- Existing configs: no change in behavior
- `ws://` / `wss://` configs that were already working: same behavior + explicit TLS setup for `wss://`

### DIFF
--- a/cmd/ingestor/config.go
+++ b/cmd/ingestor/config.go
@@ -162,15 +162,24 @@ func LoadConfig(path string) (*Config, error) {
 }
 
 // ResolvedSources returns the final list of MQTT sources to connect to.
+//
+// Scheme mapping:
+//
+//	mqtt://  → tcp://   (paho plain TCP)
+//	mqtts:// → ssl://   (paho TLS over TCP)
+//	ws://               (paho WebSocket — passed through, no mapping needed)
+//	wss://              (paho WebSocket TLS — passed through, no mapping needed)
 func (c *Config) ResolvedSources() []MQTTSource {
 	for i := range c.MQTTSources {
-		// paho uses tcp:// and ssl:// not mqtt:// and mqtts://
+		// paho uses tcp:// and ssl:// for plain MQTT; ws:// and wss:// are accepted natively.
 		b := c.MQTTSources[i].Broker
 		if strings.HasPrefix(b, "mqtt://") {
 			c.MQTTSources[i].Broker = "tcp://" + b[7:]
 		} else if strings.HasPrefix(b, "mqtts://") {
 			c.MQTTSources[i].Broker = "ssl://" + b[8:]
 		}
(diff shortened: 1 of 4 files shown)


emuehlstein:
testing at scope.chicagooffline and dev-scope.chicagooffline.com

efiten (approved):
Clean documentation + coverage PR. The godoc on `ResolvedSources` is accurate, the inline comment clarifying ws/wss passthrough is useful, and the test matrix (8 scheme cases + full config round-trip) is thorough. No production logic change. Approving.

efiten:
PR is approved and ready to merge but has a merge conflict. Please rebase on `master` and this can go in.

Issue #20, opened 2026-03-19
Observers table needs horizontal scroll wrapper on mobile


Kpa-clawbot:
## Issue
`.observers-page` has no `overflow-x: auto` wrapper. The table has 7 columns and no responsive override — it will overflow the viewport on phones without any scrolling ability.

`.spark-bar` has fixed `width: 100px` that doesn't shrink, contributing to overflow.

## Fix
1. Wrap table in `<div style="overflow-x:auto">`
2. Add `min-width` to table so it scrolls rather than wrapping badly
3. Consider hiding low-priority columns on mobile

## Severity
Minor

Issue #11, opened 2026-03-19
VCR buttons use emoji-only labels with no aria-label


Kpa-clawbot:
## Issue
VCR control buttons in `live.js` ~L310-315 use emoji characters (⏪, ⏸, ▶, etc.) as their only content. Screen readers will announce meaningless characters like "black left-pointing double triangle" or nothing.

Sound toggle (~L324) has `title` but no `aria-label` and no `aria-pressed` state.

## Fix
Add `aria-label` to each VCR button:
```html
<button aria-label="Rewind">⏪</button>
<button aria-label="Pause" aria-pressed="false">⏸</button>
<button aria-label="Live">LIVE</button>
```
Add `aria-label` and `aria-pressed` to sound toggle. Update pressed state on click.

## Severity
Critical — VCR controls invisible to screen readers

PR #915, opened 2026-04-27
feat: scoped/unscoped transport-route statistics (#899)

efiten:
@
## What this PR does

Implements region-scoped transport-route packet tracking with two sub-features:

### Feature 1 — Scope statistics (`scope_name`)
- At ingest, transport-route packets (route_type 0/3) with Code1 != `0000` are HMAC-matched against configured `hashRegions` keys (mirroring the `hashChannels` pattern). Matched region name (or `""` for unknown) stored in new `transmissions.scope_name` column via migration `scope_name_v1`.
- New `GET /api/scope-stats?window=` endpoint (1h/24h/7d, 30s server-side TTL) returning transport totals, scoped/unscoped counts, per-region breakdown, and time-series.
- New **Scopes** tab in Analytics with summary cards, per-region table, and two-line SVG chart. Auto-refreshes every 60s.

### Feature 2 — Node default scope (`default_scope`)
- Per-node `default_scope` column on `nodes`/`inactive_nodes` (migration `nodes_default_scope_v1`) tracks the most recently matched region for each node, derived from transport-scoped ADVERT packets.
- `GET /api/nodes` response includes `default_scope` field when column is present.
- Node detail panel displays the default scope badge.
- Async startup backfill (`BackfillDefaultScopeAsync`) populates the column for nodes with pre-existing ADVERT data.

### Config
Add `hashRegions` to `config.json` (see `config.example.json`). One entry per region name (with or without leading `#`).
@


Kpa-clawbot (changes requested):

Reviewing PR #915 cold against the diff from `gh pr diff` only. Verdict: **request changes** — significant scope drift not described in the PR body, several tautological tests, unchecked migration errors, and a couple of project-rule violations.

### Scope mismatch with PR description (highest concern)

The PR body says:
> DB migration `scope_name_v1` adds the column; scope matching runs only on new packets

But the diff also ships an entirely separate, undocumented feature:

- `cmd/ingestor/db.go` (around the second migration block) adds **migration `nodes_default_scope_v1`** which adds a `default_scope` column to both `nodes` and `inactive_nodes` AND runs a one-shot `UPDATE … FROM transmissions … WHERE EXISTS (…)` backfill across every existing node row.
- `cmd/ingestor/db.go` adds an `UpdateNodeDefaultScope(pubkey, scope)` method.
- `cmd/ingestor/main.go` calls `UpdateNodeDefaultScope` on every ADVERT whose scope matched.
- The plan doc (`docs/superpowers/plans/2026-04-23-scope-stats.md`) Self-Review explicitly says: *"Feature 3 (excluded — firmware limitation) ✅ noted in spec"* — yet the code half-implements Feature 3 anyway.
- No API endpoint, no scan path, no frontend code surfaces `default_scope`. The column is written but never read. Dead/undocumented feature.

Either remove this code from the PR, or describe it in the PR body and add API/UI exposure + tests.

### Must-fix items

1. `cmd/ingestor/db.go` migration `nodes_default_scope_v1`: unannounced scope (see above) — drop or document + expose.
2. `cmd/ingestor/db.go` migration `nodes_default_scope_v1` UPDATE is O(nodes × transmissions) at startup with no perf justification (AGENTS.md rule 0 hard requirement for hot/scale paths). Add an index hint, scope to recent rows, or justify.
3. `cmd/ingestor/db.go` `UpdateNodeDefaultScope` only touches `nodes`, not `inactive_nodes`, even though the migration added `default_scope` to both. Inconsistent.
4. `cmd/ingestor/main.go` `UpdateNodeDefaultScope` is called on every matching ADVERT receipt with no "value unchanged" short-circuit — repeated UPDATEs for the same (node, scope). Hot path, MQTT firehose.
5. `cmd/ingestor/db.go` `applySchema` migrations: every `db.Exec(...)` for ALTER/CREATE INDEX/INSERT is unchecked. If ALTER fails for an unexpected reason the migration still gets marked done. Capture and log err per existing convention.
6. `docs/superpowers/plans/2026-04-23-scope-stats.md` File Map and Self-Review reference a `BackfillScopeNames` method and "Backfill goroutine: Runs in background" — neither exists in shipped code. Plan is stale vs. implementation.
7. `cmd/ingestor/db_test.go` `TestScopeNameMigration` only PRAGMAs the column's existence — no behavior assertion (no insert, no read-back, no index check). Tautology — doesn't gate regressions.
8. `cmd/server/routes_test.go` `TestHandleScopeStats` only asserts `Window == "24h"` and that the slices are non-nil. Doesn't seed data or assert any count. Add at least one positive count assertion against seeded rows.
9. `cmd/ingestor/main_test.go` `TestBuildPacketDataScopeMatching` and `TestMatchScope` build the expected Code1 using the same loop the implementation uses — they mirror the impl rather than asserting against an external/firmware-derived reference vector. At minimum, freeze one literal known-answer vector (region name + payload + expected Code1 hex) so a future refactor can't silently change the algorithm.
10. `cmd/server/db.go` `GetScopeStats`: `route_type IN (0, 3)` is a magic number repeated 3×. Lift to a named constant or shared SQL fragment.
11. `docs/api-spec.md` /api/scope-stats: missing the 500 response shape and a caveat that for the first 1h/24h/7d window after the migration runs on an existing deployment, `unscoped` is inflated because all pre-existing transport-route rows have `scope_name = NULL` and are counted as unscoped. Operators will misread the chart for days.
12. `cmd/ingestor/db.go` `scopeNameForDB(data)` returns `interface{}` to encode the NULL/empty-string trichotomy. A `*string` (or two helpers) is type-safer and avoids `interface{}` shenanigans at the callsite.
13. `cmd/ingestor/decoder.go` adds `PayloadRaw []byte` to the public `DecodedPacket` struct purely so `BuildPacketData` can pass it to `matchScope`. Minor surface-area pollution — `matchScope` could take the relevant slice as a parameter without persisting it on the struct, or the field could be unexported (`payloadRaw`) since `json:"-"` already hides it.
14. `public/analytics.js` `renderScopesTab`: every window button click runs `el.innerHTML = ...` for a full DOM rebuild. AGENTS.md rule 0 hard rule: *"No full DOM rebuilds. Diff or virtualize."* For a small panel this matters less, but the rule is the rule — at minimum cache the static frame and only re-render data sections.
15. `public/analytics.js` `renderScopesTab`: `padB = 24` is declared and never used in the SVG layout — dead code.
16. `public/analytics.js` `renderScopesTab`: chart only renders if `timeSeries.length > 1`, otherwise the chart area is silently empty. Add an explicit "Insufficient data points to chart" message for the 1-bucket case.
17. `cmd/ingestor/db.go` migration `nodes_default_scope_v1` ordering uses `ORDER BY t.first_seen DESC` to pick the latest advert; works only because `first_seen` is per-hash insertion time. Add a one-line comment explaining the semantics so a future maintainer doesn't "fix" it to ASC.
18. `cmd/ingestor/main.go` `loadRegionKeys`: silent dedup of duplicates and silent drop of empty entries — no warning logs. Operators won't know a typo'd duplicate was ignored.
19. `cmd/ingestor/main.go` `matchScope`: linear scan over `regionKeys` per packet (HMAC per region per packet). For O(20) regions this is fine; for larger deployments it's a per-packet cost on the hot ingest path. Worth a comment documenting the expected `len(regionKeys)` ceiling, or cache by `(payloadType, payloadRaw)` if that's realistic.
20. `cmd/ingestor/db.go` `_migrations` migration uses `row.Scan(&migDone) != nil` to detect "not yet applied" — same idiom used elsewhere in the file, but worth confirming this distinguishes `sql.ErrNoRows` from a real scan error. A real scan error currently triggers re-running the migration body (which is mostly idempotent, but still hides a real DB issue).

### Out of scope (file as separate issues)

1. The unscoped-inflation caveat for backfilled deployments (item 11 above) could be properly fixed with a one-shot reclassification job that walks existing transmissions and computes scope_name for them — out of scope here, file as a follow-up enhancement issue.
2. If `default_scope` on nodes is intended to ship, it needs its own PR with API + frontend + tests, plus a clear product story for what the UI does with it.

### Approval criteria

- Drop the `nodes_default_scope_v1` migration + `default_scope` plumbing (or document + expose it), AND
- Fix the tautological tests (#7, #8, #9), AND
- Address the unchecked migration errors (#5), AND
- Sync the plan doc with shipped code (#6), AND
- Add the API-spec caveat about pre-migration NULL semantics (#11).

The remaining items are correctness/polish must-fix but the five above are the ones I'd block on.

efiten:
### Code review

Found 2 issues:

1. **"Scopes" tab button is inserted after "Prefix Tool", violating the documented tab-ordering convention.** The comment immediately above the tab list states "before Prefix Tool (utility tabs trail)" — meaning data tabs go before the Prefix Tool utility tab. The Roles tab (added in #1085) correctly follows this convention. The Scopes button should be inserted before the Prefix Tool button.

https://github.com/Kpa-clawbot/CoreScope/blob/953874ec379db005f8f826e70fb196869d4eb8b2/public/analytics.js#L125-L129

2. **`renderScopesTab` has no auto-refresh.** The Scopes tab displays time-varying data (scoped/unscoped counts, time-series chart) but has no `setInterval` to refresh while the tab is active. The Roles tab sets up `_rolesRefreshTimer` with a 60s interval and cleans it up in `destroy()`. The same pattern is missing here — operators monitoring the Scopes tab will see stale data until they interact.

https://github.com/Kpa-clawbot/CoreScope/blob/953874ec379db005f8f826e70fb196869d4eb8b2/public/analytics.js#L3780-L3786

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

efiten:
All 20 must-fix items from round-1 review are addressed. Summary of what changed:

**Migration safety**
- `scope_name_v1` and `nodes_default_scope_v1`: all `db.Exec` calls now check and log errors; `INSERT INTO _migrations` returns a wrapped error on failure
- Removed the O(nodes × transmissions) startup UPDATE from `nodes_default_scope_v1`; replaced with `BackfillDefaultScopeAsync` (goroutine, idempotency check via `_migrations`, panic recovery — same pattern as `BackfillPathJSONAsync`)

**`default_scope` feature — fully exposed, not dropped**
- `UpdateNodeDefaultScope`: now also updates `inactive_nodes`; short-circuits when stored value matches; guard changed from `ScopeName != ""` to `IsTransportScoped`
- `GET /api/nodes` response includes `default_scope` field when column is present (`hasDefaultScope` schema detection, `nodeSelectCols()` helper, `scanNodeRow` converted to `DB` method)
- Node detail panel in `live.js` shows Scope row (null → `—`, empty → "unknown scope", named → `#region` in accent color)
- API spec documents `default_scope` field, 500 response shape for missing column, and pre-migration NULL inflation caveat

**Type cleanup**
- `scopeNameForDB` returns `*string` (was `interface{}`)
- `PayloadRaw` unexported to `payloadRaw`

**Frontend (`renderScopesTab`)**
- Tab order: Scopes now appears before Prefix Tool
- Replaced raw `fetch()` with `api()` (goes through the caching layer)
- Static DOM frame written once; window changes update only inner elements
- Auto-refresh every 60s (same pattern as Roles tab)
- Dead `padB = 24` variable removed; "Insufficient data" message when `timeSeries.length ≤ 1`

**Tests**
- `TestScopeNameMigration`: insert + read-back (was tautological column-existence check)
- `TestMatchScope`: fixed known-answer vector (`belgiumCode1 = "4A75"` for `#belgium`/type=5/`hello`)
- `TestHandleScopeStats`: seeded 4 rows, asserts all counts (`transportTotal`, `scoped`, `unscoped`, `unknownScope`, `byRegion`)

**`routeTypeTransportSQL` constant**
- Replaces three inline `route_type IN (0, 3)` literals in `GetScopeStats`

**`loadRegionKeys` / `matchScope`**
- Warn logs on empty/duplicate `hashRegions` entries
- Ceiling comment on `matchScope` (≤ 50 regions; beyond that, consider indexed lookup)

**Housekeeping**
- `docs/superpowers/` files removed from PR diff (local-only planning files)
- PR description updated to document both features

efiten (commented):
Ready for re-review.

Kpa-clawbot (changes requested):
## Consolidated review — PR #915

Two parallel reviewers (adversarial + protocol context). **Verdict: NEEDS-WORK.**

### MUST-FIX

1. **Feature 3 (`default_scope`) ships with zero tests.** `UpdateNodeDefaultScope` short-circuit + `inactive_nodes` mirror, `BackfillDefaultScopeAsync` idempotency, `/api/nodes` field presence, node detail panel rendering branches — all 4 surfaces untested. Claim from commit `1361f8a` is unverified by tests. Need at minimum: one Go test for `UpdateNodeDefaultScope` (insert → call → assert both tables updated → call with same value → assert no extra write) + one for `/api/nodes` response field presence.

2. **3 tautological tests** (`TestMatchScope`, `TestBuildPacketDataScopeMatching`, `TestLoadRegionKeys` in `cmd/ingestor/main_test.go`) recompute the production HMAC/SHA formula in the test, then assert `matchScope` produces the same answer. A wrong implementation that flips endianness or HMAC input order on both sides would pass. The `belgiumCode1="4A75"` fixed vector is the only non-tautological assertion. Promote the fixed vector to primary case; replace self-computed assertions with hardcoded hex literals.

3. **`TestGetScopeStats` + `TestHandleScopeStats` bypass the schema-migration path** by hand-creating `transmissions` and manually setting `db.hasScopeName = true`. The actual `OpenRW`/`detectScopeName` flag-setting logic isn't exercised. Add one test that opens the store via the real path and asserts `hasScopeName` ends true post-migration.

### Perf regression risk vs recent merges

4. **`BackfillDefaultScopeAsync` may contend with `Load()` background-loader (#1187) for the single cached-RW sqlite connection.** Both run async at startup, both want write access. The PR's async-backfill replaces an O(n²) startup UPDATE — good — but doesn't coordinate with the hot-startup background loader landed since this PR was last reviewed. Worth verifying: does the backfill wait for `backgroundLoadDone` (or some equivalent ready signal) before issuing UPDATEs? Or does it race the chunk-loader for the WAL writer? If it races, expect SQLITE_BUSY storms during the first 5-10 minutes after restart on a large DB.

### Prior CHANGES_REQUESTED items (20 originally flagged): 19 ✅ landed, 1 ⚠️ ([20] `BackfillDefaultScopeAsync`'s trailing `INSERT INTO _migrations` + `s.db.QueryRow().Scan == nil` idiom inherits the same pattern flagged in r1 #20 — non-blocking sweep).

### Out-of-scope (flag, don't block)
- Reclassification job to backfill `scope_name` for pre-migration rows (closes the unscoped-inflation gap fully — separate issue).
- A few upstream-merge carryovers from #1249-#1251 may have ridden along on this PR's merges — confirm via `git log --no-merges` between this PR's base and HEAD, split out if net-new here.

Once the 4 must-fix items + the perf-coordination concern are addressed this is ready to merge.

efiten:
Round-2 must-fix items addressed in commit `d989d6e`:

1. **Feature 3 tests** — Added `TestUpdateNodeDefaultScope`: inserts a node into both `nodes` and `inactive_nodes`, calls `UpdateNodeDefaultScope`, asserts both tables written; second call with same value is a no-op (short-circuit); third call with new value updates both tables.

2. **Tautological tests replaced with fixed known-answer vectors:**
   - `TestMatchScope`: self-computed HMAC replaced with two hardcoded vectors (`#test`→`2AB5`, `#belgium`→`4A75`). A wrong algorithm that produces consistent wrong results on both sides now fails.
   - `TestBuildPacketDataScopeMatching`: in-test packet construction replaced with hardcoded `rawHex = "142AB500000068656C6C6F"` and pre-computed key.
   - `TestLoadRegionKeys`: `sha256.Sum256("#belgium")` in-test replaced with hardcoded hex literal `7085b78ed010599094f8c8e7d1aa0e27`.

3. **Schema migration path tested** — Added `TestDetectSchemaScopeName` (`cmd/server/db_test.go`): creates a file-based SQLite with `scope_name` and `default_scope` columns, opens via `OpenDB()`, asserts `hasScopeName == true` and `hasDefaultScope == true`. Also asserts both flags false when columns are absent.

4. **BackfillDefaultScopeAsync concurrency** — Added comment explaining that `SetMaxOpenConns(1)` serializes all writes including this backfill goroutine; no additional locking is needed. Server is read-only post-#1289, so cross-process write contention is impossible.

CI triggered. Ready for re-review.

PR #917, opened 2026-04-27
fix(store): apply retentionHours cutoff in Load() to prevent OOM on cold start

efiten:
## Problem

`Load()` loaded all transmissions from the DB regardless of `retentionHours`, so `buildSubpathIndex()` processed the full DB history on every startup. On a DB with ~280K paths this produces ~13.5M subpath index entries, OOM-killing the process before it ever starts listening — causing a supervisord crash loop with no useful error message.

## Fix

Apply the same `retentionHours` cutoff to `Load()`'s SQL that `EvictStale()` already uses at runtime. Both conditions (`retentionHours` window and `maxPackets` cap) are combined with AND so neither safety limit is bypassed.

Startup now builds indexes only over the retention window, making startup time and memory proportional to recent activity rather than total DB history.

## Docs

- `config.example.json`: adds `retentionHours` to the `packetStore` block with recommended value `168` (7 days) and a warning about `0` on large DBs
- `docs/user-guide/configuration.md`: documents the field and adds an explicit OOM warning

## Test plan

- [x] `cd cmd/server && go test ./... -run TestRetentionLoad` — covers the retention-filtered load: verifies packets outside the window are excluded, and that `retentionHours: 0` still loads everything
- [x] Deploy on an instance with a large DB (>100K paths) and `retentionHours: 168` — server reaches "listening" in seconds instead of OOM-crashing
- [x] Verify `config.example.json` has `retentionHours: 168` in the `packetStore` block
- [x] Verify `docs/user-guide/configuration.md` documents the field and warning


efiten:
## Playwright failure is unrelated to this PR

The failing test (`Analytics Neighbor Graph tab renders canvas and stats`) is caused by **fixture DB staleness**, not by the `retentionHours` change in this PR.

**What the CI logs show:**

The server built the neighbor graph successfully on startup:
```
[neighbor] no persisted edges found, will build in background...
[neighbor-prune] auto-prune enabled: edges older than 5 days
[neighbor] persisted 317 edges
[neighbor] graph build complete
```

317 edges were built — but the auto-prune (`edges older than 5 days`) then wiped them all before the Playwright test reached that step. The fixture DB's edge timestamps have aged past the 5-day cutoff.

**Timeline confirming this is not our change:**
- 08:33 run (different PR, same fixture DB) — Neighbor Graph ✅ passed
- 10:10 run (`feat/multibyte-cap-903`) — Neighbor Graph ✅ passed
- 11:22 run (this PR) — Neighbor Graph ❌ fails with 0 cards

The fixture crossed the 5-day boundary sometime between 10:10 and 11:22 today. Every subsequent PR will hit this until `test-fixtures/e2e-fixture.db` is regenerated.

**Why this PR cannot be the cause:** The `retentionHours` filter in `Load()` is only active when `s.retentionHours > 0`. The E2E server is started without a config file, so `retentionHours` defaults to `0` and the new code path is skipped entirely. Behaviour is identical to before this change when no retention is configured.

Kpa-clawbot (approved):
## Self-Review ✅

Verified code matches the stated intent:
- `Load()` now applies the same `retentionHours` cutoff that `EvictStale()` already uses at runtime
- Both conditions (retention window + maxPackets cap) combined with AND — correct
- Tests cover both filtered and unlimited (retentionHours=0) cases
- Docs updated in config.example.json and user guide
- E2E test fix (commit 5131a94) addresses a legitimate race condition

No Frankenstein leftovers, no duplicate logic, no stale references. Clean and focused.

---

## Independent Adversarial Review ✅

Checked:
- **SQL injection**: cutoff is `time.Now().UTC().Format(RFC3339)` — not user-controlled. Safe.
- **String comparison for timestamps**: RFC3339 is lexicographically sortable. Correct.
- **Index usage**: `idx_tx_first_seen` exists and supports the `t.first_seen >= cutoff` predicate.
- **Race conditions**: `Load()` runs once at startup before concurrent access begins. Safe.
- **Edge cases**: `retentionHours=0` loads all (tested). Very small values filter aggressively (expected, documented).
- **Consistency**: Same duration calculation pattern as `EvictStale()` (lines 2957-2958).

No issues found.

---

## Carmack (Performance) Review ✅

The fix is architecturally sound from a data-flow perspective:
- Filtering happens at the SQL layer (WHERE clause) — the expensive work (loading rows, building subpath/path-hop/distance indexes) only processes the retention window
- The `idx_tx_first_seen` index makes the cutoff filter O(log n) rather than a full table scan
- This is startup-only code (runs once) — the `fmt.Sprintf` and string building are irrelevant to runtime performance
- Memory usage becomes proportional to retention window size, not total DB history — exactly the right fix for OOM

**Verdict: merge-ready** ✅

PR #926, opened 2026-04-28
fix(ingestor): unblock startup when one of multiple MQTT sources is unreachable

### LINKED ISSUE #910: Ingestion blocked if multiple MQTT sources defined but one is unreachable on restart
If two MQTT source are defined and only one is responding the following MQTT status is reported in the logs and no ingestion occurs.

2026/04/24 22:37:31 [ingestor] MQTT [eastmesh] connected to tcp://mqtt.eastmesh.au:8001
2026/04/24 22:37:31 [ingestor] MQTT [eastmesh] subscribed to meshcore/+/+/packets

If the failing source is removed and the server is restarted then the follow is reported:
2026/04/25 03:04:53 [ingestor] Running â 1 MQTT source(s) connected
2026/04/25 03:04:53 [ingestor] MQTT [eastmesh] connected to tcp://mqtt.eastmesh.au:8001
2026/04/25 03:04:53 [ingestor] MQTT [eastmesh] subscribed to meshcore/+/+/packets

So it appears that if a failing source cannot be reached, the ingestor does not reach Running status.

If the source fails whilst the server and ingestor are running, the surviving source is still ingested. It is only on restart that this issue occurs.

efiten:
## Summary

- With `ConnectRetry=true`, paho's `token.Wait()` only returns on success — it blocks forever for unreachable brokers, stalling the entire startup loop before any other source connects
- Switches to `token.WaitTimeout(30s)`: on timeout the client is still tracked so `ConnectRetry` keeps retrying in background; `OnConnect` fires and subscribes when it eventually connects
- Adds `TestMQTTConnectRetryTimeoutDoesNotBlock` to confirm `WaitTimeout` returns within deadline for unreachable brokers (regression guard for this exact failure mode)

Fixes #910

## Test plan

- [x] Two MQTT sources configured, one unreachable: ingestor reaches `Running` status and ingests from the reachable source immediately on startup
- [x] Unreachable source logs `initial connection timed out — retrying in background` and reconnects automatically when the broker comes back
- [x] Single source, reachable: behaviour unchanged (`Running — 1 MQTT source(s) connected`)
- [x] Single source, unreachable: `Running — 0 MQTT source(s) connected, 1 retrying in background`; ingestion starts once broker is available
- [x] `go test ./...` passes (excluding pre-existing `TestOpenStoreInvalidPath` failure on master)

### DIFF
--- a/cmd/ingestor/main.go
+++ b/cmd/ingestor/main.go
@@ -114,6 +114,7 @@ func main() {
 
 	// Connect to each MQTT source
 	var clients []mqtt.Client
+	connectedCount := 0
 	for _, source := range sources {
 		tag := source.Name
 		if tag == "" {
@@ -167,11 +168,19 @@ func main() {
 
 		client := mqtt.NewClient(opts)
 		token := client.Connect()
-		token.Wait()
+		// With ConnectRetry=true, token.Wait() blocks forever for unreachable brokers.
+		// WaitTimeout lets startup proceed; the client keeps retrying in the background
+		// and OnConnect fires (subscribing) when it eventually connects (#910).
+		if !token.WaitTimeout(30 * time.Second) {
+			log.Printf("MQTT [%s] initial connection timed out — retrying in background", tag)
+			clients = append(clients, client)
+			continue
+		}
 		if token.Error() != nil {
 			log.Printf("MQTT [%s] connection failed (non-fatal): %v", tag, token.Error())
 			continue
 		}
+		connectedCount++
 		clients = append(clients, client)
 	}
 
@@ -179,7 +188,11 @@ func main() {
 		log.Fatal("no MQTT connections established — check broker is running (default: mqtt://localhost:1883). Set MQTT_BROKER env var or configure mqttSources in config.json")
 	}
 
-	log.Printf("Running — %d MQTT source(s) connected", len(clients))
+	if connectedCount < len(clients) {
+		log.Printf("Running — %d MQTT source(s) connected, %d retrying in background", connectedCount, len(clients)-connectedCount)
+	} else {
+		log.Printf("Running — %d MQTT source(s) connected", connectedCount)
+	}
 
 	// Wait for shutdown signal
 	sig := make(chan os.Signal, 1)
--- a/cmd/ingestor/main_test.go
+++ b/cmd/ingestor/main_test.go
@@ -7,6 +7,8 @@ import (
 	"path/filepath"
 	"testing"
 	"time"
+
+	mqtt "github.com/eclipse/paho.mqtt.golang"
 )
 
 func TestToFloat64(t *testing.T) {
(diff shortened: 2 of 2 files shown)


efiten:
### Code review

Found 2 issues:

1. **`len(clients) == 0` guard semantics silently changed** — timed-out clients are now appended to `clients` before any connection is established. In a single-broker setup where the broker is down, the startup no longer fatally exits; it logs "retrying in background" and continues running with zero active connections. The original guard meant "at least one connected"; it now means "at least one attempted".

https://github.com/Kpa-clawbot/meshcore-analyzer/blob/86654487db71862e0dd6d9fd46a167415adfa7ea/cmd/ingestor/main.go#L185-L190

2. **Goroutine leak on hard-failure path** — when `token.Error() != nil` (fast refusal, bad credentials, etc.), the client is discarded with `continue` but `Connect()` was already called with `ConnectRetry=true`. Paho spawns background retry goroutines on `Connect()`; without a `client.Disconnect()` call on this path those goroutines run indefinitely with no way to stop them.

https://github.com/Kpa-clawbot/meshcore-analyzer/blob/86654487db71862e0dd6d9fd46a167415adfa7ea/cmd/ingestor/main.go#L178-L182

<sub>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

Kpa-clawbot (approved):
## PR Polish — All Passes Complete

### Pass 1: Self-review ✅ PASS
- Change is minimal and well-scoped to the exact failure mode (#910)
- `WaitTimeout(30s)` correctly replaces blocking `Wait()` when `ConnectRetry=true`
- Timed-out clients still tracked so `OnConnect` fires and subscribes when broker becomes reachable
- `connectedCount` vs `len(clients)` logic is correct — fatal only fires when zero clients exist (neither connected nor retrying)
- Log messages are clear and actionable
- Test is a solid regression guard using port 1 (nothing listening) with a reasonable timeout assertion

### Pass 2: Adversarial ✅ PASS
- **Concurrency**: No new goroutines or shared state introduced. Paho handles background retry internally.
- **Resource leak**: Timed-out clients are tracked in `clients` slice and properly `Disconnect()`ed on shutdown (existing shutdown loop).
- **Edge case — all sources timeout**: Intentionally non-fatal per design (documented in PR body). Ingestor enters "0 connected, N retrying" state and begins ingesting once any broker connects.
- **Edge case — permanent auth failure after timeout**: Client retries forever silently. Pre-existing paho behavior with ConnectRetry=true, not introduced here.
- **30s timeout hardcoded**: Acceptable for now; noted as future config candidate.

### Pass 3: Domain Expert (DJB concurrency/reliability) ✅ PASS
- The fix correctly identifies that `token.Wait()` with `ConnectRetry=true` is an unbounded blocking call — a classic startup-ordering deadlock.
- The solution is minimal: deadline the wait, let the retry machinery handle recovery. No complex state machines or coordination added.
- `OnConnect` handler already handles subscription, so late-connecting clients work correctly without additional logic.

### Verdict: **APPROVE** — clean, minimal, correct fix with good regression test.

**Out-of-scope (follow-up):** expose `30s` connect timeout as config value.

PR #927, opened 2026-04-28
fix(customize): skip panel re-render while a text field has focus

### LINKED ISSUE #896: bug: mobile keyboard collapses after every key press on Customize window fields
In the Customize window, on mobile, when editing a text field, after every key press my keyboard is collapsed/closed. I believe this may be occuring because of an auto-save functionality after every key press.

efiten:
## Summary

- `_debouncedWrite()` was calling `_refreshPanel()` 300ms after every keystroke
- `_refreshPanel()` sets `container.innerHTML`, destroying the focused input element
- On mobile, losing the focused input collapses the virtual keyboard after each keypress

Guard the `_refreshPanel()` call so it is skipped when `document.activeElement` is inside the panel. The pipeline (`_runPipeline`) still runs immediately — CSS updates apply. Override dots update on the next natural re-render (tab switch, dark-mode toggle, panel reopen).

## Root cause

`customize-v2.js` → `_debouncedWrite()` → `_refreshPanel()` → `_renderPanel()` → `container.innerHTML = ...`

## Test plan

- [ ] New Playwright E2E test: open Customize, focus a text field, type, wait 500ms past debounce — asserts input element is still connected to DOM and focus remains inside panel
- [ ] Manual: open Customize on mobile (or DevTools mobile emulation), type in Site Name — keyboard must not collapse after each character

Fixes #896

### DIFF
--- a/public/customize-v2.js
+++ b/public/customize-v2.js
@@ -629,7 +629,11 @@
       }
       writeOverrides(delta);
       _runPipeline();
-      _refreshPanel();
+      // Skip re-render while the user is typing inside the panel — setting
+      // innerHTML would destroy the focused input and collapse the mobile keyboard.
+      if (!(_panelEl && _panelEl.contains(document.activeElement))) {
+        _refreshPanel();
+      }
     }, 300);
   }
 
--- a/test-e2e-playwright.js
+++ b/test-e2e-playwright.js
@@ -1358,6 +1358,38 @@ async function run() {
     await page.evaluate(() => localStorage.removeItem('cs-theme-overrides'));
   });
 
+  await test('Customizer v2: typing in text field does not collapse focus (re-render guard)', async () => {
+    await page.goto(BASE, { waitUntil: 'domcontentloaded' });
+    await page.waitForSelector('nav, .navbar, .nav, [class*="nav"]');
+    await page.waitForFunction(() => window._customizerV2 && window._customizerV2.initDone, { timeout: 5000 });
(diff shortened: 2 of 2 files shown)


Kpa-clawbot (approved):
## PR Polish — All Passes Complete

### Pass 1: Self-review ✅ PASS
- Guard correctly skips `innerHTML` rebuild when `document.activeElement` is inside the panel
- `_runPipeline()` still fires immediately — CSS variables update in real-time
- Override dots become stale while typing but update on next natural re-render (tab switch, panel reopen) — acceptable tradeoff
- Non-mobile flows unaffected: focus outside panel → re-render runs normally

### Pass 2: Adversarial (cross-browser/input method) ✅ PASS
- **Android Chrome / iOS Safari**: virtual keyboard stays up because focused input is never destroyed ✓
- **Desktop**: guard is harmless (no keyboard to collapse, dots update on next interaction)
- **Password managers / autofill**: typically set values without maintaining focus → re-render runs normally ✓
- **Dictation**: keeps focus in input → guard prevents collapse ✓
- **Edge case**: user clicks away within 300ms debounce → `activeElement` moves outside panel → re-render fires ✓

### Pass 3: Torvalds (code quality) ✅ PASS
- 4 lines including comment. Minimal, surgical, well-commented.
- No unnecessary abstractions. Condition is immediately readable.
- Test covers both DOM connection and focus retention after debounce window.

### Verdict: **Approve** — no must-fixes, no out-of-scope items needed.

PR #930, opened 2026-04-29
fix(paths): exclude false-positive paths from short-prefix collisions

### LINKED ISSUE #929: Paths Through: false positives from short-prefix collisions in byPathHop index
## Bug

When viewing a node's \"Paths Through\" section, some paths shown do not actually contain the target node.

## Root cause

`handleNodePaths` looks up candidate transmissions from `byPathHop` using short prefix keys:
- `prefix1` (2-char, e.g. `\"7a\"` for a node using 1-byte adverts)
- `prefix2` (4-char, e.g. `\"7acb\"`)

These short prefixes collide: `byPathHop[\"7a\"]` contains paths from **any** node whose pubkey starts with `7a`, not just the target node.

The post-filter (`resolvedPubkeyReverse` + SQL) is designed to disambiguate, but falls back conservatively when a transmission has no `resolved_path` data (`hasRev = false → inIndex = true`). Those false-positive candidates are then returned in the result.

## Fix

During the aggregation phase, `resolveHop()` is already called on every hop in every path. Add a check: if every resolved hop maps to a different node's pubkey than the target, skip the path. Keep paths where at least one hop resolves to the target, or where a hop is unresolvable (conservative).

Transmissions confirmed via the full-pubkey key in `byPathHop` or via SQL bypass this check.

efiten:
Fixes #929

## Summary

- `handleNodePaths` pulls candidates from `byPathHop` using 2-char and 4-char prefix keys (e.g. `"7a"` for a node using 1-byte adverts)
- When two nodes share the same short prefix, paths through the *other* node are included as candidates
- The `resolved_path` post-filter covers decoded packets but falls through conservatively (`inIndex = true`) when `resolved_path` is NULL, letting false positives reach the response

**Fix:** during the aggregation phase (which already calls `resolveHop` per hop), add a `containsTarget` check. If every hop resolves to a different node's pubkey, skip the path. Packets confirmed via the full-pubkey index key or via SQL bypass the check. Unresolvable hops are kept conservatively.

## Test plan
- [x] `TestHandleNodePaths_PrefixCollisionExclusion`: two nodes sharing `"7a"` prefix; verifies the path with no `resolved_path` (false positive) is excluded and the SQL-confirmed path (true positive) is included
- [x] Full test suite: `go test github.com/corescope/server` — all pass

### DIFF
--- a/cmd/server/paths_through_test.go
+++ b/cmd/server/paths_through_test.go
@@ -1,0 +1,78 @@
+package main
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/gorilla/mux"
+)
+
+// TestHandleNodePaths_PrefixCollisionExclusion verifies that paths through a node
+// sharing a 2-char prefix with another node are not returned as false positives
+// when they have no resolved_path data (issue #929).
+//
+// Setup:
+//   - nodeA (target): pubkey starts with "7a", no GPS
+//   - nodeB (other):  pubkey starts with "7a", has GPS → "7a" resolves to nodeB
+//   - tx1: path ["7a"], resolved_path NULL → false positive candidate, must be excluded
+//   - tx2: path ["7a"], resolved_path contains nodeA pubkey → SQL-confirmed, must be included
+func TestHandleNodePaths_PrefixCollisionExclusion(t *testing.T) {
+	db := setupTestDB(t)
+	recent := time.Now().Add(-1 * time.Hour).Format(time.RFC3339)
+	recentEpoch := time.Now().Add(-1 * time.Hour).Unix()
+
+	nodeAPK := "7acb1111aaaabbbb"
+	nodeBPK := "7aff2222ccccdddd" // same "7a" prefix, has GPS so resolveHop("7a") picks B
+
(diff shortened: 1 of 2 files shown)


Kpa-clawbot (approved):
## PR Polish — All 3 Passes: ✅ PASS

### Pass 1: Self-review (Frankenstein/duplication check)
**Verdict: PASS** — No duplication. The fix reuses the existing `resolveWithContext` 4-tier disambiguation system via the `resolveHop` closure. The new `containsTarget` is a lightweight post-filter, not a parallel implementation.

### Pass 2: Independent adversarial review
**Verdict: PASS** — Edge cases are handled correctly:
- Multi-hop paths: any hop resolving to target marks it valid
- `totalTransmissions++` correctly moved after the filter
- Conservative fallback for unresolvable hops (`HasPrefix`) prevents false negatives
- No race conditions in map usage across lock boundaries

### Pass 3: Domain expert — MeshCore Protocol Engineer
**Verdict: PASS** — Protocol-correct. 1-byte advert nodes sharing 2-char prefixes is a real collision scenario (birthday problem on 256 values with 50+ nodes). The fix correctly leverages neighbor affinity resolution to distinguish which "7a" node a path actually transited. SQL `resolved_path` bypass is authoritative since it stores full pubkeys at ingest time.

### Tests
All tests pass including the new `TestHandleNodePaths_PrefixCollisionExclusion`.

**No must-fixes. No out-of-scope items. Ready to merge.**

PR #933, opened 2026-04-30
feat(server): add opt-in gzip and WebSocket compression config

efiten:
## Summary

- Adds a `compression` config block to `config.json` (both fields default to `false`)
- `compression.gzip: true` — wraps the HTTP handler with a gzip middleware; compresses API responses and static assets when the client sends `Accept-Encoding: gzip`. WebSocket upgrade requests are passed through unmodified
- `compression.websocket: true` — enables `permessage-deflate` on the gorilla WebSocket upgrader

## Motivation

Operators running behind a reverse proxy that does not compress upstream responses (e.g. nginx proxying to the Go server over a separate network leg) get no compression today. This adds an opt-in way to compress at the source.

## Usage

```json
"compression": {
  "gzip": true,
  "websocket": true
}
```

## Test plan

- [x] `TestCompressionConfigDefaults` — nil config returns false for both helpers
- [x] `TestCompressionConfigExplicitFalse` — explicit false returns false
- [x] `TestCompressionConfigEnabled` — explicit true returns true
- [x] `TestGZipMiddlewareCompresses` — response is valid gzip when `Accept-Encoding: gzip`
- [x] `TestGZipMiddlewareSkipsNoAcceptEncoding` — plain response when no header
- [x] `TestGZipMiddlewareSkipsWebSocket` — WebSocket upgrade not compressed
- [x] Existing test suite passes unchanged (zero-arg `NewHub()` calls untouched)

### DIFF
--- a/cmd/server/compress.go
+++ b/cmd/server/compress.go
@@ -1,0 +1,46 @@
+package main
+
+import (
+	"compress/gzip"
+	"net/http"
+	"strings"
+)
+
+type gzipResponseWriter struct {
+	http.ResponseWriter
+	gz    *gzip.Writer
+	wrote bool
+}
+
+func (g *gzipResponseWriter) WriteHeader(code int) {
+	g.ResponseWriter.Header().Del("Content-Length")
+	g.ResponseWriter.WriteHeader(code)
+}
+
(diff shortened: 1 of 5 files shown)


efiten:
Duplicate of #934 which includes the e2e fix for the detached-element race.

PR #1164, opened 2026-05-07
perf(nodes): batch relay stats to fix O(N×M) /api/nodes regression

efiten:
## Problem

`handleNodes` enriches each repeater/room node by calling `GetRepeaterRelayInfo` and `GetRepeaterUsefulnessScore` **per node** inside a loop. `GetRepeaterUsefulnessScore` acquires `s.mu.RLock()` and then iterates **all** `byPayloadType` entries to compute the non-advert denominator — once per node.

On a deployment with ~1500 repeater/room nodes and ~145K transmissions in memory, this is **~220M iterations per `/api/nodes` request**, plus ~3000 separate lock acquisitions. Response times of 18–44 seconds have been observed in production, especially during startup backfill when write-lock contention compounds the issue.

## Fix

Add `GetRepeaterNodeStatsBatch(pubkeys []string, windowHours float64) map[string]RepeaterNodeStats` to `repeater_usefulness.go`:

- Takes **one** `s.mu.RLock()` for the entire node list
- Computes the non-advert denominator **once** (shared across all nodes)
- Snapshots `byPathHop` slice headers for all requested pubkeys under that single lock
- Processes timestamps and counts **outside** the lock

Update `handleNodes` to collect repeater/room pubkeys first, call the batch method once, and apply results.

**Complexity: O(M + N) instead of O(N × M)** per request (M = total transmissions, N = repeater nodes).

`GetRepeaterRelayInfo` and `GetRepeaterUsefulnessScore` are unchanged — they are still correct for single-node calls (e.g. `handleNodeDetail`).

## Test plan

- [ ] `go build ./cmd/server` passes
- [ ] `/api/nodes` response is correct (relay_active, relay_count_1h/24h, usefulness_score fields present for repeaters)
- [ ] No change in output for `/api/nodes/{pubkey}` (uses existing single-node methods)
- [ ] CI passes


Kpa-clawbot (changes requested):
## Independent review — PR #1164 (perf: batch relay stats)

Thanks for the perf work — the diagnosis is right, the denominator-once + single-RLock shape is the correct fix, and the `EnrichNodes`-style call site change in `routes.go` is clean. I'd ship this with the items below addressed.

### Must-fix

**1. Data race: slice headers snapshotted under RLock, elements read after RUnlock.**

In `GetRepeaterNodeStatsBatch`:

```go
s.mu.RLock()
// ...
snaps[pk] = nodeSnap{txList: s.byPathHop[key]}
// ...
s.mu.RUnlock()

// ... outside lock ...
for _, tx := range snap.txList {  // racy
    if tx != nil { uniq[tx.ID] = struct{}{} }
}
```

Only the slice *header* is copied under the lock; the backing array is shared with writers. Writers DO mutate the backing array in place:

- `removeTxFromSlice` (store.go:2886): `idx[key] = append(list[:i], list[i+1:]...)` — this **shifts elements left in the same backing array**. A reader iterating the snapshot will see torn / shifted / duplicated `*StoreTx` entries.
- `append(s.byPathHop[key], tx)` (store.go:619, 1630, 1996, 2117) — at-or-over-capacity grows in place into spare cap, which is invisible to the snapshot length but doesn't trigger a copy of existing slots, so still safe by itself; but combined with the remove path above it's not.

Compare with the existing `GetRepeaterRelayInfo` (`repeater_liveness.go`), which does `collect(txList) / collect(prefixList)` **before** `s.mu.RUnlock()` — i.e. it reads every `tx.ID / tx.PayloadType / tx.FirstSeen` it needs while still holding the read lock and only the resulting `scratch` slice escapes. Please mirror that: extract the per-node `(ts, pt, isInTxList)` tuples into a per-pubkey `[]entry` (or a single flat slice keyed by pubkey index) **inside the RLock**, then release and process. Keeps the perf win — the denominator is still computed once and the lock is still held only once across the whole node set — but stops crossing the lock boundary with mutable state.

This will also be flagged by `go test -race` on a contention-heavy fixture, which is part of why a real test (item 2) matters.

**2. No behavior tests for the new public API.**

Per `AGENTS.md` TDD rule, this is a behavior change — `GetRepeaterNodeStatsBatch` is a new public method on `PacketStore`. The "Test plan" in the PR body is `go build` + "CI passes", which are not behavior tests. Required, in `cmd/server/repeater_usefulness_test.go` (or a new `repeater_batch_test.go`):

- **Parity test**: build a `PacketStore` with a representative mix (one node with full-key hops only, one with prefix-only hops, one with both, one with no hops, one repeater that sees only adverts). Call `GetRepeaterNodeStatsBatch(pubkeys, w)`. For each pubkey, assert `result[pk].Info` equals `GetRepeaterRelayInfo(pk, w)` field-by-field and `result[pk].Score` equals `GetRepeaterUsefulnessScore(pk)`.
- **Empty input**: `GetRepeaterNodeStatsBatch(nil, 24)` and `(...[]string{}, 24)` return non-nil empty map.
- **Nodes with no `byPathHop` entries** appear in the result map with zero `Info` and zero `Score` (this is what `routes.go`'s `if stats, ok := relayStats[pk]; ok` relies on, and it currently always succeeds because the loop unconditionally writes `result[pk]` — that's the right behavior, please pin it with a test so a future refactor can't regress it).
- **`-race` test** that hammers `GetRepeaterNodeStatsBatch` from one goroutine while another calls a mutating store method (e.g. an upsert that triggers `addTxToPathHopIndex` / `removeTxFromPathHopIndex`). With the current snapshot-then-release shape this should fail; with the fix from item 1 it should pass.

The red→green sequence on the parity test should also be visible in the commit history per `AGENTS.md` "TDD IS MANDATORY".

**3. `uniq` map is dead work.**

```go
uniq := make(map[int]struct{}, len(snap.txList)+len(snap.prefixList))
for _, tx := range snap.txList   { if tx != nil { uniq[tx.ID] = struct{}{} } }
for _, tx := range snap.prefixList{ if tx != nil { uniq[tx.ID] = struct{}{} } }
entries := make([]entry, 0, len(uniq))
seen := make(map[int]bool, len(uniq))
collect := func(list []*StoreTx) { /* ... uses `seen`, not `uniq` ... */ }
collect(snap.txList); collect(snap.prefixList)
```

`uniq` is built only to size `entries` / `seen`, then `seen` re-does the exact same dedup. Either drop `uniq` and oversize `entries`/`seen` with `len(snap.txList)+len(snap.prefixList)` (the existing `GetRepeaterRelayInfo` accepts that ~2× over-allocation in exchange for one fewer pass), or drop `seen` and key off `uniq`. Two passes of dedup per node, in a hot path you're explicitly trying to deflate, is the wrong direction.

**4. Type and helper duplication with `GetRepeaterRelayInfo`.**

`type entry struct { ts string; pt int }`, the `collect` closure, the cutoff math (`cutoff1h`, `cutoff24h`, `windowCutoff`), and the latest-tracking + `RelayCount1h/24h` accumulation are copy-pasted verbatim from `repeater_liveness.go`. After fix #1 you'll have to keep two copies of the snapshot-read shape in lockstep. Please extract a small helper, e.g.

```go
func computeRelayInfoFromEntries(entries []entry, now time.Time, windowHours float64) RepeaterRelayInfo
```

and call it from both `GetRepeaterRelayInfo` and the batch method. Same applies to the prefix-lookup logic (`if len(key) >= 2 { prefix := key[:2]; if prefix != key { ... } }`) — that's now in two places; one helper.

**5. Doc comment claim doesn't match implementation.**

The comment says "Snapshots `byPathHop` slice headers for all requested pubkeys under that single lock — Processes timestamps and counts **outside** the lock." That description IS the bug from item 1. After the fix, please rewrite to: "Copies the per-node `(timestamp, payload-type)` tuples under the single read lock; lock is released before parsing/scoring." (And the doc-comment paragraph in the PR body should match.)

### Out-of-scope (not in this diff, file separately if you agree)

- `GetRepeaterUsefulnessScore` recomputes `totalNonAdvert` on every call — even single-node detail pages eat O(byPayloadType) per call. The TODO already mentions memoization off store invalidation; worth its own issue.
- `byPathHop` keying by 1-byte raw prefix is acknowledged-lossy in `repeater_liveness.go`'s comment and the same caveat now applies to the batch method's prefix lookup. Disambiguation via the path-listing endpoint's resolved-path SQL post-filter is referenced but not done here. Track separately.

### Verdict

Right diagnosis, right shape, but the lock-release ordering as written is unsafe and there are no behavior tests gating it. Once items 1–3 land (and ideally 4–5), happy to re-review.

Kpa-clawbot (commented):
## Kent Beck Gate (final) — NEEDS-WORK

### TDD compliance: FAIL

Three commits, four source files changed (`neighbor_persist.go`, `repeater_usefulness.go`, `routes.go`, `store.go`), **zero test files touched**. No red commit. No regression test. No benchmark.

### Exemption analysis

The PR body frames this as a perf fix ("O(N×M) → O(M+N)"), not a pure refactor. Even granting a refactor framing, this PR introduces **new observable behavior** that breaks the "same behavior, just faster" exemption:

1. **`GetRepeaterNodeStatsBatchCached` — 300s TTL cache.** Two `/api/nodes` calls within 5 minutes now return byte-identical relay stats even if the underlying store changed. That is user-visible behavior — stale `relay_count_1h`, `last_relayed`, `usefulness_score`. Not "same behavior."
2. **`neighbor_persist.go` — 200-row pagination + 5ms sleep between pages.** Changes commit semantics: one transaction per page instead of one transaction per batch. Failure mode is different (partial persistence is now possible mid-batch). Not "same behavior."
3. **New public method `GetRepeaterNodeStatsBatch`** has no test demonstrating it returns equivalent results to the per-node `GetRepeaterRelayInfo` + `GetRepeaterUsefulnessScore` it replaces.

### The Six Questions

1. **"Could a regression to O(N×M) pass the current tests?"** → **Yes.** There is no assertion on call count, lock acquisitions, or query shape. Reverting `handleNodes` to the per-node loop would leave CI green.
2. **"Smallest test that would have caught the original bug?"** → A unit test on `PacketStore` with N≥10 repeater pubkeys + M≥100 transmissions that asserts either (a) one `s.mu.RLock`/RUnlock pair via instrumented counter, or (b) a `testing.B` benchmark with a regression threshold. Neither exists.
3. **"Could a wrong implementation pass these tests?"** → N/A, no tests added. Worse: the existing `repeater_usefulness_test.go` doesn't exercise the new batch path at all.

### Must-fix (3)

1. **Correctness equivalence test** — `TestGetRepeaterNodeStatsBatch_MatchesPerNode`: seed the store with a realistic mix of payload types and path hops, call the batch method on a slice of pubkeys, then call `GetRepeaterRelayInfo` + `GetRepeaterUsefulnessScore` per-pubkey, assert field-by-field equality (`RelayActive`, `RelayCount1h`, `RelayCount24h`, `LastRelayed`, `Score`). This is the smallest test that catches a future refactor diverging the two code paths.
2. **Cache behavior test** — `TestGetRepeaterNodeStatsBatchCached_TTL`: call twice in quick succession, assert second call returns the cached map (e.g. mutate the store between calls and assert no change); then advance time / clear the cache and assert recompute. Without this, the 30s→300s TTL change in commit `00035736` is untested behavior on a user-visible path.
3. **Pagination semantics note OR test** — `asyncPersistResolvedPathsAndEdges` now commits per-page; a `break` on prepare/begin error leaves prior pages committed. Either add a test asserting prior pages persist on mid-batch failure, or document the intentional weakening of atomicity in a code comment so future readers don't "fix" it back.

### Nice-to-have

- A `BenchmarkHandleNodes_1500Repeaters` with a documented threshold (e.g. `b.ReportAllocs`; fail if >1s on N=1500). The PR body cites "18–44 seconds in production" as the trigger — that exact regression class deserves a guard.

### Verdict

**NEEDS-WORK.** Production-impact perf fix shipping with zero tests on the new code path, a new caching layer with no TTL test, and changed transaction semantics with no atomicity test. Even under the most generous refactor exemption, items 1 and 2 above are mandatory before merge.

— *Kent Beck gate, per AGENTS.md TDD policy*

Kpa-clawbot (commented):
## Carmack Review (final) — PR #1164

Focus: is the O(N+M) claim real, are there allocation/correctness bugs in the batching layer, and is the win measured?

### Verdict: **NEEDS-WORK**

The algorithm is right in spirit — the N×M denominator walk in `GetRepeaterUsefulnessScore` was the real cost, and hoisting `totalNonAdvert` out of the per-node loop kills it. But the PR ships an unmeasured perf claim and a cache-key correctness bug, both of which I have to flag.

---

### MUST-FIX (2)

**1. No benchmark. The entire PR premise is unmeasured.**

The description and code comments cite specific numbers — "20-30s on large datasets", "O(N × byPayloadType) → O(byPayloadType + N)", "cold compute ~25s" — but the diff contains zero benchmark code. No `BenchmarkGetRepeaterNodeStats*` in any `_test.go`. No before/after pprof output in the PR body. No timing harness.

This is the part that matters most on a perf PR. The optimization is plausible by inspection, but "I think this is faster" is exactly the kind of claim a benchmark exists to prove or kill. Add `BenchmarkGetRepeaterNodeStatsBatch` with `pubkeys = 1900`, `byPayloadType` populated to representative depth, and report ns/op + allocs/op before and after. Without it, the 300s TTL choice (which trades real-time accuracy for cost) has no grounding either — "8% duty cycle" is a calculation from a number nobody verified.

**2. Cache key ignores the pubkey set.**

`GetRepeaterNodeStatsBatchCached` keys on `windowHours` alone:

```go
if s.relayStatsCache != nil &&
    s.relayStatsCacheWindow == windowHours &&
    time.Since(s.relayStatsCacheAt) < 300*time.Second {
    return cached
}
```

The cached `map[string]RepeaterNodeStats` was computed for whatever pubkey set was passed at the cold-fill call. If a new repeater appears in `nodes` between fills (new advert, role flip from `client` → `repeater`), `handleNodes` requests stats for it but the cached map does not contain that key. The downstream `if stats, ok := relayStats[pk]; ok` then silently skips populating `relay_active`, `relay_count_1h`, `relay_count_24h`, and `usefulness_score` for that node for up to 5 minutes. A brand-new repeater renders on the map with no relay fields, which looks like a bug to operators.

Two acceptable fixes: (a) hash the sorted pubkey slice into the cache key so a set change forces a refill; or (b) detect missing keys after the cache hit and compute the missing entries inline (cheap — the per-node loop body is the small part).

---

### NITS (cleanup, won't block)

1. **`uniq` map is dead code.** Built in full but its only consumer is `len(uniq)` as a capacity hint for `entries` / `seen`. `seen` already does dedup. Drop `uniq` entirely; use `len(snap.txList)+len(snap.prefixList)` as the cap. That's two map allocations per node × ~1900 nodes per cache miss saved for nothing.

2. **Two-pass over entries is wasteful.** `collect()` builds `entries`, then the next loop iterates them again to apply cutoffs. Inline the `parseRelayTS` + cutoff logic inside `collect()` and the `entries` slice disappears. Saves one slice allocation per node and one full pass.

3. **Stale comment in `store.go`:** `// Cached relay stats batch result — recomputed at most once every 30s.` The code is 300s and the docstring on `GetRepeaterNodeStatsBatchCached` says 5 min. Three numbers, two correct, one stale — fix the struct comment to match.

4. **Cache miss thundering herd.** The fill releases `relayStatsCacheMu` before doing the expensive compute, then re-acquires it to store. If multiple `/api/nodes` requests hit a cold cache concurrently they all redo the full batch. With a 300s TTL this is rare, but trivial to fix with `singleflight.Group`.

5. **Behavioral shift not surfaced to clients.** Previously `relay_count_1h` was live; now it can be ~5 minutes stale. Worth adding `relay_stats_age_seconds` to the response (or at least a release note) so map/dashboard clients know the freshness contract.

6. **Unrelated change:** the `neighbor_persist.go` page-200 write loop is its own perf fix on a different hot path. Reasonable change but it should be a separate PR — easier to bisect a regression that way. Not a blocker if you're rushing, but flagging.

---

### Perf claim verified: PARTIAL (inspection only)

The shape of the fix is correct: hoisting `totalNonAdvert` out of the per-node loop turns the per-call cost from `O(N × len(byPayloadType))` into `O(len(byPayloadType) + sum over N of (|txList|+|prefixList|))`. For a node-rich, packet-rich store this is a real win. But "20-30s → fast" is asserted, not shown. Until a benchmark lands, I can verify the *direction* but not the *magnitude*.

Kpa-clawbot (changes requested):
## Independent review (final) — PR #1164

Verdict: **NEEDS-WORK**. The two follow-up commits (`0765dfce`, `00035736`) add a 300s TTL cache and SQLite write-paging, but **none of the 5 must-fix items from the prior review have been addressed**, and the cache layer introduces new correctness issues.

### Prior review items — status against current diff (HEAD `00035736`)

**1. Data race (snapshot slice headers under RLock, read elements after RUnlock) — NOT FIXED.**

`cmd/server/repeater_usefulness.go`:

- L92–105: still snapshots only headers under `s.mu.RLock()`:
  ```go
  snap := nodeSnap{txList: s.byPathHop[key]}
  // ...
  snaps[pk] = snap
  ```
- L107: `s.mu.RUnlock()` released before any element is read.
- L127, L132, L143, L165: `for _, tx := range snap.txList { ... tx.ID / tx.PayloadType / tx.FirstSeen ... }` all execute outside the lock.

`removeTxFromSlice` in `store.go` (the original concern) still shifts elements in place in the same backing array under the write lock, racing with these readers. The original review's recommended fix — extract `(ts, pt)` tuples (and the relayed count) inside the RLock, release, then process — has not been applied.

**2. No behavior tests for `GetRepeaterNodeStatsBatch` / `…Cached` — NOT FIXED.**

The diff contains no `_test.go` changes. There is no parity test against `GetRepeaterRelayInfo` / `GetRepeaterUsefulnessScore`, no empty-input test, no missing-pubkey test, no `-race` test. Per `AGENTS.md` "TDD IS MANDATORY", the new public methods on `PacketStore` require a red→green commit pair gating behavior. The PR's "Test plan" remains `go build` + "CI passes", which are not behavior tests. The cache layer added in `0765dfce` is itself a new behavior surface (TTL, key, mutation visibility) and also has zero tests.

**3. `uniq` map is dead work — NOT FIXED.**

`repeater_usefulness.go` L123–133 still builds `uniq` only to size `entries` / `seen`, then `seen` re-dedups in `collect`. Two passes of dedup per node in the hot path you are explicitly trying to deflate.

**4. Type/helper duplication with `GetRepeaterRelayInfo` — NOT FIXED.**

`type entry struct { ts string; pt int }`, the `collect` closure, the cutoff math, the latest-tracking + `RelayCount1h/24h` accumulation, and the 2-byte prefix lookup are still copy-pasted verbatim from `repeater_liveness.go`. No shared helper extracted.

**5. Doc-comment / impl mismatch — PARTIALLY MASKED, NOT FIXED.**

The new doc comment on `GetRepeaterNodeStatsBatch` (L72–75) no longer says "Processes timestamps and counts outside the lock", but the implementation still does exactly that. The doc no longer flags the bug, but the bug still exists — and the PR body in the description box still claims the original (unsafe) snapshot-headers-then-release behavior as the intended design.

### New must-fix introduced by follow-up commits

**6. `GetRepeaterNodeStatsBatchCached` ignores `pubkeys` on cache hit (`repeater_usefulness.go` L227–243).**

The cache key is `(windowHours, age < 300s)`. The `pubkeys` argument is not part of the key. Consequence:

- First call: `pubkeys = [A, B, C]` → cache stores `{A:…, B:…, C:…}`.
- Second call (within 300s) with `pubkeys = [A, B, C, D, E]` → cache hit, returns the **old** map with no entries for `D, E`.

In `routes.go` L1116 (`if stats, ok := relayStats[pk]; ok`), the `ok` is silently false for `D, E`, so newly-onboarded repeater/room nodes get **no** `relay_active`, `relay_count_1h`, `relay_count_24h`, or `usefulness_score` fields at all for up to 5 minutes after first appearing in `/api/nodes`. With 300s TTL and an actively growing node set this is a user-visible regression.

Fix: either include a hash of the sorted `pubkeys` slice in the cache key, or (simpler) compute the batch over *all* repeater pubkeys in the store and let callers filter — the denominator is already store-wide so this costs little.

**7. Cache has no invalidation on ingest (`repeater_usefulness.go` L227–253, `store.go` L192–196).**

The 300s TTL means `relay_active`, `relay_count_1h`, `last_relayed`, and `usefulness_score` on `/api/nodes` are stale by up to 5 minutes against the live store, even though the WS broadcast and `/api/packets` are real-time. This is a behavior regression for a real-time monitoring dashboard: a node that just relayed will continue to show `relay_active: false` for up to 5 min, and `relay_count_1h` lags by the same. There is no hook from `addTxToPathHopIndex` / `removeTxFromPathHopIndex` / `byPayloadType` mutations to invalidate `relayStatsCache`.

Either invalidate on mutation (cheap, just zero `relayStatsCacheAt`), or document this lag explicitly in the field-level comments on `node["relay_active"]` etc. The PR body should acknowledge this freshness tradeoff — currently it doesn't.

**8. Stale comment on cache TTL (`store.go` L192).**

```go
// Cached relay stats batch result — recomputed at most once every 30s.
```

TTL is 300s, not 30s. Commit `00035736` raised the TTL but didn't update the field comment. Make these match (and reconcile with the doc comment on `GetRepeaterNodeStatsBatchCached` which correctly says 300s / 5min).

### Out-of-scope (file separately if you agree)

- `neighbor_persist.go` SQLite obs-write paging (`00035736`): unrelated to the `/api/nodes` perf path. Reasonable change on its own merits, but it muddies this PR's scope and should land in its own PR with its own justification + test.
- `GetRepeaterUsefulnessScore` recomputing `totalNonAdvert` per call on single-node detail pages — still applies, still its own issue.
- 2-byte raw-prefix collisions in `byPathHop` lookup — same caveat as the original `repeater_liveness.go`, now duplicated here; track separately.

### Verdict

Caching on top of an unsafe implementation does not make it safe; it makes the unsafe window cover 5 minutes instead of one request. **Items 1–5 from the prior review must land before this is merge-ready**, and items 6–8 must be addressed before the cache layer ships at all. Happy to re-review once those are in and a real `-race` parity test gates them.

PR #1168, opened 2026-05-08
fix(#1056): row-detail slide-over panel at narrow widths (AC #4)

Kpa-clawbot:
Red commit: 8ac568bac37a808edab7fb66dfecab2a86123be8 (CI run: pending)

## Summary
Implements AC #4 of #1056: row-detail **slide-over panel** at narrow viewports for the Packets, Nodes, and Observers tables.

ACs #1–#3, #5 already shipped in #1099; this PR closes the remaining criterion.

## Approach
- Shared `window.SlideOver` helper (`packets.js`, top of file next to `TableResponsive`) — singleton overlay (`.slide-over-backdrop` + `.slide-over-panel`) injected into `<body>`. Close affordances: X button (`.slide-over-close`), backdrop click, Escape key. `aria-modal="true"`, focus moved to close button on open.
- Breakpoint: `window.innerWidth <= 1023` (matches the `data-priority="3"` threshold reused by `TableResponsive`). At `>=1024` the existing right-side panel / full-screen behavior is preserved — no regression.
- Each page (`packets.js`, `nodes.js`, `observers.js`) checks the breakpoint at row-click time and routes the same detail content into `SlideOver.open(node)` instead of the side panel / full-screen navigation.
- Reuses the existing `slideInRight` keyframe in `style.css`.
- CSS additions live in the table section of `style.css` only.

## E2E
`test-slideover-1056-e2e.js` — at 800x800 clicks the first row of each of the three tables, asserts `.slide-over-panel` + `.slide-over-backdrop` are visible and the close X exists; verifies Escape, backdrop click, and X click all dismiss; verifies that at 1440 the slide-over does NOT appear.

E2E assertion added: `test-slideover-1056-e2e.js:71`

## TDD
- Red commit: `8ac568b` — E2E asserts on `.slide-over-panel` which does not exist yet.
- Green commit: forthcoming in this PR.

Fixes #1056

PR #1177, opened 2026-05-08
fix(#1173): replace #liveDot with packet-driven brand-logo node-pulse

### LINKED ISSUE #1173: feat(logo): replace WS connected indicator with packet-driven logo node-pulse animation
## Behavior

Replace the current `#liveDot` WS-connected indicator (`public/app.js:479-481, 556`) with a packet-driven animation on the brand logo's two inner node circles.

### Animation
- The two inner circles already exist in the SVG (index.html:55-56):
  - **Node A**: `circle cx=540 cy=130 fill=var(--logo-accent)` — left
  - **Node B**: `circle cx=660 cy=130 fill=var(--logo-accent-hi)` — right
- On every received WS message, animate ONE quick "ping" between the two: Node A → Node B (or B → A — see below). Affected elements: ONLY the two inner circles. The arcs and wordmark do not animate. (Keeps it from looking busy.)
- A "ping" = a brief brightness/scale pulse on the source circle (~80ms), then the destination circle (~80ms). Total ≤200ms.
- **Direction alternates** per message: A→B, then B→A, then A→B... (no per-message rx/tx mapping needed; it's an aesthetic "data flowing" cue).
- Coalesce: if a ping is already mid-flight, the next trigger waits the minimum interval; don't queue.

### Throttle
Per Carmack: WS bursts must not queue 60 animations/sec. Cap at **15 animations/sec** (66ms minimum gap). Trigger model:
- `lastPingTs` tracked. If `now - lastPingTs < 66ms`, drop the trigger.
- Counter incremented for any dropped trigger; not exposed in UI.

### Connection-state encoding
With `#liveDot` removed, logo state IS the WS state:
- **Connected + flowing**: logo pulses per message (rate-limited).
- **Connected + idle ≥10s**: logo at full brightness, no animation.
- **Disconnected**: logo desaturated (`filter: grayscale(0.6) opacity(0.7)`) until reconnect.

### Theming / customizer
The animation MUST respect theme + customizer:
- All color comes from `--logo-accent` and `--logo-accent-hi` (the existing tokens that the customizer writes).
- Animation tweaks `filter: brightness()` and `transform: scale()` on the circle elements only. No hardcoded colors. No `rgb()`/`hsl()` literals.
- Theme switch (light ↔ dark) and customizer color changes apply with no JS reload.
- The 10s-idle and disconnect filters use only `var(--*)` tokens or filter functions that compose on top of the token-driven fill.

## Implementation notes
- Pure CSS animations. Class toggles only (`.logo-pulse-a`, `.logo-pulse-b`, `.logo-disconnected`) — no per-frame SVG mutation.
- Tagged classes on the two circles in BOTH SVG instances (`.brand-logo` and `.brand-mark-only`) so the mobile mark animates too.
- WS handler in `public/app.js` toggles the class with rate-limit gate; uses `requestAnimationFrame` once per ping to schedule the destination half.
- Honor `prefers-reduced-motion: reduce` — fall back to a single-step opacity blip on the destination circle only. No scale, no chained pulse.

## Acceptance criteria
- [ ] `#liveDot` element + its CSS + its WS-event toggles removed
- [ ] Two-circle ping fires on each WS message, alternating direction
- [ ] Rate-limited to 15/sec; verified via a Playwright test that bursts 100 synthetic WS frames and asserts ≤ ~15 animation classes toggled per second
- [ ] Disconnected state visible (desaturated logo)
- [ ] `prefers-reduced-motion` respected
- [ ] No layout shift at any viewport
- [ ] Both `.brand-logo` and `.brand-mark-only` (mobile) animate
- [ ] Theme switch + customizer color change reflected without reload (E2E)

## Out of scope
- Different animations per message type (ADVERT/TXT/etc.)
- Server-side message volume metrics
- Sound
- Directional rx/tx mapping (the alternation is aesthetic, not semantic)

## Open questions (low-stakes — pick during impl unless flagged)
1. Animation curve: `ease-out` (soft "blip") vs `linear` (hard "tick"). Default to `ease-out` unless we want a more telemetry-feel look.
2. 15 or 30 fps cap? 15 is half a busy radio's max; 30 is closer to "smooth" but doubles browser cost. Default to 15 — bump to 30 only if it feels laggy in field.

Kpa-clawbot:
Red commit: PENDING (will update)

Fixes #1173.

Replaces the `#liveDot` WebSocket-connected indicator with a packet-driven node-pulse animation on the brand logo's two inner circles.

## Behavior (locked per issue spec)
- **Animation curve:** `ease-out` (default per open-question 1).
- **Rate cap:** 15/sec (66ms gap; default per open-question 2). Excess triggers are dropped, never queued.
- **Direction:** alternates A→B / B→A across messages (aesthetic, not semantic).
- **Idle ≥10s:** logo at full brightness, no animation.
- **Disconnected:** `.logo-disconnected` applies `filter: grayscale(0.6) opacity(0.7)`.
- **`prefers-reduced-motion: reduce`:** single-step `.logo-pulse-blip` on destination only.

## Implementation
- WS handler hook lives in `public/app.js` `connectWS()` (`ws.onmessage` triggers `Logo.pulse()`; `ws.onopen`/`ws.onclose` toggle `Logo.setConnected()`).
- `Logo` is a small IIFE in `app.js` that exposes `window.__corescopeLogo` for E2E injection.
- All animation is pure CSS; JS only toggles `.logo-pulse-active` / `.logo-pulse-blip` / `.logo-disconnected`. Colors come exclusively from `--logo-accent` / `--logo-accent-hi` tokens.
- Two new classes (`.logo-node-a`, `.logo-node-b`) attached to inner circles in both `.brand-logo` and `.brand-mark-only` SVGs so the mobile mark animates too.

## `#liveDot` removal proof
```
$ grep -rn liveDot public/
(no output)
```

## E2E
- E2E assertion added: `test-logo-pulse-1173-e2e.js:54` and follows.
- Wired into the Playwright matrix in `.github/workflows/deploy.yml` (mirrors PR #1168 pattern from commit `5442652`).
- Test injects synthetic pings via `window.__corescopeLogo.pulse({ synthetic: true })`; matches the existing harness style (no new WS-mock pattern invented).

Red→green discipline preserved: the test commit lands first and CI fails on assertion; the implementation commit follows.


Kpa-clawbot (commented):
**Independent review** — APPROVED, zero must-fix

Reviewed via `gh pr diff` only (no checkout). Verified against `origin/fix/issue-1173-logo-pulse`.

### Correctness — Logo IIFE
- Rate gate: `lastPingTs` is updated only after the gap check passes; drops increment `stats.dropped`. Burst-resistant.
- Alternation: `aToB = (flip === 0)` captured before `flip ^= 1`. Strict A,B,A,B sequence.
- Generation counter: bumped in every `setConnected()` call; checked in both the source-half timer AND inside `requestAnimationFrame` before destination toggle. Zombie pulse cannot fight `.logo-disconnected`.
- `connected` re-check after rAF closes the second race window. Solid.
- `try { window.__corescopeLogo = api }` guards SSR/jsdom; `Object.freeze` keeps the seam read-only — correct call.

### `#liveDot` removal — clean
- `grep -rn 'liveDot\|live-dot' public/ test*.js` against the head ref:
  - Production: zero `#liveDot` references. Remaining `.live-dot` hits are in `live.js` legend bullets and `.vcr-live-dot` — pre-existing, unrelated, intended (CSS comment in `style.css` calls this out explicitly so a global no-op rule isn't shipped).
  - `test-pull-to-reconnect*.js` still call `makeEl('liveDot')` in their vm sandbox stubs. Now dead (app.js no longer touches `getElementById('liveDot')`) but harmless — not coverage-weakening.

### CSS
- `logo-pulse-step` keyframes mutate only `filter` + `transform`. `logo-pulse-blip-step` mutates only `opacity`. `.logo-disconnected` is `filter: grayscale(0.6) opacity(0.7)`. No hardcoded colors anywhere; circle fills resolve via `--logo-accent` / `--logo-accent-hi` tokens (test (h) asserts both fills are non-default).
- `transform-box: fill-box; transform-origin: center;` is the right idiom for scaling SVG `<circle>` about its own center.
- Both light + dark themes inherit via the existing `--logo-accent*` tokens (no theme-specific overrides in this PR — correct, not in scope).

### Both SVGs patched
`.brand-logo` and `.brand-mark-only` each carry `class="logo-node-a"` (cx=540) and `class="logo-node-b"` (cx=660) on the inner circles. E2E step (b) asserts both.

### Reduced-motion gated in BOTH layers
- JS: `reducedMotion()` short-circuits to `pulseBlip()` (destination opacity blip only).
- CSS: `@media (prefers-reduced-motion: reduce)` neutralizes `.logo-pulse-active` `animation: none` defensively.
- E2E (f) runs in a `reducedMotion: 'reduce'` context and asserts blip-only path.

### Orphan-test patches
- `test-e2e-playwright.js` (Nodes WS test): replaces `#liveDot` lookup with `.brand-logo` + `window.__corescopeLogo.setConnected` seam assertions. Coverage *strengthened* (now asserts the public seam exists).
- `test-logo-rebrand-e2e.js` step 3: replaces dot-right-of-logo geometry with brand-logo visibility + min-size (≥60×16) + seam check. Equivalent regression coverage for SVG layout, plus the seam contract — accepted.
- No coverage weakened.

### PR body
- "Red commit: PENDING (will update)" header is stale (the actual red commit `1fa2794` is on the branch ahead of green `ba644a3`), but accurate red→green ordering is in the commit log. Cosmetic, not blocking.
- All listed behaviors match the diff. No quietly deferred items detected.

### Verdict
APPROVED. Ship it.

Kpa-clawbot (commented):
## Carmack Review

**REQUEST CHANGES — 2 must-fix**

The animation is well-scoped: pure CSS keyframes, class-toggle-only, generation-cancellation on disconnect, rate-gate computed off a single time source. The chained-pulse half/rAF/setTimeout structure is correct, and the reduced-motion path really is cheaper (no scale + no rAF chain). Style invalidation is localized to `circle.logo-node-{a,b}` — the SVG paths don't get dragged through recalc.

Two real performance issues, both in the code that runs while the user has the tab in another window — i.e., the case nobody tests but the app spends most of its life in.

---

### M1 — Hidden-tab pulse storm (`public/app.js` `pulse` / `pulseChained`)

WS messages keep arriving when the tab is hidden. `pulse()` keeps running synchronously off `ws.onmessage`, the rate-gate happily passes 15/sec, and each pass schedules a `setTimeout(80)` → `rAF` → `setTimeout(80)` chain.

On a hidden tab Chrome:
- clamps `setTimeout` to ~1 Hz, and
- **does not invoke `rAF` callbacks at all** until the tab becomes visible.

So the chain's middle stage never fires while hidden. Each call captures a closure with `gen`, `srcSel`, `dstSel`, and survives in the pending-rAF queue. Steady state at 15/sec × hidden seconds:

- 60 s hidden → ~900 pending callbacks
- 1 h hidden → ~54 000 pending callbacks

Two costs: GC-rooted closures + object literals while hidden, and a stutter when the tab returns and they all unwind at once (the gen check will discard most of them, but only after they run).

The rate-gate isn't the right gate for this — the rate-gate's purpose is "render <16 frames/sec," and a hidden tab is rendering zero. Gate on `document.hidden` first.

```js
function pulse(_msg) {
  if (!connected) { stats.dropped++; return false; }
  if (typeof document !== 'undefined' && document.hidden) { stats.dropped++; return false; }
  const now = ...
```

Place it before `lastPingTs` is updated so the first ping after returning from a hidden tab fires immediately rather than being gated by a stale 66 ms-ago timestamp captured an hour earlier.

### M2 — `matchMedia` allocated per pulse (`public/app.js` `reducedMotion`)

```js
function reducedMotion() {
  try { return window.matchMedia && window.matchMedia('(prefers-reduced-motion: reduce)').matches; }
  catch (_) { return false; }
}
```

Called inside `pulse()` after the rate-gate, so up to 15/sec for the lifetime of the tab. `window.matchMedia(query)` returns a fresh `MediaQueryList` per call — a real allocation, not a cached lookup. At 15/sec for an hour that's ~54 000 throwaway MQLs. Reduced-motion preference is a user setting; it does not change between WS frames.

Cache once at module load:

```js
const reducedMotionMql = (typeof window !== 'undefined' && window.matchMedia)
  ? window.matchMedia('(prefers-reduced-motion: reduce)')
  : null;
function reducedMotion() { return !!(reducedMotionMql && reducedMotionMql.matches); }
```

Zero allocations on the post-gate path, same observable behavior (MQL `.matches` updates live as the OS preference changes — that was already the case here).

---

### Out of scope (noting, not blocking)

- `Array.prototype.slice.call(qsa(...))` in `$all` — modern `NodeList` has `.forEach` directly; the slice is a per-call array allocation. 4 calls per ping × 15/sec is ~60/sec — measurable on a profiler but not the cliff.
- `will-change: filter, transform` permanently on the four circles promotes a GPU layer per circle for the entire app lifetime. The circles are 13px radius so the layer cost is small; mention it in case a future change makes the SVG more complex.
- The `setTimeout` half-1 + `rAF` half-2 asymmetry is intentional (the rAF gives a pixel-aligned start to half-2) — fine.

Once M1 + M2 land, this is a clean, allocation-discipline-correct hot path.

PR #1191, opened 2026-05-12
fix(analytics): align region filter and time-window select in a single row

efiten:
## Summary

- Wraps the region filter and time-window `<select>` in a `.analytics-filters` flex container so they sit side-by-side at the same height
- Adds `.analytics-time-window-select` CSS class to match the `region-dropdown-trigger` button appearance (34px height, same border/font/radius)
- Overrides the global `select { min-height: 48px }` touch target rule that was making the select taller than the region button
- Removes the separate "Time window:" label — the select speaks for itself

## Test plan

- [ ] Open Analytics page — region filter and time-window select appear in a single row
- [ ] Both controls are the same height (~34px)
- [ ] Time-window select hover shows accent border, matching region button hover
- [ ] Selecting a time window still triggers a data reload
- [ ] On mobile, the controls wrap to a second line (flex-wrap)

### DIFF
--- a/public/analytics.js
+++ b/public/analytics.js
@@ -96,10 +96,9 @@
         <div class="analytics-header">
           <h2>📊 Mesh Analytics</h2>
           <p class="text-muted">Deep dive into your mesh network data</p>
-          <div id="analyticsRegionFilter" class="region-filter-container"></div>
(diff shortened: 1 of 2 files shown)


efiten:
Superseded by #1192 which applies the fix generically across all pages.

PR #1233, opened 2026-05-16
fix: ingestor uses ingest-time now() instead of observer receive time

marcelverdult:
## Problem
The ingestor stamps every stored packet with its own ingest-time `time.Now()`
(`BuildPacketData` in `db.go`; channel/DM paths in `main.go`), discarding the
observer receive time the uploader already puts in the MQTT envelope's
`timestamp` field. `MQTTPacketMessage` had no `Timestamp` field and
`handleMessage` parsed every envelope field except that one.

Observers that buffer packets offline and upload hours later get every
buffered packet displayed at upload time, not receive time — a 5-hour
deferred upload shows packets 5 hours late. Retained messages and broker
backlog hit the same skew.

## Why the envelope timestamp is trustworthy
Uploaders stamp `timestamp` when the radio receives the frame and freeze it;
the MQTT *message* is published late, but the `timestamp` *field* is not
re-stamped at publish. A buffered packet uploaded hours late still carries
its true receive time.

## Fix
New `resolveRxTime` helper reads `msg["timestamp"]` and falls back to
`time.Now()` only when it is missing, unparseable, or implausibly in the
future. Applied to all three ingest paths (raw packet, channel, DM). No
wire-format change — the field already exists.

Channel/DM dedup hashes intentionally stay on ingest time, since those bridge
messages carry no real packet hash and need ingest-unique input.

## Observer/node last_seen correction
Packet timestamps must reflect receive time, but observer/node `last_seen`
must not. `InsertTransmission` fed `data.Timestamp` (now rxTime) into
`observers.last_seen` and `UpsertNode`'s `last_seen`, so a buffered upload
could drag both fields backwards, and retained-message replay on MQTT
reconnect could flash long-offline observers as Online.

- `UpsertObserverAt` takes an explicit `lastSeen`; the status-packet and BLE
  companion handlers pass the resolved rxTime. `UpsertObserver` keeps its
  wall-clock behaviour for other callers.
- All three `last_seen` writes are guarded with
  `MAX(MIN(existing, ingestNow), rxTime)`: `last_seen` never moves backwards
  from a stale retained message, and never locks in a future value.

## Naive UTC+N timestamps
`resolveRxTime` rejects a timestamp only when it is >14h ahead (UTC+14 is the
maximum standard offset — anything further is a genuine clock error). A
timestamp that is merely in the future is soft-clamped to ingest time: a
future rxTime means a live packet from a UTC+N observer whose naive local
clock parses as-if UTC, not a buffered packet, so ingest time is correct and
no future timestamp reaches the DB.

For buffered packets from naive-clock uploaders a bounded residual offset
remains (equal to the observer's UTC offset); uploaders emitting zone-aware
ISO8601 everywhere would be the full cure but is a separate format change.

## Test
`cmd/ingestor/rxtime_test.go` covers `parseEnvelopeTime` (zone-aware, naive,
microseconds, garbage, empty) and `resolveRxTime` (plausible past used
verbatim, missing/garbage/future → ingest-time fallback). The existing
`TestBuildPacketData` is updated to supply an envelope timestamp and assert it
propagates, since `BuildPacketData` no longer self-stamps.


Kpa-clawbot (changes requested):
## Review — PR #1233 (substance only)

Two reviewers (adversarial + meshcore protocol). **Verdict: NEEDS-WORK — one real correctness bug; everything else is sound and the change is a net improvement.**

### The fix is correct
- Envelope `timestamp` propagates correctly through `handleMessage` → `BuildPacketData` → `InsertTransmission` (`db.go:670, 691, 723-725`).
- Other `time.Now()` callsites in the MQTT-reachable ingestor paths are either intentional (channel/DM dedup-hash input; `observer_metrics` time bucketing) or fallback-only after `resolveRxTime` parse failure.
- Backward-compat with old uploaders (no `timestamp` field): `resolveRxTime` falls back to `time.Now()` — same behavior as before.
- Future-clamp (`+14h`) + soft-clamp on `last_seen` (MAX with ingestNow) correctly handle retained-message replay.
- **Clock-skew (#1285) actually improves:** pre-PR, every buffered observer's observation contributed `(ingestTime - advertTS)` instead of `(rxTime - advertTS)`, inflating skew by hours for deferred uploaders. Post-PR, skew reflects actual radio-side offset.

### Real must-fix (single item)

**No past-bound on `resolveRxTime`** (`main.go:945-974`). Future timestamps `>14h` are rejected; arbitrarily old timestamps are accepted verbatim. Then in `InsertTransmission` (`db.go:685-687`):
```go
if rxTime < existingFirstSeen {
    stmtUpdateTxFirstSeen.Exec(rxTime, txID)
}
```
A single observation from a node with RTC reset to factory date (e.g. 2020-01-01) **permanently rewrites the canonical `transmissions.first_seen` to that bogus value** — pollutes clock-skew, retention queries, distance analytics, and downstream cursor pagination. Same failure class #1285 was just fixed for, just in a different layer.

**Fix**: add a lower bound to `resolveRxTime`. Either reject (fallback to `time.Now()`) or soft-clamp to ingestNow when rxTime is more than ~30 days in the past, OR before a hard floor year (e.g. 2024). Keeps the legitimate "buffered observer uploads from a few hours ago" use case; rejects the "RTC factory date 5 years ago" pathology.

Add a test for this: an envelope with `timestamp=2020-01-01T00:00:00Z`, fed into `resolveRxTime`, returns ingestNow (not the bogus 2020 value).

### Confirmation negative (honest)
We couldn't cite the firmware-side MQTT publisher convention from the local clone — only the C++ radio firmware is checked out, not the uploader-bridge repo. The "stamped at radio-receive, frozen" semantic asserted by the PR body matches CoreScope's downstream consumer expectations (`clock_skew.go` assumes `obs.Timestamp` is receive-time), but we can't cite an uploader source line. If you have a link to the uploader code, that closes the loop.

### Soft observations (informational, not blocking)
- Mixed-vintage `observations.timestamp` (old ingest-time rows + new rxTime rows) coexist until old rows roll off. Self-heals within `recentSkewWindowSec=3600` for live nodes.
- `transmissions.first_seen` semantic is now "earliest observer rxTime across all observations" instead of "ingest order." Downstream queries assuming monotonic id↔first_seen ordering (e.g. `oldestLoaded`) may see buffered uploads with first_seen < the slot's id-position. Real but small, not a blocker.

Once the past-bound is added, this is ready to merge — substance is approved on both protocol and design axes.

Kpa-clawbot (changes requested):
## Review — PR #1233 (substance only)

Two reviewers (adversarial + meshcore protocol). **Verdict: NEEDS-WORK — one real correctness bug; everything else is sound and the change is a net improvement.**

### The fix is correct
- Envelope `timestamp` propagates correctly through `handleMessage` → `BuildPacketData` → `InsertTransmission` (`db.go:670, 691, 723-725`).
- Other `time.Now()` callsites in the MQTT-reachable ingestor paths are either intentional (channel/DM dedup-hash input; `observer_metrics` time bucketing) or fallback-only after `resolveRxTime` parse failure.
- Backward-compat with old uploaders (no `timestamp` field): `resolveRxTime` falls back to `time.Now()` — same behavior as before.
- Future-clamp (`+14h`) + soft-clamp on `last_seen` (MAX with ingestNow) correctly handle retained-message replay.
- **Clock-skew (#1285) actually improves:** pre-PR, every buffered observer's observation contributed `(ingestTime - advertTS)` instead of `(rxTime - advertTS)`, inflating skew by hours for deferred uploaders. Post-PR, skew reflects actual radio-side offset.

### Real must-fix (single item)

**No past-bound on `resolveRxTime`** (`main.go:945-974`). Future timestamps `>14h` are rejected; arbitrarily old timestamps are accepted verbatim. Then in `InsertTransmission` (`db.go:685-687`):
```go
if rxTime < existingFirstSeen {
    stmtUpdateTxFirstSeen.Exec(rxTime, txID)
}
```
A single observation from a node with RTC reset to factory date (e.g. 2020-01-01) **permanently rewrites the canonical `transmissions.first_seen` to that bogus value** — pollutes clock-skew, retention queries, distance analytics, and downstream cursor pagination. Same failure class #1285 was just fixed for, just in a different layer.

**Fix**: add a lower bound to `resolveRxTime`. Either reject (fallback to `time.Now()`) or soft-clamp to ingestNow when rxTime is more than ~30 days in the past, OR before a hard floor year (e.g. 2024). Keeps the legitimate "buffered observer uploads from a few hours ago" use case; rejects the "RTC factory date 5 years ago" pathology.

Add a test for this: an envelope with `timestamp=2020-01-01T00:00:00Z`, fed into `resolveRxTime`, returns ingestNow (not the bogus 2020 value).

### Confirmation negative (honest)
We couldn't cite the firmware-side MQTT publisher convention from the local clone — only the C++ radio firmware is checked out, not the uploader-bridge repo. The "stamped at radio-receive, frozen" semantic asserted by the PR body matches CoreScope's downstream consumer expectations (`clock_skew.go` assumes `obs.Timestamp` is receive-time), but we can't cite an uploader source line. If you have a link to the uploader code, that closes the loop.

### Soft observations (informational, not blocking)
- Mixed-vintage `observations.timestamp` (old ingest-time rows + new rxTime rows) coexist until old rows roll off. Self-heals within `recentSkewWindowSec=3600` for live nodes.
- `transmissions.first_seen` semantic is now "earliest observer rxTime across all observations" instead of "ingest order." Downstream queries assuming monotonic id↔first_seen ordering (e.g. `oldestLoaded`) may see buffered uploads with first_seen < the slot's id-position. Real but small, not a blocker.

Once the past-bound is added, this is ready to merge — substance is approved on both protocol and design axes.

efiten:
Confirming the existing CHANGES_REQUESTED review is accurate. The missing past-bound in `resolveRxTime` is a real correctness bug — an RTC-reset node with factory date (e.g. 2020-01-01) will permanently overwrite `transmissions.first_seen` to that bogus value via the `stmtUpdateTxFirstSeen` path in `InsertTransmission`. The fix is straightforward: reject (fallback to ingestNow) when rxTime is more than ~30 days in the past, and add a test for that case. Once that's in, this is ready to merge — the rest of the design is sound.

nwithan8:
Is there any documentation on how to set the proper timezone for the various observer softwares/firmwares available? Doesn't seem to be any pattern to it, half our observers are reporting in UTC and the other half reporting local time, making it look like those reporting local time are several hours out of date (and marking the observers are offline as a result).

PR #1328, opened 2026-05-23
fix(live): add dark/light toggle to live-header-critical strip

efiten:
## Summary

- On mobile (≤640px) the top-nav is hidden entirely on `/live` — `#darkModeToggle` inside `.nav-right` was unreachable
- Adds a `☀️`/`🌙` button to the always-visible `live-header-critical` strip (beacon + pkt count row)
- Click delegates to `#darkModeToggle` — `app.js` stays the single owner of all theme logic, no duplication
- Matrix mode disables/enables the live button alongside the top-nav button so both stay in sync

## Test plan

- [ ] Mobile (≤640px) on `/live`: `☀️`/`🌙` button visible in the header strip; tap toggles theme
- [ ] Button icon matches current theme on page load and after each toggle
- [ ] Enable Matrix mode → both `#darkModeToggle` and the live button become disabled (greyed out)
- [ ] Disable Matrix mode → both buttons re-enabled, icons reflect restored theme
- [ ] Desktop: button not visible (hidden via CSS — top-nav toggle is used instead)
- [ ] `≤768px` non-phone breakpoint: button visible and correctly sized (36×36px)

### DIFF
--- a/public/live.css
+++ b/public/live.css
@@ -121,7 +121,8 @@
    operability rule). Visible glyph stays small (decorative); transparent
    padding expands the hit area without changing the visual chrome. */
 .live-header-toggle,
-.live-controls-toggle {
+.live-controls-toggle,
+.live-dark-toggle {
   display: none;
   align-items: center;
   justify-content: center;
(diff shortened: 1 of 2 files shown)


efiten:
Closing — wrong approach. The dark toggle belongs in the More sheet (bottom-nav, PR #1327), not in the live-header-critical strip.

PR #1331, opened 2026-05-23
fix(ingestor): bound neighbor-edges builder scan to retention window

efiten:
## Summary

- `buildAndPersistNeighborEdges()` ran an **unbounded** `JOIN observations+transmissions` every 60s with no `WHERE` clause
- On a large live deployment (9.7GB DB) this full-table scan takes several minutes, serialising SQLite writes and starving MQTT ingestion — the watchdog fires repeatedly with _"no messages received for 5m+"_ while staging (same broker, smaller DB) stays clean
- Regression introduced in #1289 (Option 4 neighbor-graph refactor)

## Fix

Add `lookbackSecs int64` parameter to `StartNeighborEdgesBuilder` and `buildAndPersistNeighborEdges`. The query now filters:
```sql
WHERE o.timestamp > strftime('%s', 'now') - ?
```
The lookback is derived from `NeighborEdgesDaysOrDefault()` (default 5 days), matching the existing edge retention window. The `idx_observations_timestamp` index (already present) makes this efficient regardless of DB size.

Historical edges already in `neighbor_edges` are preserved by the `ON CONFLICT DO UPDATE` upsert — only the incremental scan window shrinks.

## Test plan

- [ ] `TestNeighborEdgesBuilderUpsertsFromObservations` updated: observation seeded with `time.Now().Unix()` so it falls inside the lookback window; lookback arg `7*86400` passed explicitly
- [ ] Full ingestor test suite passes (`go test ./...`)
- [ ] Verify on live: `[neighbor-build]` log line now includes `lookback=5d`; MQTT watchdog stops firing

### DIFF
--- a/cmd/ingestor/main.go
+++ b/cmd/ingestor/main.go
@@ -200,9 +200,10 @@ func main() {
 	// Neighbor-edges builder (#1287 — Option 4): ingestor owns
 	// neighbor_edges writes. Runs every 60s. Server reads the snapshot
 	// via cmd/server/neighbor_recomputer.go on the same cadence.
-	stopNeighborBuilder := store.StartNeighborEdgesBuilder(NeighborEdgesBuilderInterval)
+	neighborLookback := int64(cfg.NeighborEdgesDaysOrDefault()) * 86400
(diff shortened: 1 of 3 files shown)


efiten:
Superseded by #1341 which was merged upstream with a better watermark/delta approach. The fix has been hot-deployed to live.

PR #1490, opened 2026-05-29
Make packet movement on the live map hardware-accelerated using HTML5

EldoonNemar:
Instead of forcing Leaflet to recalculate and paint heavy SVG DOM nodes 60 times a second for every moving packet, we will draw the flying dots and lines directly onto a hardware-accelerated HTML5 <canvas> overlaid on the map. Once the animation finishes, it will drop a static Leaflet line to handle the fading tail effect.


EldoonNemar:
Extended this branch to include fixes for what were described in Major and the first bullet point in minor

EldoonNemar:
Apologies, left some logic around while contemplating. Pushed a new change that addresses the blocker, major, and first minor. I'm still working on making a test

Kpa-clawbot (changes requested):
👋 First — thank you for this PR, @EldoonNemar. Migrating the per-packet flight from SVG to a single canvas is exactly the right direction; the live map gets brutal once a flood event lands. Real, useful contribution.

Multi-lens review below (perf + visual-integrity + adversarial). One blocker, a few majors, and minor polish. Happy to walk through any of these.

---

### 🔴 BLOCKER

**1. `drawAnimatedLine` no longer null-guards before pushing into `activeAnimations` (`public/live.js:3563–3603`).**
Old code did `if (!animLayer || !pathsLayer) { if (onComplete) onComplete(); return; }` at entry. The new version dropped that check. After `cleanup()` runs (route teardown, theme refresh) it nulls `pathsLayer` and `animLayer` — but if a queued WS callback fires `drawAnimatedLine` between teardown and `map` destruction, the anim still enters the array, and `renderAnimations` calls `map.latLngToContainerPoint(...)` which throws if `map` was disposed. Result: a console exception per stale callback and the rAF loop dies.

Suggested fix at top of `drawAnimatedLine`:
```js
if (!map || !animCtx) { if (onComplete) onComplete(); return; }
```
And in `cleanup()` (~line 3893) clear the queue:
```js
activeAnimations.length = 0;
isAnimating = false;
if (animCtx && animCanvas) animCtx.clearRect(0, 0, animCanvas.clientWidth, animCanvas.clientHeight);
```

---

### 🟠 MAJOR

**2. Canvas overlay desyncs from Leaflet during pan/zoom (`public/live.js:1182–1217`).**
The canvas is appended as a sibling of `.leaflet-map-pane` and uses `latLngToContainerPoint` per frame — but it does NOT participate in Leaflet's zoom/pan transform animation. During `zoomAnimation:true` transitions, node markers slide via CSS transform while in-flight packets snap to instantaneous container coords. Result: visible "the dot jumped off the line" for the duration of any zoom (≈250ms). This is a Tufte-style data-integrity regression — the flight visually decouples from its endpoints.

Either (a) listen for `zoomstart`/`zoomend` and pause the loop (acceptable, packets briefly freeze in place), or (b) add the canvas inside an overlay pane (e.g. via `map.createPane('animCanvas')` with `pane.style.zIndex = 625`) so Leaflet's transform applies to it. (b) is cleaner.

**3. Hardcoded `z-index: 625` (`public/live.js:1185`).**
Leaflet pane order: tilePane=200, overlayPane=400, markerPane=600, tooltipPane=650, popupPane=700. `625` sits between markerPane and tooltipPane — so packet flights cover node-marker labels but hide under tooltips. Probably intentional, but please either (a) hoist into a named pane with `map.createPane('animPane')` so the value lives in one place, or (b) at minimum add a code comment explaining the choice. Future Leaflet upgrades will rearrange pane z-indexes and silently break this.

**4. DPR change on monitor drag is missed (`public/live.js:1196–1213`).**
`ResizeObserver` fires on size changes, not DPR changes. Dragging the browser between a 1× and 2× monitor leaves the canvas backing store at the old DPR — animations get blurry until the next viewport resize. Optional fix:
```js
matchMedia(`(resolution: ${window.devicePixelRatio}dppx)`).addEventListener('change', resizeAnimCanvas, { once: true });
```
called after each resize to re-arm.

**5. Dead CSS hook (`public/live.js:3585`-ish).**
The old `L.polyline` had `className: 'live-packet-trace'` so themes could style flying packets. The new canvas path drops it (correct — canvas isn't styleable by CSS class). Please grep the repo for `.live-packet-trace` and remove orphaned CSS rules so reviewers don't trip on dead code later. I checked `public/` — no current consumers, so this is "delete with confidence" not "preserve."

**6. No new test asserts the canvas-engine refactor (repo AGENTS.md `TDD IS MANDATORY`).**
The existing `test-live-anims.js` only greps for `requestAnimationFrame` presence, which still passes with this diff (false-positive coverage). A minimal red-then-green could grep for the new symbols:
```js
assert.ok(src.includes('activeAnimations'), 'canvas engine missing');
assert.ok(src.includes("preferCanvas: true"), 'preferCanvas flag missing');
```
Not glamorous, but it's what the rest of the file does and it gates future regressions. If you want, I can suggest the exact patch in a follow-up.

---

### 🟡 MINOR

**7. Inconsistent indentation in the new block** (`public/live.js:1198–1212` uses 4-space inside the ResizeObserver branch; surrounding code is 2-space). Pure cosmetic — `prettier` or `eslint --fix` will sort it. Several trailing-whitespace lines too (`grep -nP ' $' public/live.js | head`).

**8. `if (window._wakeCanvasEngine)` self-check (line ~746)** is defensive but unnecessary — the function is defined unconditionally a few lines above. Safe to call directly.

**9. Magic `(dt / 660)` for animation duration** — matches the legacy `33ms × 20 steps`, but a `const ANIM_DURATION_MS = 660;` at the top of the block helps the next reader.

---

### ✅ Strengths worth calling out

- `setTransform(dpr, 0, 0, dpr, 0, 0)` to avoid scale compounding — exactly right; many canvas DPR bugs come from `ctx.scale(dpr,dpr)` accumulating.
- Pause/wake design with `lastTick = null` to avoid post-resume `dt` time-jumps — clean.
- Bounding `recentPaths` to 5 — kept the legacy cap, good.

---

### Merge state

`mergeStateStatus=BLOCKED` because:
- First-time contributor → workflow run required maintainer approval. Approved by me, now running.
- Required check `✅ Go Build & Test` must pass before merge.

I'll let CI finish on its own. Once green AND the BLOCKER above is addressed, this is good to land. Thanks again — and welcome to CoreScope.

EldoonNemar:
I'm working on it Mr. bot 🤣 

Completely rewriting the live map function is a bit of work.

EldoonNemar:
I cant figure out where your normal test file folder is, so I left it in the root like every other test- file. 

- I've cleaned up the test into a single test.describe. I've not worked with playwright before. 
- The only file I could find reference to live-packet-trace was in a root level test-*e2e*.js file. I've removed that test.
- I've added my playwright test seems to live.js
- I've updated the test-e2e-playwright.js test file so it works with the new scene instead of `requestAnimationFrame` . I also added two memory leak tests, a battery and gpu optimization check, and a DPR resize test.

EldoonNemar:
I'm hoping that finishes off the issues, unless I introduced more issues with the changes. I also cleaned up tabs and spaces

EldoonNemar:
Sorry Mr. Bot. I've been at it since 3am. I'm just tired ;)

EldoonNemar:
GRRRRRRRRRRRRRRRRRRRRRR. It did it again!

EldoonNemar:
antigravity autoformat on save was turned on, which completely changed the css when I tried to save it. My apologies.

EldoonNemar:
so angry it is

PR #1503, opened 2026-05-29
refactor: move version/commit badge from navbar to Perf dashboard

emuehlstein:
## Summary

The version/commit badge currently rendered in the nav stats bar (alongside packet counts, node counts, and observer counts) is operator-facing diagnostic information — not something end users need visible on every page load. For most visitors, it adds visual noise without adding value.

## Changes

- **perf.js**: Add a **Version** card to the Perf dashboard overview row. Shows `version` + short `commit` hash, both already available from `/api/health` (no new API surface needed). Card renders conditionally — if neither field is set it stays hidden.
- **app.js**: Remove `formatVersionBadge()` and `formatEngineBadge()` helper functions (now unused); strip the badge call from `updateNavStats()` so the navbar shows only packet/node/observer counts.
- **style.css**: Remove now-dead `.nav-stats .version-badge`, `.nav-stats .engine-badge`, and their link sub-rules.

## Rationale

The Perf page is explicitly the right place for this information — it's already scoped to operators and developers who want to know what version is running. The navbar is a high-visibility surface shared by all users; version strings belong in a diagnostic context, not a navigation bar.

Net result: navbar is cleaner for end users; operators can still find version info immediately on the Perf tab.


emuehlstein:
Follow-up per AGENTS.md requirements (commit 94d215c):

**Tests added** — `test-perf-go-runtime.js` +5 cases covering version card logic:
1. Version tag shown when `health.version` is set
2. Commit hash shown when version is `'unknown'`
3. Both version + commit shown when both are known
4. Card absent when neither field is present
5. Card absent when both values are `'unknown'`

All 18 tests pass (`node test-perf-go-runtime.js`).

**Browser validation** — confirmed on `dev-scope.chicagooffline.com/#/perf`: overview row shows `chicagooffline-dev · Server Version` card; navbar shows only pkts/nodes/obs with no version badge.

**Perf note** — this change touches only the Perf dashboard render path, which runs on a 5s interval and is not a hot path. The version card adds a single conditional string interpolation per refresh cycle — no additional API calls, no new data fetching.

emuehlstein:
Thanks for the thorough review — really appreciate the detailed findings breakdown. All six addressed in commit `25bd08d`:

**#1 (MAJOR)** — Rewrote the Playwright test. `.version-badge` / `.engine-badge` assertions are gone; new test asserts those classes are *absent* from `#navStats`, then navigates to `#/perf` and waits for a perf-card with label `Version`. CI should flip green.

**#2 (MAJOR)** — Dropped all three stale blocks from `test-frontend-helpers.js`: `formatEngineBadge` (L1810–1832), `formatVersionBadge` (L1980–2115), and the CSS `version-badge` contrast section (L2116–2131). Also removed both globals from `.eslintrc.json` and added `renderVersionCard`.

**#3 (MINOR)** — Replaced `style="font-size:0.85em"` with `.perf-num--small` modifier in `style.css`, with link + hover rules to match.

**#4 (MINOR)** — Links are back. Version tag → `.../releases/tag/vX.Y.Z`, commit → `.../commit/<sha>`, same pattern as the deleted `formatVersionBadge`. Intentional trade-off would have been the wrong call.

**#5 (MINOR)** — Hoisted the one-liner into `renderVersionCard(health)` at the top of `perf.js`. Added 5 direct unit tests for it in `test-perf-go-runtime.js` (null, unknown/unknown, release link, commit link, CSS class). Total: 23 tests, 23 pass.

**#6 (MINOR)** — Noted. Commit order is inverted from TDD ideal; the new `renderVersionCard` tests would have been red against the old inline code, so the ordering is closer now but I take the point for next time.

emuehlstein:
Follow-up fix (`761190e`): the e2e test was using `innerText` to read the version card content, which returns an empty string in headless Playwright when the element isn't fully painted. Switched to `textContent` (reliable regardless of visibility) and simplified the assertion — `waitForFunction` already confirmed the Version label exists, so we just read the `.perf-num` text directly and assert it's non-empty. CI re-running.

Issue #1146, opened 2026-05-07
bug(node-detail): "Paths Through This Node" links fail WCAG AA contrast in dark mode


Kpa-clawbot:
**Severity:** P1 (visibly wrong, user-flagged accessibility)

**Where:** Side panel + full node detail page → "Paths Through This Node"
**Theme:** Dark mode (light theme is also poor but less catastrophic)

## What's wrong

Path entries are rendered as `<div>` blocks (not `<table>`), and the only CSS rule that overrides link color inside `.node-detail-section` is:

```css
.node-detail-section .data-table td a,
.node-full-card .data-table td a {
  color: var(--accent);
}
```

(see `public/style.css:1097`)

Paths therefore fall back to UA default `rgb(0, 0, 238)`. Computed style sample (dark theme, side panel `#pathsContent a`):

```
link color:    rgb(0, 0, 238)   ← UA default blue
section bg:    rgb(26, 26, 46)
ratio:         ~3.0:1           ← fails WCAG AA (4.5:1 for body text)
```

## Evidence

- `node-analytics-review-screenshots/05-sidepanel-paths-dark-CONTRAST-FAIL.jpg`
- `node-analytics-review-screenshots/06-sidepanel-paths-dark-CONTRAST-FAIL.jpg`

Most hop names are barely readable.

## Recommended fix

Smallest fix:

```css
#pathsContent a,
#fullPathsContent a {
  color: var(--accent);
}
```

Better fix: introduce `--link-on-card` and `--link-on-card-hover` CSS variables, use them throughout `.node-detail-section` / `.node-full-card`, and validate against the customizer presets (issue #1086-style themes).

Verify in BOTH themes and against custom-theme color presets — this surface should pass WCAG AA in every shipped theme.

## Source

Section B3 of the workspace-local `node-analytics-review.md`.

PR #1517, opened 2026-05-31
fix(nodes): align --card-bg with --surface-2 in dark mode — low-contrast card fix (#1470)

### LINKED ISSUE #1470: bug(nodes): node details card backgrounds appear too bright in dark mode (low contrast vs page bg)
# bug(nodes): node details page card backgrounds look too bright in dark mode (low contrast vs page bg)

## Symptom
Operator report: "node details page doesn't honor dark mode" on mobile. Visual inspection on prod (`http://172.17.0.10:3000/#/nodes/<pubkey>`) and staging shows the `.node-full-card` and `.node-stats-table` elements rendering with a background that appears lighter / off-white compared to the surrounding page chrome.

## Reality check
CSS-wise the page IS in dark mode:
- `data-theme="dark"` set on `<html>`
- `--card-bg: #1a1a2e` (dark navy)
- `--surface-1: #1a1a2e`
- Computed `.node-full-card` background: `rgb(26, 26, 46)` — technically dark
- Text color `rgb(226, 232, 240)` — light, correct

But the visual contrast vs the page background (`rgb(15, 15, 35)` = `#0f0f23`) is only ~11 RGB units per channel. On many mobile screens (especially OLED with high black levels or LCDs with weak black-point), this 11-unit delta makes the cards look almost white-ish against the deep navy background.

## Root cause
The card background is too close in value to the page background. The intent was probably "subtle elevation" (a card slightly lighter than the page), but the chosen delta is too small to read as "elevated dark" — instead it reads as "out of place white card on dark page".

## Fix options
**A. Reduce delta** — make `--card-bg: #15152a` (delta=5) so cards blend with page; rely on `--border` for separation. Cleanest, least surprising.

**B. Increase delta** — make `--card-bg: #232340` and ensure borders contrast well. Cards stay distinct but feel more "dark mode" because they're decidedly darker than the chrome.

**C. Match `--card-bg` to `--surface-2` (`#232340`)** which is already used elsewhere — consistent across the app.

Recommend C. Test that other card-using surfaces (analytics, observers, packet detail) don't regress.

## Acceptance
- Card backgrounds visually distinct from page background but recognizably part of the same dark theme
- Text contrast remains WCAG AA (4.5:1)
- No light-mode regression (the same `--card-bg` var is used in both themes)

## Refs
- `public/style.css:2314` — `.node-full-card { background: var(--card-bg); ... }`
- `public/style.css:2768` — `.node-stats-table { background: var(--card-bg); ... }`
- Operator's mobile device specifics not yet captured (could be screen-profile contributing)

efiten:
## Problem

In dark mode, `.node-full-card` and `.node-stats-table` (and all other `var(--card-bg)` consumers) rendered with a background only ~11 RGB units away from the page background:

- Page bg: `--surface-0` = `#0f0f23` (RGB 15,15,35)  
- Card bg: `--surface-1` = `#1a1a2e` (RGB 26,26,46)  
- Delta: ~11 units per channel → appears near-white on OLED/high-contrast LCD screens

## Fix

Align `--card-bg` to `--surface-2` (`#232340`) in dark mode — the same value already used for `--detail-bg` throughout the app. Delta from page bg increases to ~35 units per channel, which reads clearly as an elevated dark surface rather than a washed-out off-white card.

Both dark-mode variable blocks updated in sync (`@media prefers-color-scheme: dark` + `[data-theme="dark"]`). Light mode is unchanged.

## Impact

All `var(--card-bg)` consumers in dark mode get the corrected colour: node full cards, stats tables, analytics cards, packet detail panels, dropdowns, etc. The value now matches `--detail-bg` so cards and detail panels use a consistent surface colour.

Closes #1470.

### DIFF
--- a/public/app.js
+++ b/public/app.js
@@ -1030,7 +1030,7 @@ window.addEventListener('DOMContentLoaded', () => {
         if (themeData[key]) root.setProperty(varMap[key], themeData[key]);
       }
       if (themeData.background) root.setProperty('--content-bg', themeData.contentBg || themeData.background);
-      if (themeData.surface1) root.setProperty('--card-bg', themeData.cardBg || themeData.surface1);
+      if (themeData.surface1) root.setProperty('--card-bg', themeData.cardBg || themeData.surface2 || themeData.surface1);
       // Nav gradient
       if (themeData.navBg) {
         var nav = document.querySelector('.top-nav');
--- a/public/customize-v2.js
+++ b/public/customize-v2.js
@@ -572,7 +572,7 @@
 
     // Derived vars
     if (themeSection.background) root.setProperty('--content-bg', themeSection.contentBg || themeSection.background);
-    if (themeSection.surface1) root.setProperty('--card-bg', themeSection.cardBg || themeSection.surface1);
+    if (themeSection.surface1) root.setProperty('--card-bg', themeSection.cardBg || themeSection.surface2 || themeSection.surface1);
 
     // Node colors → --node-X CSS var only (legacy compat).
     // #1412: do NOT push server-config nodeColors into window.ROLE_COLORS —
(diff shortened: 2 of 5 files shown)


efiten:
**Review findings — addressed in follow-up commit `aeaa483`**

---

**Finding 1 (medium) — JS theme-apply paths restore old card-bg for users with saved themes**

`customize.js:554`, `app.js:1033`, `customize-v2.js:575` all fall back to `surface1` for `--card-bg` when no explicit `cardBg` is present in a saved theme. After the CSS fix the default diverges: CSS says `surface-2`, JS says `surface-1`. Any user who has ever saved a custom theme would silently get the old low-contrast value back.

Fixed by updating the fallback chain to `cardBg || surface2 || surface1` in all three locations.

---

**Finding 2 (low) — Stale inline comment in `style.css`**

The `#1146` comment at line 1689 still read `--card-bg: #1a1a2e` (old surface-1 value) and cited a contrast ratio computed against that colour. After the CSS change the card surface is `#232340`. Comment updated with the corrected hex and ratio.

Issue #1514, opened 2026-05-31
Follow-up to #1490: canvas-anim cleanup (DPR listener, fade-pane, perf nits)


Kpa-clawbot:
## Follow-up to #1490 (canvas-based live map animations)

Tracking items surfaced during post-merge review of #1490 that were out of scope or deferred. Not blockers; quality + perf cleanup.

### Must-fix carryover

- [ ] **DPR-change listener brittleness** (`public/live.js:1257-1271`). `_dprChangeHandler` self-rebinds inside a `finally` after `removeEventListener`. A synchronous DPR re-entry between remove and re-add would lose the handler. Replace with a stable `window 'resize'` hook that recomputes DPR, or assert the new `MediaQueryList` is fully constructed before binding.
- [ ] **Stale "Uncomment if you created the custom pane" comments shipped** (`public/live.js:3682, 3687`). `createFadingLeafletLine()` polylines were meant to render on the new `animationsPane` (z=625) but the `pane:` lines are commented out, so fades render on `overlayPane` (z=400), under node markers. Decide: uncomment for consistent stacking, or delete the comments and document why fades live on overlayPane.

### Should-fix (quality/perf)

- [ ] **Auto-format churn** in `public/live.js`: whitespace-only flips (`function ()` ↔ `function()`, `slice(0,3)` ↔ `slice(0, 3)`, brace spacing) across ~20+ lines. Same pattern that commit `baeb49dc` was supposed to revert. Add a whitespace-revert pass.
- [ ] **Per-frame allocations in render loop** (`public/live.js:3551-3625`): two object literals (`fromPt`, `toPt`) allocated per anim per frame. At 50 concurrent anims × 60fps = ~6000 allocs/sec. Hoist to module-scoped scratch vars.
- [ ] **`onComplete` callbacks dropped on destroy** (`public/live.js:3539`). When `destroy()` clears `activeAnimations`, any pending `onComplete` (used by audio `onHop` hooks) never fires. Iterate `activeAnimations` and invoke `onComplete` before clearing.
- [ ] **`_liveTestSeams` defined twice** (`public/live.js:2778` and `:4079`). Second definition skips the pause/empty-queue guard, so tests can wake an empty engine. Delete the duplicate.
[...]

PR #1521, opened 2026-06-01
Additional live map performance optimizations

EldoonNemar:
This PR introduces a major performance optimization by migrating the final DOM-heavy animation (node pulses) into the hardware-accelerated canvas engine by migrating the concentric "pulse" rings (rendered when a node receives a packet) from DOM-based Leaflet L.circleMarker elements into the high-performance HTML5 canvas animation loop (activePulses). This completely eliminates DOM thrashing when dozens of nodes broadcast simultaneously, ensuring a buttery-smooth 60 FPS even under extreme packet volume.


EldoonNemar:
wake: wakeCanvasEngine was safely preserved in the consolidated _liveTestSeams object at the top of the exposure block. The duplicate we removed at the bottom of the file was an outdated implementation that bypassed the VCR pause/empty-queue guards (noted in #1514 S4). Any Playwright tests calling .wake() will now correctly hit the guarded production function.

EldoonNemar:
In response to "MAJOR — kent-beck axis: test commit present but assertion-shaped check needed"

Confirmed. The rewritten test specifically targets the new canvas engine path and utilizes the newly exposed test seams (`triggerPulse` and `getPulses`). 

It asserts the following behaviors:
1. It injects a synthetic pulse via `_liveTestSeams.triggerPulse()`.
2. It monitors the engine's internal array via `_liveTestSeams.getPulses()` using a `requestAnimationFrame` loop.
3. It asserts that the internal `pulse.hl_weight` never drops below `2` for the entire duration that the highlight ring is visible (`hl_op > 0`).

If the stroke weight math in `live.js` is reverted or altered to drop below 2, this test will accurately catch the a11y regression and fail.

PR #1524, opened 2026-06-01
fix(ui): prevent animation fast-forward on tab wake

EldoonNemar:
When the browser backgrounds the tab,drops frames due to DOM bloat, or user goes to another page; the uncapped delta time (`dt`) in the `requestAnimationFrame` loop caused the physics engine to simulate massive time jumps, making packets appear to fast-forward at 8x speed.

This commit:
- Clamps `dt` to a maximum of 32ms in both the path animation and node pulse loops to ensure graceful slowdowns during lag.
- Restricts the `VCR.speed` multiplier strictly to `REPLAY` mode so live packets are not accidentally accelerated.


EldoonNemar:
I'm on a roll

PR #1525, opened 2026-06-01
fix(ui): Area dropdown is no longer unreadable.

EldoonNemar:
## Summary
Resolves a visual bug where the text in the "Area" map filter dropdown was completely unreadable (white text rendering on a light background with borders). 

## Cause
The `.region-dropdown-item` CSS class is shared between the Region filter (which uses `<label>` elements) and the Area filter (which uses `<button>` elements). Browsers apply their own default OS-level backgrounds and borders to `<button>` elements. In dark mode, this resulted in white text (`color: var(--text)`) rendering on top of the browser's default light-grey button background.

## Changes
- **Button Reset:** Added `background: transparent; border: 0px;` to `.region-dropdown-item` to strip native browser button styling.
- **Layout Alignment:** Replaced `max-width: 320px;` with `width: 100%;` so the dropdown items consistently fill the full width of the `max-content` dropdown menu.

## Verification
- Verified that the Area dropdown now correctly inherits the intended background color from the parent `.region-dropdown-menu`.
- Confirmed that this styling change does not negatively impact the Region filter dropdown.
- Checked that there are no automated CSS assertions in the test suite that would break from this style change.


EldoonNemar:
well poo. I only wanted to create a pr for the fix(ui)

Issue #1505, opened 2026-05-29
Rename Ghost on the live map to something that's clearer to understand to the average person


EldoonNemar:
The ghost checkbox on the live map is confusing. There has to be a better name for the checkbox that's more descriptive. I have a few suggestions if you agree. I also suggest we change the color to some more prominent with dashes so it's easier to notice. Right now it matches ack/other but being gray you can't tell if it's a real ack or a ghost. I suggest bright yellow as the next closest thing is an orange direct message.

**Descriptive & Technical:**
Inferred Hops (My top recommendation. We are inferring their location based on the previous and next hops)
Interpolated Hops (Technically accurate, but maybe a bit jargon-heavy for casual users)
Estimated Hops (Very clear to average users that we are guessing the location)
Implied Hops
Synthetic Hops

**Describing the lack of data:**
Unmapped Hops
Locationless Hops
Blind Hops

**If you still want something slightly stylized/abstract:**
Transit Hops
Phantom Hops (Similar to ghost, but sounds slightly more technical)

Kpa-clawbot:
**Severity:** P3 — terminology + checkbox color contrast on Live map.

**Claim:** confirmed. "Ghost" label is the rendered string; underlying concept is hop interpolation.

**Code today:**
- `public/live.js:1035` — `<label>...Ghosts</label>` checkbox + `aria-describedby="ghostDesc"` ("Show interpolated ghost markers for unknown hops").
- `public/live.js:24,1281-1285` — state key `live-ghost-hops` (localStorage), toggle wired to `showGhostHops`.
- `public/live.js:2857-2884` — render: dashed line + pulsing `circleMarker` for unreached/inferred TRACE hops, color from `--trace-ghost-color` (`public/style.css:184` `#94a3b8` slate-gray).
- `public/live.css:722-726` — `.ghost-hop-marker` + `ghostPulse` keyframes (already a dashed/pulsing visual; reporter's "looks like ack" complaint is about the *checkbox* color, not the marker).

**Findings:**
- (tufte): "Ghost" is decorative jargon; the UI already self-describes the concept in the sr-only hint as "interpolated…unknown hops." Promote that to the label. **Inferred Hops** wins — it names the operation (interpolation between known endpoints) without metaphor. "Estimated" implies error bars we don't compute; "Phantom/Synthetic" trade one metaphor for another.
[...]

PR #1527, opened 2026-06-02
Fr(UI) - Rename ghost to inferred hops on live.js as partial fix for issue #1505

EldoonNemar:
Rename of ghost to inferred hops as described as partial fix for issue #1505
Update of ghostDesc in live.js, also mentioned as partial fix for issue #1505


EldoonNemar:
@Kpa-clawbot I've updated the body text and title to "as partial fix for issue"

Issue #1528, opened 2026-06-02
Bug: inappropriate use of `--accent` in titles/headers


EldoonNemar:
There are a few places on the site where the `--accent` color variable is used that isn't consistent with the rest of the page and poses the potential to make the text unreadable, falling outside of the WCAG standards we're trying to maintain.

- The table headers in panel-right on the packets page inherit the `--accent` value which falls outside of every other table/card heading that use `--text`.
- The copy link button in panel-right inherits the `--accent` value which can make the text unreadable if the user chooses light blue.
- The selected `vcr-scope-btns` on the live map page
- The `badge multibyte-badge` on the nodes page

<img width="405" height="684" alt="Image" src="https://github.com/user-attachments/assets/2d4b2606-94da-4d4d-80b6-3df3d252228c" />

<img width="288" height="80" alt="Image" src="https://github.com/user-attachments/assets/15b35ba6-5197-4bc5-9f25-f60fa45b7a0a" />

<img width="394" height="158" alt="Image" src="https://github.com/user-attachments/assets/70dcabca-595f-4f84-964e-c7065f3bb8e0" />

Kpa-clawbot:
**Severity:** P2 — WCAG contrast risk when operator picks a light accent; theming illusion turns text unreadable.

**Claim:** confirmed. Four surfaces inherit `var(--accent)` for foreground text where surrounding peers use `--text`:

**Code today:**
- `public/style.css:1244` — `.field-table .section-row td { … color: var(--accent); }` (panel-right detail section headers on Packets).
- `public/style.css:2624` — `.copy-link-btn { color: var(--accent, #3b82f6); … background: rgba(59,130,246,0.12); }` (foreground AND tinted bg both bound to blue rgba; the bg literal does not track `--accent` either — separate theming bug).
- `public/live.css:987` — `.vcr-scope-btn.active { color: var(--accent); background: rgba(59,130,246,0.2); border-color: rgba(59,130,246,0.3); }` (same dual-binding wart).
- `public/nodes.js:212` — inline `color:var(--accent, #14b8a6); background:var(--accent-bg, rgba(20,184,166,0.2))` on `.badge.multibyte-badge`.

**Findings:**
[...]

PR #1530, opened 2026-06-02
UI accent partial fix for issue #1528

EldoonNemar:
Made the suggested changes as listed in the fix path provided by @Kpa-clawbot 

Fix path:

`style.css:1244` `.field-table .section-row td` → `color: var(--text)` (or new `--section-header-fg`).
`style.css:2620-2631` `.copy-link-btn` → `color: var(--text);` background/border via `--accent-bg` / `--accent-border` tokens with safe defaults.
`live.css:987` `.vcr-scope-btn.active` → same token swap; ensure text remains `--text` on the tinted bg.
`nodes.js:212` `.multibyte-badge` → move inline styles to style.css, `color:var(--text)`, keep `--accent-bg` background.

When creating the defaults for `--accent-bg` and `--accent-border`, I chose to go with the default style values embedded in nodes.js, as that was the safest bet. 

We should probably extend the custom themes to include these variables as well as not to confuse users if they see it. This also causes the delima of, sometimes the `--accent` is use as the background for objects, and not `--accent-bg`, example:
`btn active` has background set to` --accent` and border set to `--accent`.

If we don't extend the config to accept accent-bg and accent-border, we risk users still making accents of light blue that will be drown out with the defaults we've set.

Also updated the badge above the multi-byte badge that contains X bytes of the nodes public key, where X is determined by the path byte length. This was done because it had styles set that were easy to add to the styles.css file, to clean up coe. The node-type badge above it is unfortunately driven by javascript in the nodes.js page, and requires syling.

**Note:** Accidentally added ghost changes into this push for a second time. They can be ignored as they were previously merged and shouldn't have been seen as new.


EldoonNemar:
Post fix look:

image
image
image

Issue #1045, opened 2026-05-05
Add an optional Filter to Live mode by IATA region


Bjorkan:
It would help clear up the live view a bit when there is a lot of traffic coming in. This would make it easier to focus on the regions you care about. It would also make the live mode feel less noisy during busy periods.

This would obviously be something a user can active themselves, and not something that would be server wide.

PR #1533, opened 2026-06-03
Fixes #1165: add OSM/Stamen tile providers with per-provider Leaflet layer control.

### LINKED ISSUE #1165: Add regular OSM map colors please?
As in title. The very black and very white maps are kinda not really useful.

EldoonNemar:
List of changes too long to describe, so I'll hit high level.

- Config now supports the json map tiles that were suggested by @Kpa-clawbot.
- Leaflet map layer button appears in the top right of live.js and map.js (because all the work was already done on live.js... Added bonus)
- Allows users to enter creds for OSM and Stamen to get enterprise related perks, in the config file
- Added a default light map under customizer. Still suggest removing them all together and relying on the config
- You can enable OSM and Stamen in the config without a license, but at your own risk!!!
- Config comment explains where to register and the providers for osm, as well as the general limits per X interval
- Updated tests (28) to address the changes made to the maps

### TDD Exemption

**Reason**: Net-new UI surfaces (per `AGENTS.md`)

This PR introduces a net-new UI surface (the multi-provider map tile selector). Under the `AGENTS.md` exemption for net-new UI surfaces, the absence of an initial failing (red) commit is permitted, as the UI was built first. However, the underlying public APIs are fully covered. 

The following tests serve as the first assertions for these new APIs:
- `window.MC_createLayerControl`: Asserted in `MC_createLayerControl handles Auto mode and explicit layers correctly`
- `window.MC_setDarkTileProvider` & `window.MC_getDarkTileProvider`: Asserted in `MC_setDarkTileProvider persists to localStorage...`
- `window.MC_setLightTileProvider` & `window.MC_getLightTileProvider`: Asserted in `MC_setLightTileProvider persists to localStorage...`
- `window.MC_initTileRegistry`: Asserted in `MC_initTileRegistry(true) dispatches mc-tile-provider-changed`
- `applyTileFilter`: Asserted in `applyTileFilter sets invert CSS for inverted dark provider...`
- Cross-tab synchronization: Asserted in `Cross-tab storage event re-dispatches mc-tile-provider-changed`

### DIFF
--- a/cmd/server/config.go
+++ b/cmd/server/config.go
@@ -64,6 +64,7 @@ type Config struct {
 	Roles            map[string]interface{} `json:"roles"`
 	HealthThresholds *HealthThresholds      `json:"healthThresholds"`
 	Map              map[string]interface{} `json:"map"`
+	Tiles            map[string]interface{} `json:"tiles"` // deprecated
 	SnrThresholds    map[string]interface{} `json:"snrThresholds"`
 	DistThresholds   map[string]interface{} `json:"distThresholds"`
 	MaxHopDist       *float64               `json:"maxHopDist"`
@@ -97,6 +98,10 @@ type Config struct {
 	CORSAllowedOrigins []string `json:"corsAllowedOrigins,omitempty"`
 
 	DebugAffinity bool `json:"debugAffinity,omitempty"`
+
+	// MapDarkTileProvider selects the default dark-mode basemap provider for
+	// new visitors. Deprecated: use Map.Tiles.DarkDefault instead.
+	MapDarkTileProvider string `json:"mapDarkTileProvider,omitempty"`
 
 	// ObserverBlacklist is a list of observer public keys to exclude from API
 	// responses (defense in depth — ingestor drops at ingest, server filters
@@ -352,12 +357,46 @@ func LoadConfig(baseDirs ...string) (*Config, error) {
 			continue
 		}
 		cfg.NormalizeTimestampConfig()
+		cfg.migrateDeprecatedConfig()
 		applyCORSEnv(cfg)
 		return cfg, nil
 	}
 	cfg.NormalizeTimestampConfig()
+	cfg.migrateDeprecatedConfig()
 	applyCORSEnv(cfg)
 	return cfg, nil // defaults
+}
+
+func (c *Config) migrateDeprecatedConfig() {
+	migrated := false
+	if c.Map == nil {
+		c.Map = make(map[string]interface{})
+	}
+	if c.Map["tiles"] == nil {
+		c.Map["tiles"] = make(map[string]interface{})
+	}
+	tilesMap, ok := c.Map["tiles"].(map[string]interface{})
+	if !ok {
+		return
+	}
+
+	if c.MapDarkTileProvider != "" {
+		if tilesMap["darkDefault"] == nil {
+			tilesMap["darkDefault"] = c.MapDarkTileProvider
+		}
+		migrated = true
+	}
+	if c.Tiles != nil {
+		for k, v := range c.Tiles {
+			if tilesMap[k] == nil {
+				tilesMap[k] = v
+			}
+		}
+		migrated = true
+	}
+	if migrated {
+		fmt.Fprintf(os.Stderr, "[deprecated] Top-level 'mapDarkTileProvider' and 'tiles' keys in config.json are deprecated and will be ignored in a future release. Please move them into 'map': { 'tiles': { ... } }.\n")
+	}
 }
 
 func LoadTheme(baseDirs ...string) *ThemeFile {
--- a/cmd/server/config.go
+++ b/cmd/server/config.go
@@ -386,7 +386,7 @@ func (c *Config) migrateDeprecatedConfig() {
 		}
 		migrated = true
 	}
-	if c.Tiles != nil {
+	if len(c.Tiles) > 0 {
 		for k, v := range c.Tiles {
 			if tilesMap[k] == nil {
 				tilesMap[k] = v
@@ -395,7 +395,7 @@ func (c *Config) migrateDeprecatedConfig() {
 		migrated = true
 	}
 	if migrated {
-		fmt.Fprintf(os.Stderr, "[deprecated] Top-level 'mapDarkTileProvider' and 'tiles' keys in config.json are deprecated and will be ignored in a future release. Please move them into 'map': { 'tiles': { ... } }.\n")
+		fmt.Fprintf(os.Stderr, "[deprecated] Top-level 'mapDarkTileProvider' and 'tiles' keys in config.json are deprecated and will be ignored in v3.5.0 (see #1165). Please move them into 'map': { 'tiles': { ... } }.\n")
 	}
 }
 
--- a/cmd/server/config.go
+++ b/cmd/server/config.go
@@ -63,7 +63,7 @@ type Config struct {
 
 	Roles            map[string]interface{} `json:"roles"`
 	HealthThresholds *HealthThresholds      `json:"healthThresholds"`
-	Tiles            map[string]interface{} `json:"tiles"`
+	Map              map[string]interface{} `json:"map"`
 	SnrThresholds    map[string]interface{} `json:"snrThresholds"`
 	DistThresholds   map[string]interface{} `json:"distThresholds"`
 	MaxHopDist       *float64               `json:"maxHopDist"`
@@ -97,13 +97,6 @@ type Config struct {
 	CORSAllowedOrigins []string `json:"corsAllowedOrigins,omitempty"`
 
 	DebugAffinity bool `json:"debugAffinity,omitempty"`
-
-	// MapDarkTileProvider selects the default dark-mode basemap provider for
-	// new visitors. The client may override per-browser via the customizer
-	// (persisted to localStorage). Allowed values: "carto-dark" (default),
-	// "esri-darkgray-labels", "voyager-inverted", "positron-inverted". See
-	// public/map-tile-providers.js for the registry. #1420.
-	MapDarkTileProvider string `json:"mapDarkTileProvider,omitempty"`
 
 	// ObserverBlacklist is a list of observer public keys to exclude from API
 	// responses (defense in depth — ingestor drops at ingest, server filters
--- a/cmd/server/routes.go
+++ b/cmd/server/routes.go
@@ -355,7 +355,7 @@ func (s *Server) handleConfigClient(w http.ResponseWriter, r *http.Request) {
 	writeJSON(w, ClientConfigResponse{
 		Roles:               s.cfg.Roles,
 		HealthThresholds:    s.cfg.GetHealthThresholds().ToClientMs(),
-		Tiles:               s.cfg.Tiles,
+		Map:                 s.cfg.Map,
 		SnrThresholds:       s.cfg.SnrThresholds,
 		DistThresholds:      s.cfg.DistThresholds,
 		MaxHopDist:          s.cfg.MaxHopDist,
@@ -367,7 +367,6 @@ func (s *Server) handleConfigClient(w http.ResponseWriter, r *http.Request) {
 		PropagationBufferMs: float64(s.cfg.PropagationBufferMs()),
 		Timestamps:          s.cfg.GetTimestampConfig(),
 		DebugAffinity:       s.cfg.DebugAffinity,
-		MapDarkTileProvider: s.cfg.MapDarkTileProvider,
 	})
 }
 
--- a/cmd/server/routes.go
+++ b/cmd/server/routes.go
@@ -367,6 +367,8 @@ func (s *Server) handleConfigClient(w http.ResponseWriter, r *http.Request) {
 		PropagationBufferMs: float64(s.cfg.PropagationBufferMs()),
 		Timestamps:          s.cfg.GetTimestampConfig(),
 		DebugAffinity:       s.cfg.DebugAffinity,
+		MapDarkTileProvider: s.cfg.MapDarkTileProvider,
+		Tiles:               s.cfg.Tiles,
 	})
 }
 
--- a/cmd/server/types.go
+++ b/cmd/server/types.go
@@ -992,7 +992,7 @@ type MapConfigResponse struct {
 type ClientConfigResponse struct {
 	Roles              interface{} `json:"roles"`
 	HealthThresholds   interface{} `json:"healthThresholds"`
-	Tiles              interface{} `json:"tiles"`
+	Map                interface{} `json:"map"`
 	SnrThresholds      interface{} `json:"snrThresholds"`
 	DistThresholds     interface{} `json:"distThresholds"`
 	MaxHopDist         interface{} `json:"maxHopDist"`
@@ -1004,9 +1004,6 @@ type ClientConfigResponse struct {
 	PropagationBufferMs float64         `json:"propagationBufferMs"`
 	Timestamps          TimestampConfig `json:"timestamps"`
 	DebugAffinity       bool            `json:"debugAffinity,omitempty"`
-	// #1420 — server default for dark-tile provider picker. Client uses this
-	// as the fallback when no localStorage override is set.
-	MapDarkTileProvider string `json:"mapDarkTileProvider,omitempty"`
 }
 
 // ─── IATA Coords ───────────────────────────────────────────────────────────────
--- a/cmd/server/types.go
+++ b/cmd/server/types.go
@@ -1005,7 +1005,7 @@ type ClientConfigResponse struct {
 	PropagationBufferMs float64         `json:"propagationBufferMs"`
 	Timestamps          TimestampConfig `json:"timestamps"`
 	DebugAffinity       bool            `json:"debugAffinity,omitempty"`
-	MapDarkTileProvider string          `json:"mapDarkTileProvider,omitempty"` // deprecated
+	MapDarkTileProvider string          `json:"mapDarkTileProvider,omitempty"` // deprecated. TODO: remove after v3.5.0
 }
 
 // ─── IATA Coords ───────────────────────────────────────────────────────────────
--- a/cmd/server/types.go
+++ b/cmd/server/types.go
@@ -993,6 +993,7 @@ type ClientConfigResponse struct {
 	Roles              interface{} `json:"roles"`
 	HealthThresholds   interface{} `json:"healthThresholds"`
 	Map                interface{} `json:"map"`
+	Tiles              interface{} `json:"tiles,omitempty"` // deprecated
 	SnrThresholds      interface{} `json:"snrThresholds"`
 	DistThresholds     interface{} `json:"distThresholds"`
 	MaxHopDist         interface{} `json:"maxHopDist"`
@@ -1004,6 +1005,7 @@ type ClientConfigResponse struct {
 	PropagationBufferMs float64         `json:"propagationBufferMs"`
 	Timestamps          TimestampConfig `json:"timestamps"`
 	DebugAffinity       bool            `json:"debugAffinity,omitempty"`
+	MapDarkTileProvider string          `json:"mapDarkTileProvider,omitempty"` // deprecated
 }
 
 // ─── IATA Coords ───────────────────────────────────────────────────────────────
--- a/config.example.json
+++ b/config.example.json
@@ -55,8 +55,28 @@
     "opacity": 1,
     "_comment": "#1488/#1506 — outline around each map marker (live + map pages). Defaults restored to v3.7.2 visual (solid white, 2px). 'color' accepts any CSS color (hex, rgb, rgba); use rgba() or drop 'opacity' below ~0.5 to soften the outline when hundreds of nodes feel overwhelming. 'width' is the SVG stroke width (0 hides the outline entirely). Operators can override per-browser via the in-app Theme Customizer (Colors tab → Marker Stroke)."
   },
-  "mapDarkTileProvider": "carto-dark",
-  "_comment_mapDarkTileProvider": "Default dark-mode basemap provider. Allowed: 'carto-dark' (Carto dark_all — default), 'esri-darkgray-labels' (Esri Dark Gray Canvas + reference labels), 'voyager-inverted' (Carto Voyager with CSS invert filter), 'positron-inverted' (Carto Positron with CSS invert filter). Light mode is unaffected. Users can override per-browser via the in-app customizer (persisted to localStorage). #1420.",
+  "map": {
+    "tiles": {
+      "darkDefault": "carto-dark",
+      "lightDefault": "carto-light",
+      "providers": {
+        "carto": {
+          "enabled": true,
+          "domain": ""
+        },
+        "osm": {
+          "enabled": false,
+          "provider": "",
+          "token": ""
+        },
+        "stamen": {
+          "enabled": false,
+          "token": ""
+        }
+      }
+    }
+  },
+  "_comment_map_tiles": "Configure available map tile providers and the default light/dark basemaps. Carto is the default free-tier provider (custom domain can be specified for Carto enterprise). OSM and Stamen are disabled by default due to strict rate limits; if enabling, you must provide a valid provider name and API token. For Stamen (hosted by Stadia - ~200k tiles/mo free), register at https://stadiamaps.com/. For OSM, supported providers are: 'mapbox' (~50k tiles/mo free, https://www.mapbox.com/), 'thunderforest' (~150k tiles/mo free, https://www.thunderforest.com/), and 'maptiler' (~100k tiles/mo free, https://www.maptiler.com/).",
   "home": {
     "heroTitle": "CoreScope",
     "heroSubtitle": "Find your nodes to start monitoring them.",
--- a/config.example.json
+++ b/config.example.json
@@ -60,15 +60,18 @@
       "darkDefault": "carto-dark",
       "lightDefault": "carto-light",
       "providers": {
+        "_comment_carto": "Carto is the default free-tier provider. Optional: specify 'domain' for Carto enterprise.",
         "carto": {
           "enabled": true,
           "domain": ""
         },
+        "_comment_osm": "OSM providers: 'mapbox', 'thunderforest', 'maptiler'. WARNING: Tokens are sent to the browser. Apply origin/referrer restrictions in your provider dashboard.",
         "osm": {
           "enabled": false,
           "provider": "",
           "token": ""
         },
+        "_comment_stamen": "Stamen (hosted by Stadia). WARNING: Tokens are sent to the browser. Apply origin/referrer restrictions.",
         "stamen": {
           "enabled": false,
           "token": ""
@@ -76,7 +79,9 @@
       }
     }
   },
-  "_comment_map_tiles": "Configure available map tile providers and the default light/dark basemaps. Carto is the default free-tier provider (custom domain can be specified for Carto enterprise). OSM and Stamen are disabled by default due to strict rate limits; if enabling, you must provide a valid provider name and API token. For Stamen (hosted by Stadia - ~200k tiles/mo free), register at https://stadiamaps.com/. For OSM, supported providers are: 'mapbox' (~50k tiles/mo free, https://www.mapbox.com/), 'thunderforest' (~150k tiles/mo free, https://www.thunderforest.com/), and 'maptiler' (~100k tiles/mo free, https://www.maptiler.com/).",
+  "_comment_deprecated_tiles": "The old top-level 'tiles' and 'mapDarkTileProvider' keys are deprecated. Please use the 'map' block above.",
+  "mapDarkTileProvider": "",
+  "tiles": {},
   "home": {
     "heroTitle": "CoreScope",
     "heroSubtitle": "Find your nodes to start monitoring them.",
--- a/config.example.json
+++ b/config.example.json
@@ -60,7 +60,7 @@
       "darkDefault": "carto-dark",
       "lightDefault": "carto-light",
       "providers": {
-        "_comment_carto": "Carto is the default free-tier provider. Optional: specify 'domain' for Carto enterprise.",
+        "_comment_carto": "Carto is the default free-tier provider. Optional: specify 'domain' for Carto enterprise (e.g. 'mycompany' for 'https://{s}.mycompany.cartocdn.com').",
         "carto": {
           "enabled": true,
           "domain": ""
@@ -79,9 +79,6 @@
       }
     }
   },
-  "_comment_deprecated_tiles": "The old top-level 'tiles' and 'mapDarkTileProvider' keys are deprecated. Please use the 'map' block above.",
-  "mapDarkTileProvider": "",
-  "tiles": {},
   "home": {
     "heroTitle": "CoreScope",
     "heroSubtitle": "Find your nodes to start monitoring them.",
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -1,496 +1,475 @@
-# CoreScope Deployment Guide
+# Deploying CoreScope
 
-Comprehensive guide to deploying and operating CoreScope. For a quick start, see [DEPLOY.md](../DEPLOY.md).
+Get CoreScope running with automatic HTTPS on your own server.
 
 ## Table of Contents
 
-- [System Requirements](#system-requirements)
-- [Docker Deployment](#docker-deployment)
-- [Configuration Reference](#configuration-reference)
-- [MQTT Setup](#mqtt-setup)
-- [TLS / HTTPS](#tls--https)
-- [Monitoring & Health Checks](#monitoring--health-checks)
-- [Backup & Restore](#backup--restore)
+- [What You'll End Up With](#what-youll-end-up-with)
+- [What You Need Before Starting](#what-you-need-before-starting)
+- [Installing Docker](#installing-docker)
+- [Quick Start](#quick-start)
+- [Connecting an Observer](#connecting-an-observer)
+- [HTTPS Options](#https-options)
+- [MQTT Security](#mqtt-security)
+- [Database Backups](#database-backups)
+- [Updating](#updating)
+- [Customization](#customization)
 - [Troubleshooting](#troubleshooting)
+- [Architecture Overview](#architecture-overview)
 
----
+## What You'll End Up With
 
-## System Requirements
+- CoreScope running at `https://your-domain.com`
+- Automatic HTTPS certificates (via Let's Encrypt + Caddy)
+- Built-in MQTT broker for receiving packets from observers
+- SQLite database for packet storage (auto-created)
+- Everything in a single Docker container
 
-| Resource | Minimum | Recommended |
-|----------|---------|-------------|
-| RAM | 256 MB | 512 MB+ |
-| Disk | 500 MB (image + DB) | 2 GB+ for long-term data |
-| CPU | 1 core | 2+ cores |
-| Architecture | `linux/amd64`, `linux/arm64` | — |
-| Docker | 20.10+ | Latest stable |
+## What You Need Before Starting
 
-CoreScope runs well on Raspberry Pi 4/5 (ARM64). The Go server uses ~300 MB RAM for 56K+ packets.
+### A server
+A computer that's always on and connected to the internet:
+- **Cloud VM** — DigitalOcean, Linode, Vultr, AWS, Azure, etc. A $5-6/month VPS works. Pick **Ubuntu 22.04 or 24.04**.
+- **Raspberry Pi** — Works, just slower to build.
+- **Home PC/laptop** — Works if your ISP doesn't block ports 80/443 (many residential ISPs do).
 
----
+You'll need **SSH access** to your server. Cloud providers give you instructions when you create the VM.
 
-## Docker Deployment
+### A domain name
+A domain (like `analyzer.example.com`) pointed at your server's IP:
+- Buy one (~$10/year) from Namecheap, Cloudflare, etc.
+- Or use a free subdomain from [DuckDNS](https://www.duckdns.org/) or [FreeDNS](https://freedns.afraid.org/)
 
-### Quick Start (one command)
+After getting a domain, create an **A record** pointing to your server's IP address. Your domain provider's dashboard will have a "DNS" section for this.
+
+**Important:** DNS must be configured and propagated *before* you start the container. Caddy will try to provision certificates on startup and fail if the domain doesn't resolve. Verify with `dig analyzer.example.com` — it should show your server's IP.
+
+### Open ports
+Your server's firewall must allow:
+- **Port 80** — needed for HTTPS certificate provisioning (Let's Encrypt ACME challenge)
+- **Port 443** — HTTPS traffic
+
+Cloud providers: find "Security Groups" or "Firewall" in the dashboard, add inbound rules for TCP 80 and 443 from 0.0.0.0/0.
+
+Ubuntu firewall:
+```bash
+sudo ufw allow 80/tcp
+sudo ufw allow 443/tcp
+```
+
+## Installing Docker
+
+Docker packages an app and all its dependencies into a container — an isolated environment with everything it needs to run. You don't install Node.js, Mosquitto, or Caddy separately; they're all included in the container.
+
+SSH into your server and run:
 
 ```bash
-docker run -d --name corescope \
+# Install Docker
+curl -fsSL https://get.docker.com | sh
+
+# Allow your user to run Docker without sudo
+sudo usermod -aG docker $USER
+```
+
+**Log out and SSH back in** (the group change needs a new session), then verify:
+
+```bash
+docker --version
+# Should print: Docker version 24.x.x or newer
+```
+
+## Quick Start
+
+The easiest way — use the management script:
+
+```bash
+git clone https://github.com/Kpa-clawbot/corescope.git
+cd corescope
+./manage.sh setup
+```
+
+It walks you through everything: checks Docker, creates config, asks for your domain, checks DNS, builds, and starts.
+
+After setup, manage with:
+```bash
+./manage.sh status       # Check if everything's running
+./manage.sh logs         # View logs
+./manage.sh backup       # Backup the database
+./manage.sh update       # Pull latest + rebuild + restart
+./manage.sh mqtt-test    # Check if MQTT data is flowing
+./manage.sh help         # All commands
+```
+
+### Manual setup
+
+```mermaid
+flowchart LR
+    A[Clone repo] --> B[Create config] --> C[Create Caddyfile] --> D[Build & run] --> E[Open site]
+    style E fill:#22c55e,color:#000
+```
+
+### 1. Download the code
+
+```bash
+git clone https://github.com/Kpa-clawbot/corescope.git
+cd corescope
+```
+
+### 2. Create your config
+
+```bash
+cp config.example.json config.json
+nano config.json
+```
+
+Change the `apiKey` to any random string. The rest of the defaults work out of the box.
+
+```jsonc
+{
+  "apiKey": "change-me-to-something-random",
+  ...
+}
+```
+
+Save: `Ctrl+O`, `Enter`, `Ctrl+X`.
+
+### 3. Set up your domain for HTTPS
+
+```bash
+mkdir -p caddy-config
+nano caddy-config/Caddyfile
+```
+
+Enter your domain (replace `analyzer.example.com` with yours):
+
+```
+analyzer.example.com {
+    reverse_proxy localhost:3000
+}
+```
+
+Save and close. Caddy handles certificates, renewals, and HTTP→HTTPS redirects automatically.
+
+### 4. Build and run
+
+```bash
+docker build -t corescope .
+
+docker run -d \
+  --name corescope \
+  --restart unless-stopped \
   -p 80:80 \
-  -v corescope-data:/app/data \
-  ghcr.io/kpa-clawbot/corescope:latest
+  -p 443:443 \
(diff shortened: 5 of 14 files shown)


EldoonNemar:
Wow. It passed all checks.. I'll drink to that! 🥂

Kpa-clawbot (changes requested):

> *"I'm not interested in whether your tests pass. I'm interested in whether they would fail if you introduced a bug."*

**Verdict: NEEDS-WORK** — 3 must-fix.

### TDD red→green history — EXEMPT (verified by reviewer)

Branch commits (non-merge):
- `552df03a` fix(ui): live filters layout — touches `public/live.css`, `public/live.js` only, no tests
- `7c313b68` fix(#1165): multi-provider tile selector — production code + tests in one commit

No red-then-green commit pair. Per AGENTS.md, the tile-provider work is a **net-new UI surface** (provider registry, layer control, map-mode switch) and qualifies for the exemption: test must land in same PR but doesn't need to be the FIRST commit. Both test files are present.

I verified the tests actually gate behavior by reverting `public/map-tile-providers.js` to master and re-running `test-issue-1420-tile-providers.js`: **23 / 28 fail**. That is the missing red signal, reconstructed after the fact. Acceptable for this PR.

Caveat: `552df03a` (live filters layout) is an existing-UI tweak that ships zero tests. Per AGENTS.md it should have a red commit, but it's a pure CSS/layout change with no behavior assertion possible at unit level — flagging, not blocking.

### Six Questions

**a. Show me the test that fails when this change is reverted.** ✅ Most do — registry/persistence/CSS-filter/dispatch tests all flip red on revert. Solid.

**b. Smallest test that would have caught a tile-URL-template injection?** ❌ Missing. See must-fix #1 below. Tokens are concatenated raw into the URL with no `encodeURIComponent` (`_getOsmUrl` thunderforest/maptiler/mapbox branches, `_getStamenUrl`). A token of `evil&z=0` silently mangles the request URL.

**c. Could a wrong implementation pass these tests?** ⚠️ Partially. Two assertions use `for (const id of Object.keys(reg))` over the registry — vacuously green on an empty registry. See must-fix #2.

**d. Edge cases NOT tested.** Three groups missing:
   - Malformed `MC_MAP_CFG` (e.g. `tiles: null`, `providers: 'string'`, `providers.osm.provider: 'mapbox'` with empty-string token)
   - `MC_setMapMode` / `MC_getMapMode` — zero coverage on the auto/light/dark mode-override surface (net-new public API, see must-fix #3)
   - `MC_createLayerControl` — zero coverage on the Leaflet layer-control wiring (net-new public API; tested behavior here is what the end user actually sees)
   - URL-template injection via untrusted token characters (see must-fix #1)
   - Tile network failure handling — out of scope for unit tests, fine to skip

**e. Test names describing behavior or implementation?** ✅ Behavior-shaped: `"OSM providers appear when osm.enabled=true"`, `"applyTileFilter clears filter when switching to native dark tile"`, etc. Good.

**f. Test setup more complex than the test?** ⚠️ Acceptable. `makeSandbox` + `loadProviders` is unavoidable for vm-sandboxed browser code, and each test has a focused single assertion afterward.

### Must-fix

1. **`test-issue-1420-tile-providers.js` lines ~96–115 — vacuous `for-of` over `Object.keys(reg)`.**
   The "Every registry entry has a url function or string with {z}" and "Every registry entry has a type of light or dark" tests iterate `Object.keys(reg)`. If `MC_initTileRegistry` ever returns an empty registry (all providers disabled by config), the loop runs zero times and the test passes vacuously.
   **Fix:** add `assert.ok(Object.keys(reg).length >= 9, 'registry must contain all 9 providers when all enabled')` at the top of each, plus an explicit IDs check via the `ALL_IDS` constant already defined at line 78. This makes the test actually capable of failing on an empty-registry regression.

2. **`public/map-tile-providers.js` — token concatenation lines 35–40 — add a URL-encoding test.**
   `_getOsmUrl` does `'?apikey=' + key`, `'?key=' + key`, `'?access_token=' + key`. `_getStamenUrl` does `'?api_key=' + _cfg.providers.stamen.token`. No `encodeURIComponent`. Operator-provided config is mostly trusted, but the public-facing risk is real: a token containing `&` or `#` produces a malformed URL silently. Either:
   - Add a test asserting `encodeURIComponent` semantics (preferred, then fix the prod code to wrap), OR
   - Add an explicit comment in `_getOsmUrl` documenting "tokens are trusted operator config; do not pass user-controlled strings" AND a test asserting the registry rejects/strips tokens containing `&`, `#`, `?`, or whitespace.

3. **Net-new public APIs ship with zero tests** — exemption requires the test to *land in the same PR*, not "ship untested if it's new."
   - `MC_setMapMode(mode)` + `MC_getMapMode()` — control auto/light/dark theme override. Add at minimum: round-trip persistence, rejection of invalid modes (e.g. `'sepia'`), `mc-map-mode-changed` event dispatch.
   - `MC_createLayerControl(map, autoLayerGroup, storageKey)` — actual user-facing Leaflet control. At minimum a smoke test: returns `null` when `L` is undefined (the early-return path on line ~210), and when stubbed `L`/`L.control.layers` is provided, exposes `_layerMap` populated from the enabled registry. Without this, the user-facing switcher could regress to an empty/broken dropdown and the suite stays green.

### Nice-to-have (non-blocking)

- `test-frontend-helpers.js` "mapTiles exposes map config directly to MC_MAP_CFG" — asserts `MC_MAP_CFG === mapCfg`. That's literally "we assigned the variable." Consider a richer assertion (e.g. `MC_initTileRegistry` is invoked after assignment so the registry reflects the new config) to make the test design-driving rather than reference-checking.
- Consider extracting `makeSandbox` to a shared test helper — it appears in two files now with slightly different shapes.

### Summary

| Question | Result |
|---|---|
| Tests fail on revert? | ✅ 23/28 — strong gating |
| Tautology-free? | ⚠️ Vacuous loop tests (#1) |
| Wrong impl rejected? | ⚠️ Empty registry passes (#1) |
| Edge cases covered? | ❌ Mode + LayerControl + injection untested (#2, #3) |
| Behavior-named? | ✅ |
| Setup proportional? | ✅ |

Requesting changes. Address #1, #2, #3 and the gate clears.

Kpa-clawbot (changes requested):
## Tufte Review (round 1) 

> *"Above all else, show the data. Clutter and confusion are not attributes of data — they are shortcomings of design."*

Reviewed via `gh pr diff` only. No checkout. Three must-fixes — all visual-integrity / layout collisions, not taste calls.

---

### MUST-FIX 1 — Layer control collides with `.map-controls` on `/map`

`public/map-tile-providers.js:1444` adds the control at `position: 'topright'`. Leaflet drops it into `.leaflet-top.leaflet-right` at the map's top-right corner.

But `/map` already owns that corner: `public/style.css:1386-1400` pins `.map-controls` (the gear/settings overlay) at `position: absolute; top: 12px; right: 12px; z-index: 1000; width: clamp(160px, 18vw, 240px)`. The new layer switcher renders **underneath / behind** that panel. Two pieces of chrome competing for the same pixels — the viewer sees one of them, never both, and which one wins depends on z-index quirks across browsers.

Fix: either move the layer control to `'topleft'` (Leaflet zoom is at topleft by default on `/map`? — verify; or move zoom to topright + layers topleft), OR collapse the layer selector into a row inside the existing `.map-controls` fieldset stack ("Map Tiles" section) so there is one control panel, not two. The latter is the Tufte answer — one panel, one grammar.

### MUST-FIX 2 — Mobile: layer control is unreachable

`public/style.css:2165` — `@media` rules make `.map-controls` full-width on phone (`width: calc(100vw - 24px); top: 8px; right: 12px`). That panel completely covers the top-right of the map, including wherever `.leaflet-top.leaflet-right` sits. Phone users cannot tap the new layer switcher at all.

Same fix as #1 resolves this — fold the selector into the existing controls panel rather than overlay it.

### MUST-FIX 3 — One flat list mixes Light + Dark providers with no grouping

`map-tile-providers.js:1416-1444` builds a single `baseMaps` object: `{ Auto, CARTO Light, OSM, CARTO Dark, Stamen Dark, ... }` and hands it to `L.control.layers`. The code comment says "natural grouping" but `L.control.layers` renders a flat radio list with no separator and no section headings. The viewer must memorize which entry is light vs dark; the data (light/dark capability) is hidden behind a label-guessing game.

Either:
- (a) Label entries explicitly: `"CARTO (light)"`, `"CARTO (dark)"`, `"OSM (light)"`, `"Stamen Toner (dark)"` — integrate the metadata with the choice. Tufte rule 9: labels belong on the data, not in a separate mental legend the user maintains.
- (b) Insert a Leaflet layer-control separator between the light and dark groups (`L.control.layers` supports this via an empty entry trick, or via a custom control). The diff already groups them in the iteration order — finish the job in the UI.

Option (a) is simpler and information-denser. Pick one.

---

### Acknowledged-good

- CSS vars throughout the new code (no hardcoded hex). ✓
- Cache buster already present at `public/index.html:126`. ✓
- Customizer "Map Tiles" section gated on `hasMultipleLight || hasMultipleDark` (`customize-v2.js`) — does not clutter when only one provider configured. ✓
- Auto entry is well-named (`"Auto (follows theme)"`) — labels on the data. ✓
- `.invert-map-filter` lives in a single CSS class with no inline duplication. Quiet. ✓

### Note (not must-fix)

`[data-theme="dark"] .leaflet-control-layers-toggle { filter: invert(1) brightness(2); }` (`style.css:+1627`) is a raster hack that will quantize the gear icon — fine pragmatically, but a recolored SVG would respect `var(--text)` and survive theme swaps cleanly. Out-of-scope for this PR.

---

**Verdict:** Request changes — three collisions/integrity issues above, all fixable without re-architecting.

Kpa-clawbot (changes requested):
Independent review (round 1) 

Cold review by an independent bot — no prior context. Diff scope matches the PR description (OSM/Stamen providers + per-provider Leaflet layer control on `live.js`/`map.js`). Below are issues I'd want fixed before merge. Two categories only: **must-fix** or **out-of-scope** (file an issue).

## Must-fix

1. **API tokens shipped verbatim to every browser.** `handleConfigClient` returns the entire `s.cfg.Map` (incl. `providers.osm.token`, `providers.stamen.token`). That is by design for tile keys, but `_comment_map_tiles` never warns operators their key is public — naive users will paste a default-unrestricted Mapbox/MapTiler/Thunderforest key. Add explicit guidance to apply referrer/origin restrictions on the provider dashboard, and consider returning only the URL-resolved string (server-side templating) instead of raw tokens. See `cmd/server/routes.go:358`, `config.example.json:76` and `public/map-tile-providers.js:35-43`.

2. **Attribution is wrong/illegal for the commercial OSM backends.** `osm-standard`/`osm-dark` use a hard-coded `'© OpenStreetMap contributors'` regardless of provider, but Mapbox, MapTiler, and Thunderforest TOS each require their own attribution alongside OSM. `public/map-tile-providers.js:54-55`.

3. **Disabling Carto in config does NOT stop hitting Carto.** When `carto.enabled=false` and OSM/Stamen aren't enabled either, `REGISTRY` becomes `{}`, but the fallback paths in `_resolveTileUrl` / `_liveResolveTile` still default to `TILE_DARK`/`TILE_LIGHT` constants which point at `basemaps.cartocdn.com`. Either fail loud, or honor the disable. `public/map.js:280-285`, `public/live.js:1336-1346`.

4. **Backward-incompat config break with no migration.** The `Tiles map[string]interface{}` field and `MapDarkTileProvider string` were removed from `cmd/server/config.go` and `cmd/server/types.go`. Existing operators with `"tiles": {"dark": "...", "light": "..."}` or `"mapDarkTileProvider": "..."` will silently lose those settings (Go json decoder ignores unknowns; `roles.js` no longer reads `cfg.tiles`). Either keep both shapes, log a startup warning on the old key, or call this out in CHANGELOG + add an upgrade note in the PR body. `cmd/server/config.go:66`, `cmd/server/types.go:995`, `public/roles.js:455-462`.

5. **Orphaned DOM reference.** `customize-v2.js:2027` reads `document.getElementById('cv2-osm-warning')`, but no element with that id is ever rendered in `_renderTileProviderSelector`. Either ship the warning element or delete the dead handler.

6. **Cross-tab map-mode sync is broken.** The `storage` listener in `map-tile-providers.js` filters with `_hasId(e.newValue)` for all three keys, but when `STORAGE_KEY_MODE` changes the new value is `'auto'|'light'|'dark'`, not a provider id, so `_hasId` returns false and the dispatch is skipped. Either branch on key type or split listeners.

7. **`_comment_` coverage gap (AGENTS.md mandatory rule).** `config.example.json:76` has one prose blob for the whole `map.tiles` block but never documents valid values for `darkDefault`/`lightDefault`, what `carto.domain` accepts (it's interpolated raw into a URL — operator-controlled but undocumented), or what happens when `enabled=true` with empty `token`. Per-field `_comment_` lines or an expanded comment listing every key + accepted values is required.

8. **Layer control rebuild side effects.** `MC_createLayerControl` calls `map.off('baselayerchange')` inside `_buildControl` which clobbers ANY other consumer that registered a baselayerchange handler on the same map. Use a named handler ref and `map.off('baselayerchange', _fn)` instead. `public/map-tile-providers.js:230`.

9. **Stamen/Stadia URL ignores `enabled` semantics.** `_getStamenUrl` always emits the Stadia URL with the token appended (if any), but even without a token Stadia requires a key for non-trivial usage. The label still says "Stamen Toner Lite" giving false confidence. Either require token when `stamen.enabled=true` (validate, refuse, log) or relabel.

10. **`_layerInstance` is a dead global.** Declared at `map-tile-providers.js:29`, set nowhere, only read inside `applyTileFilter`. Either wire it up (so theme flips refresh the explicit non-Auto layer) or remove it — currently the explicit layer keeps a stale URL after the operator changes config in another tab and storage event fires.

## Out-of-scope (file separate issues)

1. No test asserts that `/api/config/client` redacts (or doesn't) provider tokens — coverage gap worth a follow-up after the redaction decision in must-fix #1 is made.
2. `docs/deployment.md` has a 330-add/351-del churn unrelated to this feature; worth confirming in a separate PR that it isn't a stale rebase artifact dragging old content back.

**Verdict:** request changes.

EldoonNemar:
Buzzkill. 🤣

Kpa-clawbot (changes requested):

**Verdict: NEEDS-WORK**

### TDD history check: FAIL (partial exemption available but not declared)

Branch `FR-add-maptiles` has 2 non-merge commits ahead of master:

1. `552df03` — `fix(ui): let live filters break out of container` — production-only (live.css/live.js), unrelated to tile providers.
2. `7c313b6` — `fix(#1165): multi-provider map tile selector with OSM/Stamen support` — production + tests in the **same commit**.

No red commit. CI was only recorded on the merge commit (success) — there is no failing-test commit on this branch to prove the new assertions actually gate behavior.

The "Net-new UI surfaces" exemption (AGENTS.md) covers the *new* providers (OSM, Stamen, light-tile helpers, `MC_initTileRegistry`, `MC_MAP_CFG` plumbing). But this PR also **renames/restructures previously-tested behavior**:
- `voyager-inverted` → `carto-voyager-dark`
- `positron-inverted` → `positron-dark`
- `esri-darkgray-labels` removed entirely

Those are changes to *existing* surface — they need either a red→green history or an explicit exemption justification in the PR body. The PR body currently has neither. Per AGENTS.md TDD section: "Exemptions require explicit justification in PR body."

### Six Questions — flagged

**(c) Could a wrong implementation pass?** The `roles.js` MC_MAP_CFG test uses `assert.deepStrictEqual(JSON.parse(JSON.stringify(ctx.window.MC_MAP_CFG)), ...)` — round-tripping through JSON strips non-enumerable properties, Proxies, and getters. If MC_MAP_CFG were intended to be a live reference and got mutated via a getter, this assertion would silently pass. Compare against the actual object reference.

**(d) Edge cases not tested:**
- **Stamen URL generation with/without token is completely untested.** OSM has two tests (`OSM falls back to standard...`, `OSM uses Maptiler/Thunderforest URL when token provided`) — Stamen has *zero* URL-generation tests despite being a first-class configured provider with the same token-vs-no-token branching. This is the most concrete gap.
- No test that `MC_initTileRegistry(true)` actually causes an existing Leaflet map to re-sync its tile layer — the dispatch event is asserted, but no consumer-side reaction is verified. The dispatch test is therefore close to a tautology of the producer.
- `roles.js` fetch failure / non-JSON response — `MC_MAP_CFG` left in an undefined state if `cfg.json()` throws. No test.

### Must-fix

1. **PR body, lines 1–8** — Add explicit TDD exemption justification per AGENTS.md ("Net-new UI surfaces" for OSM/Stamen + light-tile helpers; cite the rename of `voyager-inverted`→`carto-voyager-dark` etc. as a *renaming* and explain why existing tests being rewritten is acceptable). Without this, axis 4 of pr-merge-gate blocks.
2. **`test-issue-1420-tile-providers.js`** — Add Stamen URL-generation tests symmetric with OSM (no-token fallback path AND token-provided path). Stamen is a config-gated provider with a token field; the asymmetric coverage means the Stamen-with-token code path is unproven.

Everything else is out of scope for this round.

— Kent Beck gate

Kpa-clawbot (changes requested):
## Independent review (round 1) — cold adversarial pass

Reviewed full diff via `gh pr diff 1533`. Did not check out the branch. The mission as written is "add OSM/Stamen tile providers with per-provider Leaflet layer control" — the diff delivers that plus a large amount of scope creep that needs to be acknowledged before merge. Marking **NEEDS-WORK** — most items are small/mechanical but several are user-visible breakages or dead code paths.

### Must-fix

1. **Breaking config schema, no migration.** `Config.MapDarkTileProvider` and `Config.Tiles` are removed from `cmd/server/config.go` / `types.go` / `routes.go`, replaced with a new `Map` field. Operators whose existing `config.json` has top-level `"mapDarkTileProvider"` or `"tiles": {...}` will have those values silently dropped — Go's JSON decoder ignores unknown keys, so there's no warning. Add a one-release deprecation: read the old keys and copy into the new shape with a stderr `[deprecated]` log line.

2. **`cfg.tiles.dark` / `cfg.tiles.light` URL overrides regressed.** `public/roles.js` previously did `if (cfg.tiles.dark) window.TILE_DARK = cfg.tiles.dark` — that block is deleted with no replacement. Operators who customized the base tile URLs via the old `tiles.dark`/`tiles.light` server config will silently fall back to hardcoded CartoCDN URLs. Either preserve the override path under `cfg.map.tiles.darkUrl/lightUrl` or document the removal in the PR body and `config.example.json`.

3. **`config.example.json` is missing the old keys with a deprecation note.** Since this is the file operators read first (per `AGENTS.md` Config Documentation Rule), the removed `mapDarkTileProvider` / `tiles` keys should appear with `"_deprecated_..."` comments pointing at the new `map.tiles.*` shape, not be deleted outright.

4. **`docs/deployment.md` is a +330/-351 total rewrite — scope creep.** This PR's title says "OSM/Stamen tile providers"; rewriting the entire deployment guide (removing the "Configuration Reference", "API Documentation", "Monitoring & Health Checks", "Backup & Restore" sections, replacing them with a Quick Start–style narrative) is unrelated work. Split into a separate PR so it can be reviewed on its own merits. As written this PR silently deletes operator-facing reference documentation.

5. **Dead code: `_layerInstance` in `public/map-tile-providers.js`.** Declared at the top of the IIFE, referenced only inside `applyTileFilter()` to call `.setUrl()`, but **never assigned anywhere in the diff**. The `if (_layerInstance && p)` block is unreachable. Either wire it up (from `MC_createLayerControl`) or delete the dead branch.

6. **Dead code: `setMapMode` / `getMapMode` / `mc-map-mode-changed` event.** Exposed on `window.MC_setMapMode` / `MC_getMapMode` and dispatch a custom event, but no caller in this diff uses them and no UI surface is added. Either wire them into the customizer or remove until the consuming feature lands.

7. **Dead code: `.invert-map-filter` CSS class in `style.css:2298`.** Added but never applied to any element in the diff. Drop it.

8. **Dead reference: `cv2-osm-warning` in `customize-v2.js:2027`.** `_renderTileProviderSelector` never emits an element with id `cv2-osm-warning`, but the change handler does `document.getElementById('cv2-osm-warning')` and toggles its display. Either render the warning element (with the rate-limit caveat operators should see for OSM) or drop the toggle.

9. **`MC_createLayerControl._buildControl` does `map.off('baselayerchange')` (no fn arg).** This clears **all** `baselayerchange` listeners on the map, not just its own. If any future code (or third-party plugin) registers a listener on this event, it will be silently unregistered on rebuild. Capture the bound handler in a local and `map.off('baselayerchange', handler)` instead.

10. **OSM/Stamen tokens injected into URL query strings without encoding.** `_getOsmUrl` / `_getStamenUrl` concatenate `_cfg.providers.*.token` directly via the query string. A token containing `&`, `=`, `?`, or a space will produce a broken URL. Use `encodeURIComponent(token)`.

11. **Security note absent re: tile-provider token exposure.** Mapbox/Maptiler/Thunderforest/Stadia tokens supplied in server config end up in client-side tile URLs (visible in DevTools Network, referrer headers, browser cache). This is industry-standard for these services, but the config comment in `config.example.json` should warn operators: "tokens are sent to the browser — use referrer-restricted/rate-limited keys, not master keys." The current comment only mentions rate limits, not key-exposure semantics.

12. **Mapbox URL template is wrong shape.** `_getOsmUrl` returns `https://api.mapbox.com/styles/v1/mapbox/streets-v11/tiles/{z}/{x}/{y}?access_token=...` — Mapbox's raster-tiles endpoint requires a tile size segment (`/tiles/256/{z}/{x}/{y}` or `/tiles/{z}/{x}/{y}@2x`). The current template returns 404 on every tile. Verify against [Mapbox Static Tiles API docs](https://docs.mapbox.com/api/maps/static-tiles/) and add a test that actually fetches one.

13. **`live.css` comment regression.** Old comment was `/* Region/Area filters inline in live header controls body */`, new is `/* Region filter (#1045) inline in live header toggles */` — but the CSS rule on the next line still targets `.live-area-filter-container`. The comment now misleads anyone scanning by area-filter. Restore the "/Area" reference or explain.

14. **Test coverage drop: `MC_setDarkTileProvider rejects unknown IDs` no longer asserts `no dispatch`.** The original test asserted both no persistence AND `ctx.events.length === 0`. The rewrite drops the dispatch assertion. Reinstate — silent dispatch on invalid IDs is a real bug class.

15. **Test coverage drop: cross-tab event no longer asserts `filter re-applied`.** Original test asserted `ctx.tilePane.style.filter.indexOf('invert(') >= 0` after the cross-tab event; rewrite removes that. Reinstate.

16. **`MC_initTileRegistry` is called once at parse time then again from `roles.js` after config fetch — but `MC_createLayerControl` is invoked from `map.js`/`live.js` synchronously at map init.** First build runs with carto-only REGISTRY; second build is triggered by the `fromConfig` event. Document this two-phase init behavior with a comment in `MC_createLayerControl`, or guard against the layer control being constructed before config is loaded (e.g. wait for `MeshConfigReady`).

17. **`MC_setLightTileProvider` / `MC_setDarkTileProvider` dispatch the same event name (`mc-tile-provider-changed`)** with `detail.type` carrying the light/dark hint — but `_buildControl`'s rebuild listener only checks `detail.fromConfig`, and `_syncDarkTiles` in map.js/live.js doesn't filter by type either. So changing the light provider while in dark mode triggers a dark resync (and vice versa). Add `if (detail.type && detail.type !== currentMode) return;` early-outs, or split the event names.

18. **`_renderTileProviderSelector` early-returns when both `hasMultipleLight` and `hasMultipleDark` are false** — but with only OSM enabled (no Carto, no Stamen) you get exactly 1 light (`osm-standard`) and 1 dark (`osm-dark`), so the picker disappears. That's surprising UX given the operator went out of their way to enable a provider. Show the section anyway, or render the single option as a read-only label.

19. **`getActiveId` / `getActiveLightId` filter stored localStorage value by `REGISTRY[stored].type`** — if the user previously chose a provider that the operator later disabled in config (e.g. picked `stamen-toner-lite` then operator flipped `stamen.enabled` to false), `REGISTRY[stored]` is undefined and `.type` throws TypeError. `_hasId(stored)` guards the property check but the `&& REGISTRY[stored].type === 'dark'` follow-up assumes `_hasId` passed — it did, so safe — but worth a defensive `REGISTRY[stored] && REGISTRY[stored].type === ...` for grep-safety in future refactors.

20. **No browser validation evidence in PR description.** Per project rule 2 ("No commit without browser validation"), tile-provider changes that touch Leaflet layer lifecycle need a screenshot of the layer control in light + dark mode against a running server with OSM/Stamen enabled. PR body doesn't include any. Add one before merge.

21. **PR body says "Updated tests (28) to address the changes made to the maps"** — that's a count, not a description. List what behavior the new tests cover and which existing tests were removed/replaced. The diff shows ~20 new tests and several deleted assertions; reviewers shouldn't have to diff to figure out the scope.

22. **`hop-resolver.js` / `nodes.js` / `customize.js`** are not in this diff but `customize-v2.js:2027` references `cv2-osm-warning` and `customize-v2.js:1418` references `MC_getLightTileProvider` — if any non-v2 customizer is still loaded for any user (legacy code path), those window globals exist but the UI element doesn't. Confirm `customize-v2.js` is the only customizer surface, or wire the equivalent in the legacy customizer.

23. **`AreaFilter` test scaffolding added in `test-frontend-helpers.js`** (3 separate spots) but no production code in this PR touches `area-filter.js`. Why? Either explain in PR body or remove the unrelated test additions (scope creep).

### Out-of-scope

- Full deployment.md rewrite (see #4 — should be its own PR, but cannot literally be "fixed" without splitting this PR).
- Issue #1165 itself ("bot complaints") — PR title acknowledges this is a partial fix. That's fine but the open complaints should be enumerated in the PR body so reviewers know the residual surface area.

### Verdict

NEEDS-WORK. Most must-fix items are 5-minute mechanical fixes; the schema-migration (#1), docs scope (#4), and Mapbox URL (#12) are the substantive ones. The dead code (#5, #6, #7, #8) suggests the diff hasn't been self-reviewed for unreferenced symbols — a grep pass before re-pushing would catch them all.

Author is community/receptive — a brief follow-up comment with the rationale on #1/#4 will likely land cleanly.

Kpa-clawbot (changes requested):
## Tufte Review (round 1) — PR #1533

**Cold review.** Inspected PR diff (no checkout). The new layer-control was rendered against the actual `public/map-tile-providers.js` + `style.css` from `e25655c` in an offline harness with Leaflet 1.9.4 (staging has not yet deployed this PR — `MC_TILE_PROVIDERS` on `analyzer-stg.00id.net` still shows the four legacy IDs, no `MC_createLayerControl`). Verified the control renders, opens, and lists nine baselayers + an "Auto" meta option.

> *"Above all else, show the data. Clutter and confusion are not attributes of data — they are shortcomings of design."*

The layer button is a real win — the user previously had no in-app way to pick a provider. But the diff ships nine flat radio choices, design variation masquerading as data variation, dead code paths, and CSS hacks. Below: what to fix before merge.

### Must-fix

1. **`public/map-tile-providers.js:48-56` — design variation masquerading as data variation.** The "(Dark)" entries for `carto-voyager-dark`, `positron-dark`, `osm-dark`, `stamen-toner-dark` are not different tilesets — they are the *same* tiles with `filter: invert(1) hue-rotate(180deg) ...`. Listing them alongside true-dark tiles (`carto-dark`) and true-light tiles (`carto-light`) as siblings tells the reader these are nine independent things; they are not. Either (a) collapse the inverted variants out of the picker and surface them only as an implicit fallback when a light provider is chosen under a dark theme, or (b) annotate the label so the user knows the difference: `Carto Voyager (CSS-inverted)`. The current "(Dark)" label is misleading.

2. **`public/map-tile-providers.js:48-56` — nine flat radio choices is chartjunk by quantity.** On the 390×844 mobile viewport the expanded control occupies ≈30% of screen height (screenshot below). Tufte: shrink and simplify. Group by provider with a separator (Leaflet supports `overlayMaps` separators, or two `L.control.layers` instances). Better still: two compact dropdowns — *Light provider* / *Dark provider* — which is already the model in `customize-v2.js` (`_renderTileProviderSelector`). Having the same choice surfaced twice (in customizer AND in the map layer control) is a third axis of confusion — pick one home for the control.

3. **`public/map-tile-providers.js:259-262, 284-290` — "Auto" is a meta option presented as a peer.** "Auto (follows theme)" sits as the first radio of nine siblings. It is not a provider; it is a *policy* (let the theme decide). When Auto is selected, the user cannot see *which* concrete provider is currently in use — the choice the system is making is hidden behind the meta label. Two fixes: (a) visually separate Auto from the provider list (Leaflet `<hr class="leaflet-control-layers-separator">`), and (b) when Auto is the active selection, append the resolved provider name in parentheses to the Auto label, or check both radios (Auto + resolved) so the user can read state without thinking.

4. **`public/style.css:2295-2297` — `filter: invert(1) brightness(2)` on `.leaflet-control-layers-toggle` is chartjunk masquerading as theming.** This is the exact pattern AGENTS.md prohibits ("all colors via CSS variables"). You are inverting a raster icon to fake a dark-mode color; the magnitudes (`brightness(2)`) are arbitrary, the result is washed out, and it breaks if Leaflet ever ships an SVG/mask icon. Replace with a proper CSS class swap (`background-image: url(...)`) keyed off `[data-theme="dark"]`, or override the toggle via a CSS mask + `background-color: var(--text)`. Either way the magic-number filter goes away.

5. **`public/style.css:2299` — dead CSS.** `.invert-map-filter { filter: invert(1) hue-rotate(180deg) brightness(0.9) contrast(1.05); }` is defined but never referenced from JS (JS writes the same string inline via `pane.style.filter = INVERT_CSS`). Also: it duplicates the `INVERT_CSS` JS constant — two sources of truth for the same magic value. Remove the CSS rule or use it from JS via `pane.classList.add('invert-map-filter')`.

6. **`public/style.css:2286-2297` — light-mode styling of the layer control is missing from the diff.** You added `[data-theme="dark"]` overrides for background / border / separator / toggle, but the default (light) Leaflet styles assume a light page; in CoreScope light mode the control still uses raw Leaflet defaults instead of `var(--card-bg)` / `var(--border)`. If dark needs the variable hookup, light needs it too — otherwise hovering over `--card-bg` and `--border` is half-applied, and a brand that customizes the light palette will see the control diverge from the surrounding chrome. Either drop the `[data-theme="dark"]` qualifier and let variables do their job in both themes, or add a parallel `[data-theme="light"]` (or unqualified) block.

7. **`public/customize-v2.js:2027-2028` — handler for a non-existent element.** `var warn = document.getElementById('cv2-osm-warning');` is never created in this diff (search the diff: zero matches for `cv2-osm-warning` outside this lookup). The line is a no-op forever. Either add the warning element (recommended — OSM without a token hits the standard OSM tile policy and is genuinely risky for an analyzer deployment) or remove the dead handler.

8. **`public/map-tile-providers.js:359-373` — cross-tab `STORAGE_KEY_MODE` listener is dead code.** Line 361 admits `STORAGE_KEY_MODE` events through the first guard; line 362 then immediately rejects via `if (!_hasId(e.newValue)) return;`. Mode values are `auto` / `light` / `dark` — never a registry id — so a mode change in tab A is silently ignored in tab B. Either branch the handler (mode events dispatch `mc-map-mode-changed`, not `mc-tile-provider-changed`) or drop `STORAGE_KEY_MODE` from the guard. Behavior gap, not Tufte, but it is in the diff.

9. **`cmd/server/config.go:101-104` + `cmd/server/types.go:1006-1009` — silent config-field removal.** The `MapDarkTileProvider` top-level field is deleted from `Config` and `ClientConfigResponse`. Operators with the legacy field in `config.json` will silently lose their setting after upgrade with no migration path. Either keep a deprecated alias that copies into `Map.tiles.darkDefault` at load time (preferred) or add a startup log warning. Documenting it in the example file is not enough — operators don't re-read examples on upgrade. Tufte's adjacent principle: *don't quietly mutate the data the user gave you.*

10. **`docs/deployment.md` (~330+/~351−) — scope creep.** This PR is "add tile providers UI"; the doc diff is a top-to-bottom rewrite of the deployment guide. Two effects: (a) reviewers can't see the tile-provider doc changes inside a wall of unrelated edits, and (b) any deployment-guide regression introduced here is attributed to a UI PR. Split the deployment.md rewrite into its own PR; in this one, the documentation delta should be the ~30 lines that explain the new `map.tiles.providers` config block.

11. **`public/live.js:1373-1377` + `public/map.js:308-315` — duplicated initialization (`MC_createLayerControl(map, ..., 'meshcore-live-map-selection')` vs `'meshcore-map-selection'`).** Two near-identical 10-line blocks; the only behavioral difference is the `storageKey` argument, which `map-tile-providers.js:288` accepts but never uses (`_buildControl` reads from `getActiveId()` / `getActiveLightId()` directly, both of which use the module-level `STORAGE_KEY*` constants). Either delete the unused `storageKey` parameter, or implement per-map-page persistence and document why. Currently it is a vestigial argument.

12. **`config.example.json:74` — single-line `_comment_map_tiles` is 700 chars of run-on prose.** Operators reading the example get a wall of text covering Stamen, three OSM providers, free-tier limits, and registration URLs in one breath. Tufte: integrate words with the structure they describe. Break this into per-provider comments (`_comment_carto`, `_comment_osm`, `_comment_stamen`) sitting next to the field they document, so a reader scanning the OSM block sees the OSM context without parsing a paragraph.

### Out-of-scope

- The customize-v2 tile selector pre-existed; this PR extends it (light + dark dropdowns). The customizer-vs-layer-control split (must-fix #2) recommends choosing one home; I'm not asking you to delete either in this PR.
- Test reorganization in `test-issue-1420-tile-providers.js` is mechanical (renaming IDs, adding new gating tests) and reads fine.

### Verdict
**NEEDS-WORK** — 12 must-fix. The core idea (multi-provider picker, server-driven enablement) is right. The execution has three classes of problem: (a) too many flat choices presented as peers when they are not, (b) CSS hacks instead of variable discipline, (c) dead code and scope creep.

### Verification artifacts
- Staging (`http://analyzer-stg.00id.net/`) still on legacy registry — PR not deployed; CDP confirmed `Object.keys(window.MC_TILE_PROVIDERS) === ['carto-dark','esri-darkgray-labels','voyager-inverted','positron-inverted']` and `typeof window.MC_createLayerControl === 'undefined'`. Called out per brief.
- PR JS loaded into the staging page via injected `<script>`; `MC_createLayerControl` confirmed present, 5 Carto-only providers (no `MC_MAP_CFG` → osm/stamen correctly gated off).
- Standalone harness (`/tmp/test.html`) at 780×500 desktop and 390×844 mobile: layer control opens, 1 meta + 4 light + 4 dark + 1 carto-dark = 9 radios + Auto. Mobile expanded menu height ≈ 250px on a 844px viewport (≈30%).
- Light-mode flip confirmed control retains its dark-mode CSS-variable styling (because the override is `[data-theme="dark"]`-only) — the default Leaflet white styling never gets the `var(--card-bg)` / `var(--border)` treatment in light mode.

— *Tufte (round 1)*

EldoonNemar:
Big bucks, no whammies!

EldoonNemar:
~There is still a bug where if you select a dark version map in the leaflet control and go back to a light version it stays dark until you hit the light/dark theme changer. Looks like I need to update my tests as well. I thought I had...~
Found and fixed in next push

EldoonNemar:
Question for the bots @Kpa-clawbot 

Can you clarify for me if you want me to change the customizer->display light and dark theme defaults if a user selects a dark or light theme within the leaflet map? That's what I'm doing now, but it feels wrong.

Kpa-clawbot (changes requested):
## Independent review (round 2) — adversarial pass against 7efa519

Cold review, no checkout — `gh pr diff 1533` + `git show pr1533-readonly:<path>` for full-file context. I deliberately avoided round-1 findings before forming opinions; overlap is coincidental.

**Verdict: NEEDS-WORK.** The OSM/Stamen feature is in good shape — the migration plumbing, async config dispatch, and the layer-control UX are solid additions. But the diff has dead code, a stale module header, a noisy default-startup warning, a breaking silent removal, and one piece of scope creep (the deployment-doc rewrite) that should not ride along.

### Must-fix

1. **`docs/deployment.md` is gutted and rewritten** (+330/−351). This is scope creep against the PR title `fix(#1165): add OSM/Stamen tile providers`. The rewrite drops the image-tag matrix (`v3.4.1` / `v3.4` / `v3` / `edge`), the `/api/spec` + `/api/docs` section, Monitoring & Health Checks, the structured Backup & Restore commands, and the troubleshooting entries for "database locked", "high memory", and ARM/RPi. Some of the new content is nicer prose, but losing the API-spec discoverability and the image-tag guidance is a real regression. Split this into its own PR (or revert it here and let the tiles change ship with a 1-paragraph tiles addendum to the existing doc).
2. **`cmd/server/config.go:357,389` — `migrateDeprecatedConfig` warns on every default install.** It sets `migrated = true` whenever `c.Tiles != nil`, but the new `config.example.json` ships `"tiles": {}` — an empty-but-non-nil map. Result: every fresh deploy that copies the example file logs `[deprecated] Top-level 'mapDarkTileProvider' and 'tiles' keys ... are deprecated`. Fix: only set `migrated = true` if `len(c.Tiles) > 0`, and similarly only when `c.MapDarkTileProvider != ""`.
3. **`config.example.json:81-82` — remove the deprecated keys from the example.** Shipping `"mapDarkTileProvider": ""` and `"tiles": {}` defeats the deprecation: it pushes operators to keep using them by example and (per #2) triggers the warning every restart. Delete both lines; the `_comment_deprecated_tiles` note above already covers the migration story.
4. **`public/map-tile-providers.js:1-14` — file header is stale.** Still says "4 providers: carto-dark (default), esri-darkgray-labels (base+ref), voyager-inverted, positron-inverted." Esri is gone, voyager/positron got renamed, and 5 more providers (osm × 2, stamen × 2, carto-light, carto-voyager) were added. Future readers will misroute on this comment.
5. **`public/map-tile-providers.js:27,169-173` — `_layerInstance` is dead code.** Declared at top, read in `applyTileFilter` for an early-`setUrl` shortcut, never assigned anywhere in the diff or the file. Either wire it up (and document what owns it) or delete the var + the dead branch — right now it's a footgun for the next contributor.
6. **Silent breaking removal of `esri-darkgray-labels`.** The provider was in the v3.4.x release and is documented as a valid value for `mapDarkTileProvider`. Any user who selected it via the customizer has `mc-dark-tile-provider=esri-darkgray-labels` in `localStorage`; on the next visit `_hasId('esri-darkgray-labels')` returns false and they silently fall back to carto-dark. Either keep esri in `BASE_STYLES` (gated `provider: 'esri'`, on by default like carto) or add a one-shot `localStorage` migration block.
7. **`public/map-tile-providers.js:124-134` — `setActive(id, type)` accepts type/id mismatch.** Nothing checks `REGISTRY[id].type === type`, so `setActive('carto-dark', 'light')` succeeds, writes `carto-dark` to `mc-light-tile-provider`, and `getActiveLightId` then silently rejects it on next read (you have a test for the read-side rejection but none for the write-side acceptance). Add a guard: `if (REGISTRY[id].type !== type) return false;`.
8. **`public/map-tile-providers.js:200-360` — `MC_createLayerControl` is wholly untested.** The new Auto-mode semaphore (`data-explicit-layer` attribute on `.leaflet-tile-pane`), the baselayerchange handler, and the rebuild-on-config-arrival path all have zero coverage. Given how fragile the implicit DOM-attribute coupling is, at least one test should: select an explicit layer → assert attribute set, select Auto → assert attribute cleared, then re-init registry and confirm the previously-selected explicit layer is re-selected after rebuild.
9. **The Auto-mode "lock" via `tilePane.setAttribute('data-explicit-layer', 'true')` needs a code comment.** It's a cross-module protocol (the tile-pane attribute is read by `applyTileFilter` to early-return) but there is no comment in either function explaining the contract. A 2-line note at each read/write site is the difference between "obvious" and "removed by next refactor."
10. **`config.example.json:64-65` — Carto enterprise domain shape is undocumented.** `_comment_carto` says "specify 'domain' for Carto enterprise" but the URL builder does `'https://{s}.' + domain + '.cartocdn.com'`. That requires a specific sub-segment, not a full hostname; nobody will guess this. Either accept a full URL prefix or spell out the expected format in the comment.
11. **`MC_DARK_TILE_DEFAULT` public export was removed** (`window.MC_DARK_TILE_DEFAULT = DEFAULT_ID;` in old file, gone in new). It was a public surface; deletion with no deprecation period can break consumers (theme.js variants, custom branding scripts). Either restore it or note the removal in the PR body / changelog.
12. **`public/style.css:2295-2306` — Leaflet zoom-control override is no longer scoped.** Diff removes the `.live-page .leaflet-control-zoom a` rules from `live.css` and re-adds them as unscoped `.leaflet-control-zoom a` in `style.css`. That changes the zoom-button appearance on the **map** page too (which previously used Leaflet defaults). Either intentional and ship a note, or re-scope to the pages that need it.
13. **`public/customize-v2.js:1418-1466` — inconsistent indentation.** The new `_renderTileProviderSelector` body uses 4-space indents inside the `if (lightIds.length > 0)` and `if (darkIds.length > 0)` blocks while the rest of the file is 2-space. Also a stray double blank line at the end of the function. Run a formatter pass.
14. **`cmd/server/types.go:1008` — `MapDarkTileProvider` is still emitted to clients but read by nobody.** `public/roles.js` was switched over to `cfg.map`; the field on `ClientConfigResponse` now exists only for back-compat that nothing consumes. Add a `// TODO: remove after vX.Y.Z` so it actually gets cleaned up rather than living forever.
15. **`migrateDeprecatedConfig` warning message wording.** "will be ignored in a future release" — name the release or link an issue, otherwise operators have no actionable signal.
16. **`docs/deployment.md` (new) describes a `./manage.sh setup` workflow** that isn't part of this PR and conflicts with point (1) above. If the doc rewrite stays, scrub references to scripts/features not in this branch.

### Out-of-scope

1. **The `area-filter.js` / `AreaFilter` stub additions in `test-frontend-helpers.js`** look like upstream-merge fallout, not this PR's scope. Confirm with `git log --oneline 7efa519 -- test-frontend-helpers.js | head -3` — if it's from the master merge, fine, leave it.
2. **The `live.js` `pruneStaleNodes` test rewrite** (markers use `getElement().style.opacity` now) — same situation, source change came from master merge, not from this PR. Mention in PR body for clarity.
3. **`public/map.js:_renderMarkersInner` null-guard for `markerLayer.clearLayers()`** — unrelated bug fix bundled in. Tiny, not worth ripping out, but call it out in the PR body.

### What's good (round 2)

- The `_cfg`-driven `BASE_STYLES` + `MC_initTileRegistry(fromAsync)` split is the right shape — registry is rebuilt deterministically once config arrives, and the `fromConfig: true` dispatch is a clean hook for `map.js`/`live.js` to re-sync.
- URL-as-function pattern (Carto domain, Stamen token, OSM provider switching) keeps the registry data-driven and is straightforward to extend.
- Server-side `migrateDeprecatedConfig` reading-then-writing migrated values into `Map.tiles` is a good back-compat play (modulo the #2/#3 noise).
- Test additions for OSM/Stamen URL shape, encode-safety, and registry gating are exactly the right level of coverage.

---
**HEAD reviewed:** `7efa519`
**Method:** `gh pr diff 1533` + `git show pr1533-readonly:<path>` for full-file context. No checkout.

EldoonNemar:
TDD Exemptions added

EldoonNemar:
@Kpa-clawbot I disagree with 
`Auto (follows theme) is initially-checked but ambiguous. When a user later picks an explicit provider then returns to Auto`

If a user is just flipping through maps, they don't want to update their customize setting, they just want to flip through them. If they want to update their default, they'll go to customize.  We shouldn't be force changing their default light and dark themes while they're browsing map tiles because it will confuse the user when they go back to auto thinking it'll be the default dark or light theme they dug through customize to configure, and they end up get tingthe last theme they were browsing for either dark or light.  It confused me, so it'll confuse everyone..

I'm removing that feature.

EldoonNemar:
I've removed `partial fix, pending bot complaints` and changed the title to `Fixes #1165`, in order to merge before you make a million changes that I have to merge 🤣

EldoonNemar:
Actually it looks like the selector is now broken. one moment

EldoonNemar:
now it's fixed

Kpa-clawbot (approved):
**Round 6 re-review** — delta-only

Scope: commit `2f1037fe` (`public/map-tile-providers.js` +6/-3). Round-5 verdict (APPROVED CLEAN @ `339d6e6`) carries forward for everything else.

### Delta verdict: ✅ correct fix

The regression was the DOM-surgery loop calling `span.innerHTML = ...` on `labelEl.querySelector('span')` — which in Leaflet 1.9.x grabs the **outer** span wrapping `<input type="radio">` + inner text span, wiping the radio input and rendering the base-layer control unselectable. The new code targets `spans[spans.length - 1]` — the inner text span — leaving the input intact.

### CDP verification (local chromium)

Loaded Leaflet 1.9.4 in the staging tab, built a `L.control.layers({...})`, then ran the **post-fix** DOM-surgery loop verbatim. Results:

- Leaflet label structure confirmed: `<label><span><input ...><span> NAME</span></span></label>` — 2 spans per label, outer contains input, inner contains text.
- After `spans[length-1].innerHTML = ' ' + label`: radio inputs **preserved** (`radioCount: 2`), `data-tile-id` correctly set from original text.
- Programmatic click on second radio flips `checked` state correctly → Leaflet event bindings intact.

### Findings on delta
- 0 BLOCKER
- 0 MAJOR
- 0 MINOR

(Could nit on the trailing whitespace after `spans[spans.length - 1];` and the bare `catch (_)` swallowing this exact class of error in future, but both are pre-existing or out-of-scope for a hotfix delta.)

### Action
Approving. Per the PR title ("partial fix, pending bot complaints") and prior author signal deferring squash-merge, **NOT** triggering auto-merge — left for the operator.

EldoonNemar:
@Kpa-clawbot The title was changed you crazy bot. It no longer says "partial fix, pending bot complaints". You can merge.

PR #1537, opened 2026-06-03
fix(security): escape mesh node names before HTML render — stored XSS (#1536)

### LINKED ISSUE #1536: Stored XSS: mesh node names (adv_name) rendered unescaped into the dashboard DOM
## Summary

CoreScope renders mesh-advertised **node names** (`adv_name`) and observer names into the dashboard DOM **without HTML-escaping** in several places, allowing **stored XSS**. This is the same vulnerability class as the publicly disclosed MeshCore dashboard XSS (CVE-2026-45323, CVSS 9.6) affecting `meshcore-card` and analyzer sites — see https://mxsasha.eu/posts/meshcore-xss-home-assistant/ which explicitly names CoreScope deployments as affected.

## Vector

`adv_name` is a 32-byte string in MeshCore advert packets with **no character validation at the protocol level**. An attacker only needs radio proximity to any mesh member. A name like `` (28 bytes) is stored and later rendered into `innerHTML` / Leaflet popups, executing script in the victim's browser session.

The Go backend does **not** mitigate this: `sanitizeName()` (`cmd/server/decoder.go`, `cmd/ingestor/decoder.go`) only strips control chars `< 0x20` and DEL — it keeps `< > " & /`, so the payload reaches the API and DB intact. The correct fix is output-encoding in the frontend (escape at the sink), matching the upstream `meshcore-card` v0.3.3 fix.

## Affected sinks (unescaped attacker-controlled name → HTML)

- `public/app.js` — global search dropdown: node name + channel name
- `public/nodes.js` — nodes table row name; node-detail Leaflet map popups (×2)
- `public/observers.js` — observers table name cell
- `public/packets.js` — observer-name cells (`obsNameOnly`, ×4) and the observer multi-select checkbox label
- `public/live.js` — node-filter `<option>` and map marker tooltip
- `public/analytics.js` — topology map node tooltip
- `public/route-view.js` — hop/union marker tooltips (×2)
- `public/area-map.html` — node popups (×2; file had no escaper)

Map popups via `map.js` (`safeEsc`), the live feed text (`escapeHtml(preview)`), packet detail text, channel messages, and `route-render.js` popups were already escaped — these are the gaps.

## Fix

Escape every attacker-controlled name/observer string at the HTML interpolation point using the existing global `escapeHtml()` (and a small local helper for the standalone `area-map.html`). Add a regression unit test for `escapeHtml` covering the CVE payload.

Fix in progress — PR incoming.

efiten:
## This PR fixes the stored XSS in full (closes #1536)

Mesh-advertised node names (`adv_name`) and observer names were rendered into the dashboard DOM **without HTML-escaping** in multiple places — the same class as the publicly disclosed MeshCore dashboard XSS (CVE-2026-45323). `adv_name` has no protocol-level validation and the Go `sanitizeName()` keeps `< > " &`, so a payload like `` reaches the frontend intact and executes.

**I audited every name/sender/text/channel render in `public/` and this PR escapes all unescaped sinks. There are no known remaining XSS sinks of this class after this change.**

### Sinks fixed (all escaped via the existing global `escapeHtml`, plus a local helper for the standalone `area-map.html`)

| File | Sink |
|------|------|
| `app.js` | global search dropdown — node name + channel name |
| `nodes.js` | nodes-table row name; node-detail Leaflet popups (×2) |
| `observers.js` | observers-table name cell |
| `packets.js` | observer-name cells via `obsNameOnly` (×4) + observer multi-select checkbox label |
| `live.js` | node-filter `<option>` + map marker tooltip |
| `analytics.js` | topology map node tooltip |
| `route-view.js` | hop + union marker tooltips (×2) |
| `area-map.html` | node popups (×2) — added a local `escapeHtml` (file is standalone) |

### Already-safe (verified, not changed)
`map.js` popups (`safeEsc`), live-feed text (`escapeHtml(preview)`), packet-detail text, channel messages (`channels.js`), `route-render.js` popups, `hop-display.js`.

### Why escape at the sink (not the backend)
`sanitizeName()` only strips control chars; HTML-escaping stored names server-side would be lossy and corrupt legitimate names containing `& < >`, and break the `meshcore://` deep-links / exports. Output-encoding at render is the correct OWASP fix and matches `meshcore-card` v0.3.3.

### Tests
- Added 6 `escapeHtml` regression tests including the CVE payload `` and an attribute-breakout payload.
- `node test-frontend-helpers.js`: **568 passed / 32 failed** — the 32 are pre-existing sandbox limitations (e.g. `AreaFilter is not defined`), identical to the untouched baseline (562/32). Zero new failures.

### Cache busting
Automatic — the server rewrites `__BUST__` in `index.html` with a restart timestamp, so no manual bump is needed.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -127,6 +127,7 @@ jobs:
           node test-issue-1420-tile-providers.js
           node test-issue-1438-marker-css-vars.js
           node test-live.js
+          node test-xss-escape-sinks.js
 
       - name: 🧹 Frontend lint (eslint no-undef) — issue #1342
         run: |
--- a/public/analytics.js
+++ b/public/analytics.js
@@ -443,7 +443,7 @@
       const pct = (t.count / total * 100).toFixed(1);
       const w = Math.max(t.count / total * 100, 1);
       html += `<div class="payload-bar-row">
-        <div class="payload-bar-label"><span class="legend-dot" style="background:${colors[i]}"></span>${t.name}</div>
+        <div class="payload-bar-label"><span class="legend-dot" style="background:${colors[i]}"></span>${escapeHtml(t.name)}</div>
         <div class="hash-bar-track"><div class="hash-bar-fill" style="width:${w}%;background:${colors[i]}"></div></div>
         <div class="payload-bar-value">${t.count} <span class="text-muted">(${pct}%)</span></div>
       </div>`;
@@ -578,7 +578,7 @@
       const barPct = Math.max(((t.avg - (-12)) / 27) * 100, 2);
       const color = t.avg > 6 ? statusGreen() : t.avg > 0 ? statusYellow() : statusRed();
       html += `<tr>
-        <td><strong>${t.name}</strong></td>
+        <td><strong>${escapeHtml(t.name)}</strong></td>
         <td>${t.count}</td>
         <td><strong>${sf(t.avg, 1)} dB</strong></td>
         <td>${sf(t.min, 1)}</td>
@@ -2562,7 +2562,7 @@ function destroy() { _stopRolesRefresh(); _stopScopesRefresh(); _analyticsData =
             tip.style.display = 'block';
             tip.style.left = (cx + 12) + 'px';
             tip.style.top = (cy - 8) + 'px';
-            tip.innerHTML = `<strong>${esc(n.name || n.pubkey.slice(0, 12) + '…')}</strong><br>Role: ${esc(n.role || 'unknown')}<br>Neighbors: ${n.neighbor_count || 0}`;
+            tip.innerHTML = `<strong>${escapeHtml(n.name || n.pubkey.slice(0, 12) + '…')}</strong><br>Role: ${escapeHtml(n.role || 'unknown')}<br>Neighbors: ${n.neighbor_count || 0}`;
           } else if (tip) {
             tip.style.display = 'none';
           }
--- a/public/analytics.js
+++ b/public/analytics.js
@@ -2006,7 +2006,7 @@
           color: isEnd ? (i === 0 ? statusGreen() : statusRed()) : statusYellow(),
           fillColor: isEnd ? (i === 0 ? statusGreen() : statusRed()) : statusYellow(),
           fillOpacity: 0.9, weight: 2
-        }).bindTooltip(n.name, { permanent: false }).addTo(map);
+        }).bindTooltip(esc(n.name), { permanent: false }).addTo(map);
       });
 
       L.polyline(latlngs, { color: statusYellow(), weight: 3, dashArray: '8,6', opacity: 0.8 }).addTo(map);
@@ -3359,7 +3359,7 @@ function destroy() { _stopRolesRefresh(); _stopScopesRefresh(); _analyticsData =
           else if (obs.current_noise_floor >= -100) nfClass = 'rf-nf-warning';
         }
 
-        return `<div class="rf-cell${isSelected ? ' rf-cell-selected' : ''}" data-observer="${obs.observer_id}" tabindex="0" role="button" aria-label="Observer ${name}, noise floor ${nf} dBm">
+        return `<div class="rf-cell${isSelected ? ' rf-cell-selected' : ''}" data-observer="${obs.observer_id}" tabindex="0" role="button" aria-label="Observer ${esc(name)}, noise floor ${nf} dBm">
           <div class="rf-cell-header">
             <span class="rf-cell-name">${esc(name)}</span>
             <span class="rf-cell-nf ${nfClass}">${nf} dBm</span>
--- a/public/app.js
+++ b/public/app.js
@@ -791,10 +791,15 @@ window.pullReconnect = pullReconnect;
 window.setupPullToReconnect = setupPullToReconnect;
 window.connectWS = connectWS;
 
-/* Global escapeHtml — used by multiple pages */
+/* Global escapeHtml — used by multiple pages.
+   5-char OWASP set: escapes ' too so this helper is safe in both
+   double-quoted AND single-quoted attribute contexts (e.g. the
+   data-conflict='${escapeHtml(JSON.stringify(...))}' attr in
+   hop-display.js, where JSON containing a single quote would
+   otherwise break out of the attribute). Fixes #1536. */
 function escapeHtml(s) {
   if (s == null) return '';
-  return String(s).replace(/&/g,'&amp;').replace(/</g,'&lt;').replace(/>/g,'&gt;').replace(/"/g,'&quot;');
+  return String(s).replace(/&/g,'&amp;').replace(/</g,'&lt;').replace(/>/g,'&gt;').replace(/"/g,'&quot;').replace(/'/g,'&#39;');
 }
 
 /* Global debounce */
@@ -1500,14 +1505,14 @@ window.addEventListener('DOMContentLoaded', () => {
         for (const n of nodeList.slice(0, 5)) {
           if (n.name && n.name.toLowerCase().includes(q.toLowerCase())) {
             html += `<div class="search-result-item" tabindex="0" role="option" data-href="#/nodes/${n.public_key}">
-              <span class="search-result-type">Node</span>${n.name} — ${truncate(n.public_key || '', 16)}</div>`;
+              <span class="search-result-type">Node</span>${escapeHtml(n.name)} — ${truncate(n.public_key || '', 16)}</div>`;
           }
         }
         const chList = Array.isArray(channels) ? channels : [];
         for (const c of chList) {
           if (c.name && c.name.toLowerCase().includes(q.toLowerCase())) {
             html += `<div class="search-result-item" tabindex="0" role="option" data-href="#/channels/${c.channel_hash}">
-              <span class="search-result-type">Channel</span>${c.name}</div>`;
+              <span class="search-result-type">Channel</span>${escapeHtml(c.name)}</div>`;
           }
         }
         if (!html) html = '<div class="search-no-results">No results found</div>';
--- a/public/area-map.html
+++ b/public/area-map.html
@@ -222,6 +222,13 @@ function buildSidebar() {
   });
 }
 
+// Escape untrusted strings (e.g. mesh-advertised node names) before HTML interpolation. (#1536)
+// 5-char OWASP set — including ' for single-quoted attribute safety.
+function escapeHtml(s) {
+  if (s == null) return '';
+  return String(s).replace(/&/g, '&amp;').replace(/</g, '&lt;').replace(/>/g, '&gt;').replace(/"/g, '&quot;').replace(/'/g, '&#39;');
+}
+
 async function loadAreaNodes(areaKey, color, group) {
   try {
     const resp = await fetch(`${baseUrl}/api/nodes?area=${encodeURIComponent(areaKey)}&limit=9999`);
@@ -230,8 +237,8 @@ async function loadAreaNodes(areaKey, color, group) {
     (data.nodes || data || []).forEach(n => {
       if (!n.lat || !n.lon) return;
       const m = L.circleMarker([n.lat, n.lon], { radius: 7, color, fillColor: color, fillOpacity: 0.85, weight: 2 });
-      m.bindPopup(`<div class="node-popup"><strong>${n.name || n.public_key?.slice(0,8) || '?'}</strong><br>
-        ${n.public_key?.slice(0,16)}…<br>GPS: ${n.lat.toFixed(5)}, ${n.lon.toFixed(5)}<br><em>${areaKey}</em></div>`);
+      m.bindPopup(`<div class="node-popup"><strong>${escapeHtml(n.name || n.public_key?.slice(0,8) || '?')}</strong><br>
+        ${escapeHtml(n.public_key?.slice(0,16) || '')}…<br>GPS: ${n.lat.toFixed(5)}, ${n.lon.toFixed(5)}<br><em>${escapeHtml(areaKey)}</em></div>`);
       group.addLayer(m);
     });
   } catch(_) {}
@@ -244,8 +251,8 @@ function buildNodeLayer() {
     if (!n.lat || !n.lon || (Math.abs(n.lat) < 0.001 && Math.abs(n.lon) < 0.001)) return;
     count++;
     const m = L.circleMarker([n.lat, n.lon], { radius: 4, color: '#999', fillColor: '#bbb', fillOpacity: 0.6, weight: 1 });
-    m.bindPopup(`<div class="node-popup"><strong>${n.name || n.public_key?.slice(0,8) || '?'}</strong><br>
-      ${n.public_key?.slice(0,16)}…<br>GPS: ${n.lat.toFixed(5)}, ${n.lon.toFixed(5)}</div>`);
+    m.bindPopup(`<div class="node-popup"><strong>${escapeHtml(n.name || n.public_key?.slice(0,8) || '?')}</strong><br>
+      ${escapeHtml(n.public_key?.slice(0,16) || '')}…<br>GPS: ${n.lat.toFixed(5)}, ${n.lon.toFixed(5)}</div>`);
     allNodeLayer.addLayer(m);
   });
   document.getElementById('node-count').textContent = count;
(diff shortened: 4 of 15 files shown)


efiten:
Follow-up commit a0279dd adds the **`map.js` popups**, which I incorrectly marked as already-safe in the original audit. Its `safeEsc` helper was a no-op: it probed `globalThis.esc` (never assigned — the real global is `escapeHtml`) and fell back to `function (s) { return s; }`, so node names, observer names, IATA, hash-prefix and channel names rendered raw on the main map. Caught via the disclosure post (https://mxsasha.eu/posts/meshcore-xss-home-assistant/). Fixed by pointing `safeEsc` at `escapeHtml` with a real (non-identity) fallback, which covers all six call sites. The PR now genuinely covers every identified sink, map.js included.

Kpa-clawbot (changes requested):

**Verdict: NEEDS-WORK — request changes**

---

### TDD red→green history check

`git log origin/master..pr-1537-head --reverse --name-only` shows three commits:

| # | SHA | Files |
|---|-----|-------|
| 1 | `0378ec7b` | 8 production files **+** `test-frontend-helpers.js` (same commit) |
| 2 | `a0279ddc` | `public/map.js` only (no test) |
| 3 | `2fa8e8e0` | `public/analytics.js`, `public/packets.js` (no test) |

**No red commit exists.** Tests were added in the same commit as the production fix. Per AGENTS.md: *"Tests added in same commit as fix (no red→green visible)"* blocks merge. This is a security bug fix on **existing** UI surfaces (search dropdown, nodes table, observers table, packets table, map popups) — explicitly NOT eligible for the net-new-UI exemption. Red→green is REQUIRED.

CI run for commit 1 alone: not queryable (no run keyed to that SHA — only HEAD has a pending run). That's a consequence of the bundled commit, not a workaround for it.

---

### Six Questions

**(a) Show the test that fails when this change is reverted.** ❌ **FAILS — tautology.**
The 6 added tests exercise the global `escapeHtml(s)` function directly. But `escapeHtml` is **unchanged from master** (`public/app.js:795` identical pre/post). Revert the entire production diff (every `${escapeHtml(n.name)}` insertion in app.js / nodes.js / observers.js / packets.js / live.js / analytics.js / route-view.js, and the `safeEsc` rewrite in `map.js`) and **all 6 tests still pass.** They don't gate the change. They are advocacy, not validation.

**(b) Smallest test that would have caught the CVE.** ❌ Not present.
Required: a JSDOM/sandbox render of each fixed sink with `name = ''`, then assert the rendered fragment contains no `` / no unescaped `"`. The PR has zero such tests for any of the 9 sinks. This is exactly the "E2E-DOM-grep tests" pattern AGENTS.md cites as valid for security bug fixes on existing UI.

**(c) Could a wrong implementation pass these tests?** ❌ Trivially yes.
The wrong implementation `// don't call escapeHtml anywhere` passes 6/6 tests. So would `escape only <>` for the `nodes.js`/`observers.js`/`packets.js` table cells (the attribute-breakout assertion only fires on direct `escapeHtml` invocation). The tests prove `escapeHtml` works in isolation — they prove nothing about whether it is wired up at the sinks. This is the single largest issue.

**(d) Edge cases NOT tested.**
- `'` single-quote — `escapeHtml` does NOT escape `'`. `area-map.html`'s local copy doesn't either. None of the current attribute renders use single quotes (`hop-display.js:85` uses `'${conflictData}'` for non-user data), so not exploitable today, but a regression test for `'` would lock the contract.
- Single-quoted attribute context — no test verifies escapeHtml is sufficient for `attr='${val}'` (it isn't).
- `bindPopup` / `bindTooltip` Leaflet path — these accept HTML; PR fixes 8 such sites and has no rendering test for any of them. The `map.js safeEsc` was a 100% identity-function no-op silently shipping for some time — the second commit's fix is invisible to the test suite.
- Double-escape / `&amp;amp;` round-trip — not tested.
- Mutation-XSS / `<noscript>` parsing — out of scope for a same-class regression, fine to skip.

**(e) Test names describe behavior.** ✅ Good. `"img onerror payload is neutralised"`, `"attribute-breakout payload escapes quotes and brackets"` — these read as behavior.

**(f) Test setup vs. test.** ✅ Acceptable. `makeSandbox()` + `loadInCtx` is reused infrastructure, not setup smell.

---

### Must-fix (BLOCKER)

1. **Anti-tautology failure — the tests do not gate the production change.** Add at least one DOM-grep test per sink category that:
   - Stubs the data fetch with a node/observer whose `name = ''`
   - Invokes the render path (or a minimal DOM-mounting helper)
   - Asserts the resulting HTML string contains neither `') !== '<x>'`. Add it. One assertion.

4. **`'` single-quote not escaped — add a contract test.** Even if no current sink uses single-quoted attributes, codifying that `escapeHtml` does NOT escape `'` (and therefore single-quoted attribute interpolation is unsafe) prevents the next contributor from `attr='${escapeHtml(x)}'` and shipping XSS round 3.

---

> *"I'm not interested in whether your tests pass. I'm interested in whether they would fail if you introduced a bug."* — These don't. Revert the production diff, watch the suite stay green, then we'll talk.

Kpa-clawbot (changes requested):
## djb Review (round 1)

> *Channeling Dan Bernstein — every input hostile, fail closed, defense in depth.*

The PR addresses the obvious DOM-text sinks correctly, and the audit table in the description is thorough. But the helper itself is **wrong**, and that wrongness propagates everywhere the helper is used in attribute context. There is no "mostly correct" XSS escape function.

### Must-fix

1. **`escapeHtml` in `public/app.js:795` escapes only 4 characters (`& < > "`) — missing `'`.** This is not academic. The codebase already contains single-quoted attribute interpolations that feed escaped data, e.g. `hop-display.js:85`:
   ```js
   const conflictData = escapeHtml(JSON.stringify({ h, conflicts, globalFallback }));
   `<button … data-conflict='${conflictData}' onclick="…">`
   ```
   `conflicts[].name` is a mesh-advertised node name. `JSON.stringify` does NOT escape `'` (JSON has no such requirement), and the current `escapeHtml` does not escape `'` either. A node named `x' onmouseover='alert(1)` (or anything containing a stray `'`) breaks out of the single-quoted attribute. **Stored XSS, same class as #1536, not fixed by this PR.** Fix: append `.replace(/'/g,'&#39;')`. The existing `esc()` in `analytics.js:7` is the correct 5-char reference — match it.

2. **`public/area-map.html:227` local `escapeHtml` has the same 4-char defect.** Standalone page, but same node-name data, same attribute-context risk if anyone ever interpolates it in single-quoted attrs. Make it 5-char from day one — don't ship a known-incomplete escaper.

3. **`public/map.js:30` fallback `safeEsc` has the same 4-char defect** in its inline fallback. Same fix.

4. **Test corpus is not adversarial.** The payloads cover double-quote breakout (`"><script>`) and angle-bracket text injection — both of which the current helper already handles. None of the new tests would have caught the single-quote / single-quoted-attribute breakout that this PR ships with. Add at least:
   - `'><script>alert(1)</script>` → must escape `'`
   - Polyglot like `</style><svg/onload=alert(1)>` and verify all five critical chars escaped
   - Round-trip: `escapeHtml(JSON.stringify({name: "x' z"}))` and assert no raw `'` survives — this is the exact `hop-display.js` shape.

### Out-of-scope (recommend, separate PRs)

- **No `Content-Security-Policy` header** anywhere in repo (`gh search code` returns zero hits). A strict CSP (`default-src 'self'; script-src 'self'; object-src 'none'; base-uri 'self'`) would have made #1536 unexploitable as a class. Cannot land in this PR because the codebase is riddled with inline `onclick="location.hash=..."` (observers.js:195, analytics.js:705, packets.js, etc.) that depend on `unsafe-inline`. **Refactor inline handlers → CSP** is the proper follow-up.
- **Server-side belt-and-suspenders sanitization.** `sanitizeName()` (`cmd/server/decoder.go`) only strips control chars — the PR body correctly justifies why it shouldn't lossily HTML-escape stored names. Fine. But consider rejecting `adv_name` containing raw `<` / `>` / `\x00`-`\x1f` at ingest as a separate hardening — a name with literal `<script>` has no legitimate use and ingesting it is fail-open.
- **No CSV / JSON export channels reviewed** — `gh search` for "csv" / "exportCsv" returns nothing in `public/`, so no second-order escape regression to chase here. Noted as "checked, nothing relevant."
- **No `javascript:` URI sinks** found via `href="${…name…}"` patterns — names don't reach `href` attributes in this code.
- **Crypto / key / secret paths untouched** — verified, no regressions.

### Verdict

NEEDS-WORK. The 4-char `escapeHtml` is the same class of bug as #1536: a sink-level escape that doesn't actually cover the sink's context. Fix the helper, add a regression test that would have caught it, and we're good.

— *djb persona*

Kpa-clawbot (approved):
## meshcore Review (round 1)

Protocol-aware review of the XSS fix. Verdict: **APPROVE** — no protocol-level XSS bypass left in the surfaces the patch touches. Notes below are out-of-scope follow-ups, not blockers.

### Completeness check — node-controlled strings vs firmware

Cross-referenced `firmware/src/helpers/AdvertDataHelpers.{h,cpp}` and `firmware/docs/payloads.md`. The node-controlled strings that surface in the UI are:

| Field | Source | Escaped on this branch? |
|---|---|---|
| `adv_name` (ADVERT, after `ADV_NAME_MASK`) | mesh, attacker-controlled | ✅ yes (Go `sanitizeName` + JS `escapeHtml`) |
| Channel `decoded.sender` (decrypted CHAN body) | attacker-controlled | ✅ yes (`channels.js:2112`, `packets.js:3246`) |
| Channel `decoded.channel` (display name) | local config lookup, not on-wire | ✅ yes |
| Channel `decoded.text` body | attacker-controlled | ✅ yes (`highlightMentions` escapes first, then re-injects regex captures from already-escaped text — safe) |
| `replyTarget` (`@TARGET` prefix in CHAN body) | attacker-controlled | ✅ yes (escaped at `channels.js:2090`) |
| Path hop names (resolved from advert) | mesh, attacker-controlled | ✅ yes (`hop-display.js:74,38,83` + new `route-view.js:820,1007` + `live.js:2351`) |
| Observer name | operator-trusted (MQTT config) | ✅ escaped defensively — threat-model-correct |
| Sender pubkey prefix on TXT/PATH/REQ | mesh, but it's hex from `srcHash.slice(0,8)` | n/a (hex, no escape needed) |

No ROOM topic or PATH "hop comment" exists on the wire — those don't show up in firmware payload defs. Nothing missed.

### ADVERT parsing — bypass attempts

Verified against `AdvertDataHelpers.cpp:51-57`: the name is `app_data_len - i` bytes of remainder, NUL-terminated by the parser. So firmware-originated names CAN'T contain embedded NUL, but **on-wire** an attacker can craft any byte sequence — the parser will copy it verbatim up to the next 0x00, and CoreScope's MQTT pipeline sees the raw bytes.

`cmd/server/decoder.go:313-324` is solid:
1. `TrimRight(name, "\x00")` — strips trailing NUL (firmware-style).
2. `sanitizeName()` — strips `< 0x20` and `0x7f` ⇒ kills embedded NUL, BEL, CR-only, SO/SI, etc. that could break a guard.
3. `len(name) > 32` truncate — matches `MAX_ADVERT_DATA_SIZE` from `firmware/src/MeshCore.h`.

No NUL/control-byte bypass of `escapeHtml`. Good.

### Out-of-scope (separate issues, NOT must-fix for this PR)

1. **`sanitizeName` does NOT strip Unicode bidi / format characters** (U+202E RLO, U+200E LRM, U+2066-2069 isolates, U+200B-200D ZWSP/ZWJ, U+2028/9 line/para separators). All `>= 0x20`, all pass through `escapeHtml` untouched. An attacker advertising `evil\u202Eevon-RP01` will spoof a different name in the UI. Different bug class (UI-spoof, not XSS) — file a follow-up to extend `sanitizeName` with a Unicode-category strip (Cf, plus the bidi-override subset).
2. **`name[:32]` truncates by BYTES, not runes** (`cmd/server/decoder.go:323`). A 4-byte UTF-8 character straddling byte 32 leaves an invalid trailing byte. Cosmetic only — frontend renders U+FFFD — but worth fixing with `utf8.ValidString` aware truncation.
3. **No call-site regression test**. The new tests cover `escapeHtml` unit behavior but don't assert each render site invokes it. The diff's manual audit is the only thing standing between this fix and future re-introduction of `${n.name}` in some new file. Consider an ESLint custom rule `no-unescaped-template-name` or a grep-based CI gate.
4. **`safeEsc` fallback in `map.js`** correctly avoids identity-passthrough now — solid defensive change worth highlighting in the commit message.

### Channels.js claim verification

PR rationale states channels.js was already safe. Confirmed:
- `channels.js:331-334` defines a local `escapeHtml` (different from `app.js` global but equivalent).
- Sender name at line 2112: `${escapeHtml(sender)}` — safe.
- Body text via `highlightMentions(rawBody)` (`channels.js:351-356`) — escapes first, then runs a regex against escaped output; captured group can no longer contain raw `<`/`>`/`&`/`"`.
- Reply prefix `replyTarget` at line 2090 — captured from `/^@([A-Za-z0-9_\-]{1,32})\s+/` (charset-restricted) then `escapeHtml`-wrapped anyway. Defensive belt-and-braces.

Channels surface is clean.

### Verdict

**APPROVE**. Ship it. File follow-up issues for the bidi-spoof and UTF-8-truncate items.

Kpa-clawbot (changes requested):

Reviewed cold against HEAD `2fa8e8e0`. The fix direction (output-encode at the sink) is right and the test additions are assertion-shaped, not tautological. However, the PR description claims "I audited every name/sender/text/channel render in `public/` and this PR escapes all unescaped sinks. There are no known remaining XSS sinks of this class after this change." — that claim does not hold. Found one missed sink that is in scope (same `adv_name` → `innerHTML` class as the rest of the diff), plus a defense-in-depth issue in the new helper.

### Must-fix

1. **`public/live.js:553-565` — `buildClickablePathPopupHtml` injects `hopNames.join(' → ')` raw into popup HTML.** `hopNames` is populated at line 3199 from `hp.name || (hp.key ? hp.key.slice(0, 8) : '?')`, where `hp` is a hop position whose `name` originates from mesh-advertised `adv_name`. The popup is later opened via `L.popup().setContent(html)` (line 593) — same `innerHTML`-equivalent stored-XSS sink class this PR is fixing elsewhere. Escape each hop name before `.join(...)` (e.g. `hopNames.map(escapeHtml).join(' → ')`). This is the exact pattern fixed in `route-view.js` and `analytics.js` topology in this same diff; it was missed in `live.js` clickable-path animation.

2. **`public/area-map.html:239-240 and 253-254` — popup HTML template only escapes `n.name`; surrounding interpolations `${n.public_key?.slice(0,16)}` and `${areaKey}` remain raw.** Pubkey is hex from DB and `areaKey` is server-supplied so these are *currently* low-risk, but the PR explicitly asserts "no known remaining XSS sinks of this class" and the surrounding template is part of the same audit surface. If `area` labels ever flow through user-defined values (or if pubkey resolution ever returns non-hex), the payload survives. Escape all interpolations in this template for consistency with the rest of the diff (the new local `escapeHtml` is already in this file).

3. **New local `escapeHtml` in `public/area-map.html:227` (and by extension the global in `app.js:795`) does not escape `'` (single quote).** The PR adds a brand-new escaper that perpetuates the existing 4-char gap (`& < > "`) rather than the 5-char OWASP set. Tests assert "attribute-breakout payload escapes quotes and brackets" but only exercise `"`. The codebase already has a 5-char version in `analytics.js:7` (`esc`) — use that shape. Today no fixed sink uses single-quoted attributes so it is not actively exploitable, but the new helper should not ship with a known incomplete escape set when the project already has a correct one.

### Out-of-scope (file separate issues if you care)

1. The two helpers `escapeHtml` (app.js global, area-map.html local), `esc` (analytics.js IIFE local), `safeEsc` (map.js), and `escapeHtmlLocal` (live.js fetchSuggestions IIFE) all do approximately the same thing with subtly different rules. Consolidate into one canonical exported helper.
2. No CSP `script-src` defense-in-depth header on the dashboard — same class of bug would have been bounded by a strict CSP.
3. The 6 new helper tests cover only the helper itself; no integration test asserts that a given sink (e.g. nodes-table row, live-feed marker tooltip) actually escapes its input. A grep-based "no raw `${n.name}` in `innerHTML` templates" test would catch future regressions; pre-existing gap, not blocking this PR.

### Verdict
NEEDS-WORK — request changes.

Issue #1532, opened 2026-06-02
The Mesh Live and Map Filters window has a large area of wasted space and obstructs the fireworks (packet) view


EldoonNemar:
Per discussions with @KpaBap on discord, we should give users the ability to minimize the  amount of information is displayed on the screen so they can enjoy the fireworks (packets) display.  To do this we should:

- Clean up the mesh live / filters window so it consumes less space
- Tie as much as we can to the pin icon so that it provides an unobstructed view of the map
- Potentially add a "Fullscreen" button that removes ALL UI overlays, except perhaps the mesh live stats in a small corner window.

Examples:

<img width="2096" height="131" alt="Image" src="https://github.com/user-attachments/assets/35968d10-eb8c-4fa7-9017-b59744badfb0" />

<img width="1283" height="170" alt="Image" src="https://github.com/user-attachments/assets/beb5fe6d-e9cd-42d6-8143-e2fab003514c" />

Kpa-clawbot:
**Severity:** P2 — chrome obscures the primary visualization (the map).

**Claim:** confirmed (CDP screenshot, staging /#/live at 780×500). Header row occupies ~110px before the map starts; left "Waiting for packets…" panel + bottom VCR bar take another ~200px. On the 2096×131 example, the row's used-ink is title + 3 stat pills + ~10 toggles = the rest is padding/whitespace.

**Code today:**
- `public/live.js:1044-1095` — `.live-stats-row` (3 pills) + `.live-controls` (8 toggles + node filter + region + area + audio sliders) live in a single fixed `#liveHeader` overlay.
- `public/map.js:180-181` — separate `#mapControlsToggle` ⚙️ pin pattern already exists for the map page; same affordance is missing on `/live`.

**Findings:**
- (tufte): data-ink ratio is poor — 11 controls + 3 KPIs displayed permanently steal pixels from THE data (the firework animation). Defaults-on chrome should collapse behind a pin/cog; only the 3 stat pills earn permanent residence (sparkline-grade density).
- (tufte): "Fullscreen" is the right primitive — Tufte's "shrink principle" says strip until unreadable, then add back. A keyboard-toggle (`F`) that hides header+VCR+bottom-nav and floats the 3 pills in a corner satisfies the request without inventing new UI.

**Fix path:**
1. `public/live.js:1044` — wrap `.live-controls` behind the existing `#liveControlsToggle` (collapsed by default on desktop too, not just mobile).
[...]

PR #1580, opened 2026-06-04
fix(UI): Additional fixes for issue #1532

EldoonNemar:
- Eliminated extra space to the right of the map filters.
- Made the map filters and mesh live a single line with a divider
- Resized the input and dropdowns in the map filters so they meet WCAG 2.5.5 by being at least 44px high, but appearing 30px high
- Turned the filters cog and the fullscreen button into native leaflet icons that are large enough to meet WCAV 2.5.5 compliance
- Increased the size of the zoom buttons to meet WCAG 2.5.5 compliance on both the live and map pages
- If the top nav bar is pinned, it won't disappear during fullscreen but if it isn't pinned, it will disappear with everything else.
- The cog and full screen button change color to show they're active

Final Outcome in 4k
image
 
Final Outcome in 1080p
image

### DIFF
--- a/public/live.css
+++ b/public/live.css
@@ -508,22 +508,8 @@ body.live-fullscreen .live-legend,
 body.live-fullscreen .legend-toggle-btn,
 body.live-fullscreen .feed-show-btn { display: none !important; }
 
-/* Collapse the header chrome to just the floating stats row pinned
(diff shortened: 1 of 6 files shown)


EldoonNemar:
@Kpa-clawbot can you look at this please. The current build is all wonky with your fullscreen commit

Issue #1574, opened 2026-06-04
FR: Now that the map has been optimized, we should allow users to increase the maximum nodes on screen


EldoonNemar:
The current on map node count is statically 2000, in live.js. Given the performance improvements we've made to the map, I've tested it with 4300 nodes and saw no noticeable performance hit.

That said, we should make a config item that lets users select the max nodes to display on the live map at one time, giving them the power to decide what performance they want to give their users.

Kpa-clawbot:
**Severity:** P3 — perf-headroom unlocked, no breakage today.

**Claim:** partial. The 2000 cap is real (`public/live.js:2515-2516`), but it's an `/api/nodes?limit=2000` query cap, not a render cap — there is no separate `MAX_NODES` constant guarding map markers. So "raise the cap" effectively means "raise the page-size knob on this one fetch."

**Code today:**
- `public/live.js:2515-2516` — `/api/nodes?limit=2000` (with and without `before=`), used by the live-map node-load path.
- `public/live.js:480` — `/api/packets?limit=2000` (unrelated; not what reporter is asking about, but worth flagging that the same magic number lives in two places).
- No `config.example.json` or instance-config field for a node-display cap; nothing today reads operator config for this value.

**Findings:**
- (tufte): a single knob ("max nodes on map") is the right read — operator picks the data-density vs. perf tradeoff per deployment. Default should remain conservative; only operators with measured headroom raise it. Reporter's 4300-node test is one data point, not a universal headroom claim.
- (carmack): "no noticeable performance hit at 4300" needs a measurement, not a vibe. Before raising the default, want a Lighthouse / frame-time number on a mid-range mobile + the heap delta at N=4300 vs N=2000. The cap also gates a `/api/nodes` SQL query — at N=10k the network/serialization cost may dominate the render cost the reporter measured.

**Fix path:** add `liveMap.maxNodes` to instance config (default 2000) + `_comment_` in `config.example.json` per AGENTS.md config rule; plumb through `/api/instance-config`; replace literals at `public/live.js:2515-2516` (and ideally factor the `/api/packets?limit=2000` literal at :480 into a sibling const at the same time). Cap-clamp server-side too so a misconfig can't request 1M. Effort: S.

**Dup check:** searched "max nodes", "live map cap", "node limit" — none open. #1011 / #1218 touch nodes-page perf but not this knob.

**Triage verdict:** ready-to-fix

Kpa-clawbot:
[...]

PR #1585, opened 2026-06-04
fix: bypass API limit clamps for internal UI requests. Revisit of issue #1540

EldoonNemar:
Follow-up to #1540 (audit-input-vulns-20260603).

The recent security patch introduced a strict maximum of 500 items on list endpoints (`/api/nodes`, `/api/packets`, etc.) to prevent Denial of Service via unbound `limit` parameters. 

While this correctly protects the server from malicious or misconfigured external scripts, it inadvertently broke the SPA frontend. The UI legitimately requests up to 2,000 nodes for the map and 10,000 packets for the live feed. The backend was silently truncating these requests, resulting in an incomplete map and live view.

This PR introduces an `isInternalUIRequest` heuristic that checks for standard browser security headers. This safely bypasses the public caps for internal UI calls without requiring changes across all 40+ frontend `fetch()` usages.

## What Changed
- **`cmd/server/clamp_limit.go`**: Added `isInternalUIRequest` helper checking `Sec-Fetch-Site: same-origin` with a `Referer` host fallback.
- **`cmd/server/clamp_limit.go`**: Updated `queryLimit` wrapper to bypass the strict `max` parameter (allowing up to 50,000) if the request originated internally.
- **`cmd/server/clamp_limit_test.go`**: Added `TestQueryLimit` coverage to assert external requests remain strictly clamped while internal ones bypass the restriction.

## Verification
- External raw API requests (`curl`, scripts) without browser headers are still safely clamped to 500.
- UI Map correctly fetches all mesh nodes up to its requested 2000 limit.
- UI Live feed correctly fetches up to 10000 packets without truncation.


EldoonNemar:
@Kpa-clawbot I need you to look at this. This should be a priority 1 as it wrecks anyones map who has more than 500 nodes, because of your DDOS protection

Kpa-clawbot (changes requested):
# djb Review (round 1) — security blocker

**Verdict: REQUEST CHANGES**

You found a real bug — #1540's 500-cap silently truncates legitimate UI fetches (map ~2000 nodes, live feed ~10000 packets). That problem is real and worth fixing. But the mechanism in this PR is the wrong shape: it regresses the security posture #1540 restored, while only *appearing* to preserve it.

---

## MUST-FIX

### 1. `Sec-Fetch-Site` is not a security boundary — the cap-bypass is trivially forgeable
**File:** `cmd/server/clamp_limit.go:31-46` (`isInternalUIRequest`), used at `routes.go:935, 961, 1255` (`queryLimit(r, 50, 500, 10000)`).

`Sec-Fetch-Site` is a *browser-provided* header. The server has no way to verify the request actually came from a browser, let alone a same-origin one. Any non-browser client sets it freely:

```
curl -H 'Sec-Fetch-Site: same-origin' \
     -H 'Origin: https://analyzer.00id.net' \
     'https://analyzer.00id.net/api/packets?limit=10000'
```

That request bypasses the 500-cap and returns up to 10,000 packets. The `Origin` defense-in-depth check at lines 38-44 doesn't help — `Origin` is *also* a client-controlled header. The attacker is not a browser; they are not subject to browser policy.

This means:

- **The PR body's verification claim is false.** "External raw API requests (`curl`, scripts) without browser headers are still safely clamped to 500" is incorrect. A `curl` with a single header bypasses the clamp.
- **#1540's DoS protection is negated for any attacker who knows the deployment hostname** — which is public information (`analyzer.00id.net`, `analyzer.cascadiamesh.org`, etc.). The clamp now only protects against attackers who don't know to spoof one header. That's not a security boundary; it's an obfuscation layer.
- **The PR body also claims a `Referer` host fallback**, but the code in `clamp_limit.go` doesn't implement one. Either way: `Referer` is just as client-controlled. Adding it would not help.

This is security-by-browser-convention. The server MUST NOT trust headers the client sets to make authorization decisions about resource consumption. The check looks defensive but does nothing against an adversarial caller, which is the only caller that matters for a DoS cap.

### 2. The `internalMax = 10000` ceiling is honest DoS surface, not "internal-only"
**File:** `cmd/server/routes.go:935, 961, 1255`.

Because finding #1 means any caller can opt into the higher cap, `internalMax: 10000` is effectively the new *public* cap on `/api/packets`. At ~1KB/packet that's a 10MB response per request, unbounded by IP. For `/api/nodes` the per-row payload is smaller but still meaningful. The cap as written is "expensive but bounded" only if you accept that *every* caller can reach it.

(Parent's brief flagged `50000`; the diff actually lands at `10000`. Still too high if treated as the public DoS bound — but at least an order of magnitude better than 50000.)

### 3. Recommended alternative — drop the heuristic, raise the caps, make them config knobs
The right shape is:

1. **Raise the hardcoded caps to match real UI need**, and make them configurable per-deployment (config.json, with sensible defaults). E.g.:
   ```
   listLimits.packetsMax: 10000
   listLimits.nodesMax: 2000
   listLimits.analyticsMax: 200   // unchanged
   listLimits.channelMessagesMax: 500
   ```
2. **Delete `isInternalUIRequest` entirely.** No `Sec-Fetch-Site` reliance, no `Origin` check, no `Referer` parsing. The server doesn't try to guess who's calling.
3. **`queryLimit(r, def, max int)` stays its original two-cap shape.** The `max` is the honest, public, DoS-bounded ceiling.
4. **Operators tune for their mesh:** small deployment → tighten the knobs; big deployment → raise them. This is the same axis #1574/#1577 already opened with `LIVE_MAP_MAX_NODES` — extend the pattern, don't sidestep it.

Result: the UI works (caps match real usage), DoS protection holds (capped at a known, bounded value, equal for all callers), operators have a real lever, and there is no security-by-browser-convention surface to misread later.

---

## OUT-OF-SCOPE

### 4. Diff scope creep — frontend files unrelated to the limit-clamp fix
The PR touches:

- `public/live.css` (+78/-37)
- `public/live.js` (+87/-42)
- `public/style.css` (+4)
- `test-issue-1532-live-fullscreen.js` (+8/-25)
- `test-live.js` (+9)
- `test-e2e-playwright.js` (+84) — new file

None of these are needed for the server-side limit-clamp change. They appear to be either (a) merge-conflict resolutions against the just-merged #1572 (`fix/1532-live-fullscreen`), or (b) unrelated drive-bys. Reviewing the live fullscreen rework requires the #1532 reviewers, not a security pass. **Please split:** server-side cap change in one PR, frontend live-page work in another. Mixing them blocks clean revert and confuses the audit trail.

---

## Recommendation

The bug is real and worth fixing — thank you for catching that #1540 truncates legitimate UI fetches. But the fix as written **regresses** #1540's posture rather than complementing it. Please either:

- **Close this PR and reopen** with the config-knob approach (drop `isInternalUIRequest`, raise caps as defaults, expose `listLimits.*` in config), with the frontend live-page changes split into a separate PR; or
- **Rework in place:** delete the heuristic, switch to config-driven caps, drop the unrelated `public/live.*` and `test-*` changes.

If neither is workable on your end, the operator can land a replacement PR taking that shape and credit your original report.

The cap is honest DoS-protection at whatever value operators set. It cannot be "honest for browsers, lax for everyone else" — the server has no way to tell them apart.

— djb (security review)

PR #1589, opened 2026-06-05
fix: bypass API limit clamps for internal UI requests. Revisit of issue #1540

EldoonNemar:
This PR replaces the strict, hardcoded limits on API list endpoints (introduced in the recent security patch) with a new operator-configurable `listLimits` block. This change is needed as issue 1540's implementation introduced a 500max node limit on the live map or any other function that leverages the api/nodes backend.

Previously, we attempted to bypass public caps for internal UI requests using a heuristic based on browser headers (`Sec-Fetch-Site`). Following review, we decided to drop that heuristic entirely to eliminate any security-by-browser-convention surface area. 

Instead, `queryLimit()` returns to its original, mathematically simple bounds-checking shape, and the absolute maximums are now drawn from `config.json`. This provides equal DoS protection against all callers while allowing server operators to tune the ceilings based on the size of their mesh (e.g. embedded devices can tighten the knobs, regional hubs can raise them).

### Changes Made:
- **`config.go`**: Introduced a `ListLimits` config struct containing `PacketsMax`, `NodesMax`, `AnalyticsMax`, and `ChannelMessagesMax`. Added safe initialization to ensure default caps (10000, 2000, 200, 500 respectively) apply even if the block is omitted from the config.
- **`clamp_limit.go`**: Deleted `isInternalUIRequest` entirely and restored `queryLimit` to its original signature (`r, def, max`).
- **`routes.go`**: Replaced all hardcoded integer ceilings on list endpoints (`/api/packets`, `/api/nodes`, etc.) with `s.cfg.ListLimits.*`.
- **`config.example.json`**: Added the `listLimits` block with documentation to guide new operators.
- **`clamp_limit_test.go`**: Purged all header-heuristic testing.

### Verification:
- All 611 backend unit tests pass (`npm run test:unit`).
- Bounds-checking math continues to enforce hard DoS clipping exactly at the operator's specified configuration limit.


EldoonNemar:
how we looking now @Kpa-clawbot ?

EldoonNemar:
I should have left it alone. Lets automerge. I've reverted to  [59458c9](https://github.com/Kpa-clawbot/CoreScope/pull/1589/commits/59458c91870386d67c0808bf27c88b18eee8a4ea) and resolved the conflict from my previous PR

EldoonNemar:
pretty please

EldoonNemar:
@Kpa-clawbot you here? I'm ready to have this pushed in. I reverted to when you said it was good to go.

PR #1602, opened 2026-06-05
fix(frontend): reliably restore row focus on panel close

EldoonNemar:
fix for the focus-restore@800 E2E test that's currently failing on master (see runs 26990436988, 26986419081)

Chromium headless is notorious for dropping synchronous or rAF-based focus restores when elements are hidden. By manually blurring the active element before hiding the panel, and staggering the focus restore with a setTimeout macrotask after the rAF, we ensure the focus call lands after the browser has completed all implicit focus resets and event handlers.

Furthermore, dynamically evaluating the focus resolver directly inside the deferred focus attempt prevents the target element from becoming stale if a live WebSocket packet triggers a background table re-render in the intervening milliseconds.

### DIFF
--- a/public/packets.js
+++ b/public/packets.js
@@ -392,6 +392,9 @@
 
   function close() {
     if (!panel || panel.hidden) return;
+    if (document.activeElement && panel.contains(document.activeElement)) {
+      try { document.activeElement.blur(); } catch (_) {}
+    }
     panel.hidden = true;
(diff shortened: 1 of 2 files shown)


EldoonNemar:
This is urgent for the test passing of PR #1589 which is also urgent to fix max node count on the map due to ddos protection added in a previous pr

Issue #4, opened 2026-03-19
home.js stacks duplicate event listeners on re-render


Kpa-clawbot:
## Bug
`setupSearch()` in `home.js` ~L136 adds `document.addEventListener('click', handleOutsideClick)`. It's removed in `destroy()` ~L141, but if `renderHome()` is called multiple times within the same page lifecycle (e.g. toggling experience level), `setupSearch()` runs again without removing the previous listener, stacking duplicates.

## Expected
Only one `handleOutsideClick` listener at a time.

## Fix
Remove the existing listener before adding a new one in `setupSearch()`:
```js
document.removeEventListener('click', handleOutsideClick);
document.addEventListener('click', handleOutsideClick);
```
Or guard with a flag to skip if already set up.

## Severity
Major — causes multiple firings per click, potential performance issues and unexpected behavior on repeated renders.

Issue #1283, opened 2026-05-19
bug(db): vacuumOnStartup fails with SQLITE_BUSY when ingestor + server share DB (auto_vacuum migration #919 broken in single-container topology)


Kpa-clawbot:
## Summary
Setting db.vacuumOnStartup=true in a container that runs BOTH ingestor and server (default supervisord setup) produces:
```
2026/05/19 06:07:53 [db] vacuumOnStartup=true — starting one-time full VACUUM (ensure 2x DB size free disk space)...
2026/05/19 06:07:58 [db] VACUUM failed: database is locked (5) (SQLITE_BUSY)
```
Both processes are started by supervisord. Ingestor opens DB first, takes write lock for ingest. Server starts second, tries VACUUM, fails immediately.

The 5s busy_timeout doesn't help because the ingestor is actively writing, not just holding a lock briefly.

## Repro
1. Set db.vacuumOnStartup=true in config.json
2. `docker restart` a container that runs both processes via supervisord
3. Observe SQLITE_BUSY in server log within 5s, VACUUM never runs

## Impact
Cannot migrate auto_vacuum=NONE -> INCREMENTAL on prod/staging without taking ingestor offline first. The feature works in isolated DBs but not in the typical CoreScope deployment.

## Fix options
1. Server VACUUM should pause/coordinate with ingestor (e.g. ingestor exposes a "I'm idle for N seconds" signal, server VACUUMs during that window)
2. supervisord priority: start server FIRST with VACUUM, only spawn ingestor after server signals VACUUM done
3. Detect topology + log clear error: "vacuumOnStartup requires ingestor to be stopped first; the current container has both processes"
4. Documented procedure: "to migrate, stop ingestor, then run server with vacuumOnStartup=true once, then restart with ingestor enabled"

## Acceptance
- [ ] Either VACUUM coordinates with ingestor, OR error message clearly tells operator the prerequisite
- [ ] Test that gates the SQLITE_BUSY case
- [ ] Documentation in config.example.json about single-container topology limitation

## Out of scope
- Online VACUUM (PRAGMA incremental_vacuum) is a different operation, can run alongside writes
- Replacing the VACUUM with something else

## References
- Original issue: #919 (auto_vacuum migration)
- File: cmd/server/db.go VACUUM path
- Topology: docker/supervisord.conf launches both ingestor + server

Kpa-clawbot:
## Broader architectural finding

[...]

Issue #26, opened 2026-03-19
Nodes table has dead 'Regions' column — always shows —


Kpa-clawbot:
## Issue
The nodes table (`nodes.js` ~L167) includes a "Regions" column that always displays "—". Region data is never populated. This wastes horizontal space, especially on mobile.

Similarly, `packets.js` ~L354-358 has hardcoded `window._regions` as a TODO hack.

## Fix
1. Remove the Regions column from the nodes table
2. Either implement region fetching from the server API or remove the region filter from packets page
3. If regions aren't supported yet, remove all references

## Severity
Minor — wasted space, confusing to users

Issue #25, opened 2026-03-19
Home page content capped at 720px — wastes space on desktop


Kpa-clawbot:
## Issue
Home page content is capped at `max-width: 720px` (home.css). My Nodes grid is 900px. On wide monitors (1440p+), over 50% of the screen is empty.

Stats cards use `flex: 1 1 120px` which would scale up naturally if the container were wider.

## Fix
Increase home page max-width to 1200-1400px. Stats grid and My Nodes grid will fill naturally.

## Severity
Minor

PR #1625, opened 2026-06-08
feat(nodes): per-node Reach page + GET /api/nodes/{pubkey}/reach (directional link quality)

efiten:
## What

Adds a per-node **Reach** view that answers "how well does this specific node hear, and get heard by, its neighbours?" — both as a standalone page (`#/nodes/{pubkey}/reach`) and as a section on the node detail page.

New endpoint: **`GET /api/nodes/{pubkey}/reach`**.

## What it measures

For the target node it derives, from raw `path_json` adjacency (a path travels origin→observer, so in `[A,B]` B received A directly):

- **Directional link counts** per neighbour: `we_hear` (how often we received them) vs `they_hear` (how often they received us).
- **Bidirectional / bottleneck**: a link is two-way stable when both directions > 0; the weaker direction is the bottleneck and rates real two-way reliability.
- **Importance**: neighbour degree + rank, relay-observation volume, bidirectional-link count, direct-observer count.
- **Direct observers**: who received the node at 0 hops, with SNR.

Reliability rule: a neighbour is only attributed when its pubkey **prefix is unique** at the path's byte length (collisions are skipped, never misattributed).

## UI

- Standalone Reach page + node-detail section.
- Reusable bidirectional link map (OSM) with links coloured by bottleneck.
- Incoming/outgoing toggles to isolate each direction.

## Naming note (deliberate, no collision)

This is distinct from the existing **per-observer reachability** in topology analytics (`ReachNode` / `ObserverReach` / `perObserverReach`). This PR adds its own `NodeReach*` response structs in a new `node_reach.go` and a new `/api/nodes/{pubkey}/reach` route — there are no symbol or route collisions (verified: `go build ./...` clean). Happy to rename to disambiguate further (e.g. "Link Quality") if you'd prefer to reserve "Reach" for the per-observer feature.

## Testing

- `cmd/server`: endpoint shape/404/limit-clamp + unit tests for token derivation and directional attribution, plus a scan benchmark — all pass.
- Frontend: helper tests + Reach-page E2E (`test-node-reach-e2e.js`), standalone route + incoming/outgoing toggles.
- `go build ./...` and `eslint public/*.js` (no-undef) clean.

## Docs

Design spec, implementation plan, and the `GET /api/nodes/{pubkey}/reach` API contract are included under `docs/`.


efiten:
📡 **New: per-node "Reach" page**

Every node now has a **Reach** button (next to Analytics). It shows which nodes a node has a **stable two-way RF link** with — nodes that hear each other directly, both ways.

**What you see:**
🗺️ A map of all 2-way links, coloured by strength (the weaker direction counts).
📋 A table: per neighbour how much **we hear them** vs **they hear us**, plus distance. Click a neighbour → go to its page.
📊 Network position: neighbour count + ranking, relay activity and direct observers.

**Extras:**
• **incoming / outgoing** toggles to also show one-way links — handy for spotting asymmetry (a node that hears you loud but that you don't hear back = usually a TX/antenna difference).
• Time window **24h / 7d / 14d / 30d**.
• **Print → PDF** for offline use.

Under the hood a node is only matched via reliable **2–3 byte** hash matches (1-byte collides too much), so the links shown are trustworthy.

Kpa-clawbot (changes requested):
## Carmack Review (round 1)

Hot-path / memory / scan-cost pass on `cmd/server/node_reach.go`. The functionality is fine; the data flow is wasteful in ways that will hurt as the DB grows past dashboard scale. The cache is a band-aid hiding several genuine cost issues underneath.

### Must-fix

1. `cmd/server/node_reach.go:402` — `scanReachRows` returns `var out []pathRow` with **no preallocation and no LIMIT**. A relay node with weeks of observations in a 30-day window loads the entire result set into memory. At 100k matching rows this is unbounded RAM + slice-doubling churn. Add a hard LIMIT (matching dashboard window expectations) and preallocate `make([]pathRow, 0, limit)`.

2. `cmd/server/node_reach.go:419-426` — per-row **double allocation** in the JSON path: `json.Unmarshal` allocates `[]string`, then `make([]string, len(raw))` allocates again to uppercase. 200k allocs at 100k rows, all GC-visible. Either (a) uppercase in place on `raw`, (b) compare lowercase against pre-lowercased `ourTokens`, or (c) skip JSON entirely and parse the bracketed comma-separated hex with a single pass over the byte slice.

3. `cmd/server/node_reach.go:430` — `strings.ToLower(oid)` and `strings.ToLower(fpk)` allocate per row. These come out of SQL — lowercase them once in the query (`LOWER(...)`) or normalize at ingest, not on every hot-path read.

4. `cmd/server/node_reach.go:398` — `o.path_json LIKE '%"TOK"%'` OR'd N times is a **full scan of path_json text for every row in the time window**. `idx_obs_ts` narrows by timestamp only; the LIKE work is O(rows × tokens × avg_path_len). At 100k rows × 3 tokens this is the dominant cost. Either: (a) add a generated/indexed column with the path token set, (b) use json1's `EXISTS(SELECT 1 FROM json_each(path_json) WHERE value IN (...))` with a covering index, or (c) gate by `transmissions.from_pubkey = ?` first when applicable, which collapses the scan to that node's transmissions.

5. `cmd/server/node_reach.go:200-208` — **cache eviction is a full wipe** (`reachCache = map[string]reachCacheEntry{}`). Under sustained load past 256 entries this thrashes every cache reaching the cap, defeating the cache's stated purpose. Use simple LRU or evict-expired-then-evict-oldest.

6. `cmd/server/node_reach.go:182,191` — `sync.Mutex` serializes **all reads** through the cache check too. Use `sync.RWMutex` and `RLock` in `reachCacheGet`. Cheap, real.

7. `cmd/server/node_reach.go:248-265` — **no singleflight / no in-flight dedup**. N concurrent requests for the same `pubkey|days` while the cache is cold all execute the full scan + JSON parse + attribution independently. Use `golang.org/x/sync/singleflight` keyed by `cacheKey`.

8. `cmd/server/node_reach.go:283-292` — the **degree/rank CTE runs the full `neighbor_edges` UNION+GROUP BY on every cache miss**. Unbounded as the graph grows; the work is identical across all reach requests except the WHERE pk match. Cache the `(neigh, rank)` aggregate at a coarser TTL (or have `neighbor_edges` maintain a per-node degree counter) and look up by pubkey.

9. `cmd/server/node_reach.go:270` — `s.buildNodeInfoMap()` builds the **entire pubkey→nodeInfo map** (every node in the DB) per cache-miss request, just to look up self + neighbors. Use the existing cached nodes from `getCachedNodesAndPM()` and a direct lookup, or scope the map build to seen pubkeys.

10. `cmd/server/node_reach.go:113-115` — `uniqueResolve` calls `strings.ToLower(token)` on **every hop, every row**. Tokens are already normalized uppercase from the scan; the prefix map should be indexed with the same case (or store both). Per-hop string allocation × N rows is pure waste.

11. `cmd/server/node_reach.go:46-90` — `attributeDirections` creates `map[string]int` and `map[string]*obsAgg` with **no size hint**. With known row count from the scan you can `make(..., n/8)` or similar; map growth is the second-biggest alloc source after JSON parsing.

12. `cmd/server/node_reach.go:248-302` — **no `context.Context`** plumbed into the handler / DB queries. A client disconnect cannot cancel the scan; a slow reach query holds the goroutine + DB cursor for the full duration. Pass `r.Context()` into `QueryContext`.

13. `cmd/server/node_reach_bench_test.go:43-54` — bench is **misleading**: 5000 rows, no `b.ReportAllocs()`, no attribution, only the scan. Add (a) `b.ReportAllocs()`, (b) sizes at 1k/10k/100k, (c) a bench that exercises `attributeDirections` over the scanned rows, (d) a bench with mixed-payload data (not all matching). Without this you cannot defend any of the perf claims.

14. `cmd/server/node_reach.go:283` — degree query + `first_seen` query are **two sequential round-trips** after the scan. Combine into one query, or fire concurrently with `errgroup`. Latency stacks.

15. `cmd/server/node_reach.go:177` — `reachCacheMax = 256` is **silently chosen** with no rationale. At ~2KB JSON per entry that's <1MB — fine, but document it, and consider sizing by total bytes (`raw`) not entry count to bound memory deterministically.

### Out-of-scope (pre-existing, noting for awareness)

- `reliableTokens` and `uniqueResolve` operate on the **current** prefix map; observations stored when a now-unique prefix was ambiguous will be attributed using today's resolution. This is a correctness concern that predates this PR but is amplified by it. Worth a follow-up issue.

### Verdict

**NEEDS-WORK** — the feature ships, but at 100k observations the endpoint as written will be slow on cold cache, thrash the cache on warm, and allocate aggressively per request. Items 1, 2, 4, 5, 7, 8 are the biggest wins; fix those and the rest are cleanup.

— Carmack Review (round 1)

Kpa-clawbot (changes requested):
## Tufte Review (round 1) — visualization & UI density

**Verdict:** NEEDS-WORK.
**Verification:** static review only — staging (analyzer-stg.00id.net) does not have this PR deployed yet (`GET /node-reach.js` returns SPA index fallback). CDP browser was started; no live route to exercise. Findings cite branch `feat/reach-upstream-pr` files.

The data shape is good and the page does the right work. The visual design, however, leaks ink and hides decoding rules from the viewer. Every must-fix below targets clarity, data-ink ratio, or accessibility — Tufte's "above all else, show the data" plus a11y basics.

### Must-fix (visualization & density)

1. **Map has no legend.** `public/node-reach-map.js:25-37` encodes bottleneck with **both** color (3 tiers) and line weight (log scale), yet there is no on-screen key for either. The viewer cannot decode red vs. yellow vs. green or what a thick line means. Add a compact in-map legend (color swatch + label "weak / medium / strong" and a 1-line note "thickness ∝ bottleneck observations").
2. **Double-encoding of bottleneck wastes a channel.** Same file, lines 28-31: color *and* width both encode `bottleneck`. Tufte: "show data variation, not design variation." Either drop one (prefer: keep color, drop variable width — make all polylines weight 2.5) or repurpose width to a different variable (e.g., total observations `we_hear + they_hear`, with the bottleneck still in color).
3. **Magic thresholds 300 / 100.** `public/node-reach.js:8-12` hard-codes color cutoffs with no rationale and no on-screen disclosure. Either derive from quantiles of the actual link distribution per node, or expose the cutoffs in the legend ("≥300 strong, 100–299 medium, <100 weak").
4. **Color-only encoding fails for red-green colorblind viewers.** Same color ramp `#1a7f37 / #bf8700 / #cf222e` (`public/style.css:62-65`) in the table bottleneck cell (`node-reach.js:32`) has no secondary channel. Add a glyph or label (●●● / ●● / ●) next to the number, or a textual tier ("strong/medium/weak"), so the encoding survives without color.
5. **No dark-mode variants for link colors.** `public/style.css:62-65` defines the three `--link-*` vars once globally. The repo has dark theme support; these will not adapt. Provide a `[data-theme="dark"]` (or whatever the project uses) override pair with brighter, higher-contrast hues.
6. **Print stylesheet does not scope the print.** `public/node-reach.css:21-25` only hides `.nq-noprint` and shrinks the map. The page's outer nav/header/sidebar (rendered by the SPA shell) will still print on top of the report. Either follow the spec's pattern (`body * { visibility:hidden }` + `#content *, #content { visibility:visible }`) or wrap the report in `#nq-report` and scope `@media print` to it.
7. **Duplicated 680px magic constant** in `public/node-reach.js:69` and `public/node-reach.css:23`. Move to one place (CSS only) and have JS read it via a CSS var or `getComputedStyle`. Also: 680px assumes letter/portrait; will clip landscape A4.
8. **`.nq-actions` lacks `flex-wrap`.** `public/node-reach.css:5` — `display:flex; gap:14px;` with `margin-left:auto` on the print button (`node-reach.js:133`). On narrow viewports the print button overflows. Add `flex-wrap:wrap` and drop or replace `margin-left:auto`.
9. **Table has heavy 4-sided gridlines.** `public/node-reach.css:11-12` borders every cell. Tufte: erase non-data ink. Use horizontal rules only (`border-bottom:1px solid …` on `td`, no vertical borders, no border on cells), and a single rule under the header. The table is dense; the grid is fighting the numbers.
10. **Uppercased 10px table headers.** `public/node-reach.css:13` — `font-size:10px; text-transform:uppercase` makes column labels harder to read than the data. Use sentence case at 11–12px.
11. **Stat-card descriptions are paragraph-length.** `public/node-reach.js:111-117` — descriptions like "Distinct neighbours in the all-time neighbour graph (built only from advert first-hop + observer last-hop, geo-filtered)." dominate the card visually and crush the data. Either move to a `?` info tooltip (`<abbr title="…">`) or trim to ≤ 8 words per card.
12. **Chartjunk: emoji decorations.** `public/node-reach.js:43` (`📡 … — Reach`) duplicates the "📡 Reach" button label/emoji from `public/nodes.js:572`, and the `ℹ️` in the disclaimer (`node-reach.js:45`) and `🖨️` on the print button (`node-reach.js:133`) are decoration not information. Drop the H2 emoji at minimum; the others are negotiable but follow the same rule.
13. **Inline styles instead of CSS classes.** `public/node-reach.js:39, 41-47, 86, 89, 95-117, 130-137` — lots of `style="…"` for layout and color, including a hard-coded `#ff6b6b` error color (`node-reach.js:89`) that bypasses the theme. Move to `node-reach.css` and use existing CSS vars (`--color-error` or equivalent) for the error path.
14. **Filter checkboxes have no group label / a11y.** `public/node-reach.js:130-132` — two raw `<input type="checkbox">` with sibling text. Wrap in `<fieldset><legend>Show one-way links</legend>…</fieldset>` and give each a `for=`/`id=` pair. Without a group label, AT users hear two disconnected toggles.
15. **Time-range buttons missing `aria-pressed`.** `public/node-reach.js:38` (`dayBtn`) and `wireTimeRange` toggle the `active` class only. Add `aria-pressed="true|false"` so screen readers announce the current window.
16. **Live count `#nqCount` is not announced.** `public/node-reach.js:144,176-177` updates row count on filter change. Add `aria-live="polite"` to the `.nq-count` span so the filter actually communicates to AT.
17. **"km" header is ambiguous.** `public/node-reach.js:147` — rename to "distance (km)" so the column reads on its own.
18. **"bottleneck" used without definition on the page.** Same row, column 5. The disclaimer explains tokens, not bottleneck. Add a one-line tooltip on the `<th>` ("smaller of we-hear/they-hear — gates two-way stability") or define it inline.
19. **Empty-state with no GPS-bearing links shows a near-empty map.** `public/node-reach-map.js:21-24` silently skips links with `lat==null`. If *all* skip, the map shows just the node pin with no lines while the table has rows — confusing. Detect and show an inline note ("X links have no location and are not drawn") above the map.
20. **`renderMap` runs on every checkbox change.** `public/node-reach.js:158-162, 181` — full Leaflet teardown + recreate per toggle is heavy and flashes the map. Either keep the map and update only the polylines/markers (preferred), or debounce.
21. **`<a>` "← Back" link has no `aria-label` and uses 12px font.** `public/node-reach.js:42` — small target, generic name. At minimum, `aria-label="Back to {nodeName} detail"`.
22. **`var(--text-muted, #57606a)` fallback hex duplicated 6+ times** in `public/node-reach.css`. The var is defined globally in the app; remove the inline fallbacks to a single source of truth, or accept they'll drift.

### Out-of-scope (pre-existing only)
- `public/nodes.js:570-573` — Detail-page action row already used inline styles before this PR; PR's added Reach button matches the existing pattern. Out of scope for this PR but worth a follow-up.
- The chartjunk emoji idiom is repo-wide; this PR is not where to rip them out, only where to stop adding new ones.

### Bottom line
Add a legend. Drop the double-encoding. Make the color survive color-blindness. Lighten the table grid. Fix the print scope. Then the page will show the data instead of the decoration.

— Tufte (round 1)

Kpa-clawbot (changes requested):

**Verdict: NEEDS-WORK**
**TDD history: FAIL** (no red commit precedes any production code; tests committed alongside or after prod)

### TDD history check
- `a45aae551b feat(server): add directional path attribution` adds `node_quality.go` (prod: `attributeDirections`) AND `node_quality_test.go` in the **same commit**. No prior red commit.
- `93f8449a6e feat(server): derive unique reliable path tokens` adds `reliableTokens`/`uniqueResolve` prod AND `TestReliableTokens` in the **same commit**.
- `87ea5a03ec feat(server): GET /api/nodes/:pubkey/quality endpoint` lands the HTTP handler with **no test commit** of any kind.
- `0806814df2 test(server): node quality endpoint shape/404/clamp` is **test-after**: endpoint already exists.
- `9d6f853dbf test(server): benchmark node quality scan` and `2e74c43673 test(e2e): node quality report section` are likewise test-after.
- `gh run list --commit <sha>` returns `[]` for all of the above — CI was never run on the intermediate SHAs (rebase/squash hid history), so we cannot independently confirm any test would have failed.

No "Net-new UI" exemption available for the Go endpoint/attribution logic — that exemption (per AGENTS.md) is scoped to net-new UI surfaces, not to brand-new backend behavior. The standalone Reach page itself (`bb21a085e7`, `6fde4aede3`) could qualify, but the E2E test (`2e74c43673`/`b7b7bf243b`) also did not land first and has tautology issues (see Must-fix #4).

### Must-fix
1. `cmd/server/node_reach_test.go` + `node_reach.go` (commits `a45aae551b`, `93f8449a6e`): TDD violation — production and tests committed together. Cannot demonstrate test gated the change. Rework history (squash a red test-only commit before the prod commit, push, prove red on CI).
2. `cmd/server/node_reach_endpoint_test.go` (commit `0806814df2`): test-after — endpoint shipped in `87ea5a03ec` without any failing test. Same rework requirement.
3. `cmd/server/node_reach_bench_test.go`: bench shipped after prod (`9d6f853dbf`); not a correctness gate. Add at least one assertion-shaped scan test (e.g. given N seeded observations with token "01FA", `scanReachRows` returns N rows with expected `observerPK`/`fromPubkey`/`payloadType` decoded) — currently `scanReachRows` has zero correctness coverage.
4. `test-node-reach-e2e.js:54` — `await page.waitForSelector('#nqMap .leaflet-container', { timeout: 10000 }).catch(() => {});` swallows the failure: tautology, map could be completely broken and the test still passes. Drop the `.catch` or hard-assert presence/absence.
5. `test-node-reach-e2e.js:42-48` — toggle assertions use `withIncoming < twoWay` and `withBoth < withIncoming`. With most fixtures rows stay equal (`==`), so the test cannot fail regardless of toggle correctness. Make the assertion direction-specific (e.g. count `we_hear`-only rows vs `they_hear`-only rows derived from the API response) or remove the toggle section.
6. `cmd/server/node_reach_endpoint_test.go:TestNodeReach_ShapeAndClamp` — seeds a node but **zero observations**, so `links`/`direct_observers` are necessarily empty. Test only validates JSON shape and `days` clamp; a completely broken `attributeDirections` or `scanReachRows` would still pass. Add an observation-level integration test: seed `transmissions`+`observations` with a known 3-hop path, assert `links[].we_hear`/`they_hear` counts.
7. `cmd/server/node_reach_test.go:TestReliableTokens` — only exercises 1-byte (collision) and 2-byte (unique). The implementation loops `[2,4,6]` hex chars (1/2/3 byte). The 3-byte (6-hex-char) branch has no test; a regression that drops it would not be caught.
8. `cmd/server/node_reach_test.go` — missing edge case: `payloadType != advertPayloadType` first-hop with non-empty `fromPubkey` must NOT credit `d.we[fromPubkey]`. No test asserts this; an off-by-one in the `else if` (e.g. dropping the `r.payloadType == advertPayloadType` guard) would silently inflate `we_hear`.
9. `cmd/server/node_reach_test.go` — missing edge case: same observer across multiple last-hop rows. `obsAgg.snrSum`/`snrN` aggregation across observations is not asserted; only single-row aggregation is covered.

### Out-of-scope
None — every concern above is in code added by this PR.

### Action
Squash/rebase to enforce red→green for the attribution + endpoint commits (or document why the net-new exemption applies and add observation-level coverage); then address tautologies #4 and #5 and the missing edges #6–#9.

Kpa-clawbot (changes requested):

Cold review of PR #1625. Reviewed only the PR diff (no local checkout). Scope: backend Go (`cmd/server/node_reach.go` + tests), `routes.go`, frontend (`public/node-reach*.{js,css}`, `public/nodes.js`, `app.js`, `index.html`, `style.css`), docs, and `test-node-reach-e2e.js`.

Verdict: **NEEDS-WORK**. Functionally the directional-attribution logic and the public route shape look right and the no-fabrication "unique prefix only" rule is the correct conservative call. But there are several real DoS / silent-error / coverage issues, plus a number of code-smell items, all of which fall under "must-fix" per the project's two-bucket severity policy.

### Must-fix

1. **`cmd/server/node_reach.go:265-279` — `scanReachRows` is an unbounded `LIKE '%"TOK"%'` sequential scan over `observations` in the window, with no `LIMIT` and no row cap. A repeater with broad reliable-token coverage on a 30d window can return millions of rows held entirely in memory (`out []pathRow`). Add a hard row cap (e.g. 200k) and surface "truncated" in the response, or stream-aggregate instead of materialising.
2. **`cmd/server/node_reach.go:194-208` — `handleNodeReach` does not validate `pubkey`.** Accepts any string. Should reject non-hex / non-64-char input with 400 before doing any work (cheap defense vs. cache-key pollution + wasted `buildNodeInfoMap()` calls on bogus IDs that still pass the `s.store==nil` gate).
3. **`cmd/server/node_reach.go:181-187` — "crude bounded reset" of the cache drops 256 entries at once.** Any user issuing 257 distinct real-pubkey requests (trivially scriptable) blows the cache for everyone. Use an LRU (or at least evict the oldest entry by `at`) rather than a full wipe.
4. **`cmd/server/node_reach.go:236-244` — both `s.db.conn.QueryRow(...).Scan(...)` calls swallow errors silently.** On a DB error the response will return `degree=0, rank=0, nodes_with_edges=0` / empty `first_seen` and look like real data. Log the error (and surface as 500 for the importance query, which is structurally required).
5. **`cmd/server/node_reach.go:215-217` — `json.Marshal` and `w.Write` errors ignored.** For the marshal at minimum, log on err and 500 instead of writing partial/empty body and caching it as success on the next path.
6. **`cmd/server/node_reach.go:170-178` — `reachCacheGet` returns the cached `[]byte` by reference.** Any future code mutating the returned slice corrupts the cache. Either document this invariant or `copy()` on return.
7. **`cmd/server/node_reach.go:240-242` — `SELECT first_seen FROM nodes WHERE LOWER(public_key)=?`** disables any index on `public_key`. The pubkey is already lowercased at handler entry (`strings.ToLower(mux.Vars(r)["pubkey"])`), so just use `public_key=?` (matches convention used elsewhere in the file's first-`buildNodeInfoMap` path).
8. **`cmd/server/node_reach.go:18-19` — `advertPayloadType = 4` duplicated locally.** Comment claims independence from decoder internals, but a stale local copy is a footgun if the protocol ever changes. Reference the canonical constant (e.g. `payloadTypeAdvert` from the decoder package) rather than re-declaring.
9. **`cmd/server/node_reach.go:127-130` — `minInt` is reinvented.** Go 1.21+ has `min(a, b)` builtin. Drop the helper.
10. **`cmd/server/node_reach_endpoint_test.go:31-37` — `TestNodeReach_UnknownNode` exercises the `s.store==nil` early-return, NOT the "valid handler, pubkey not in nodes table" 404 path** (the documented contract behavior). Add a test with a fully-wired server and a syntactically-valid pubkey that isn't in `nodes`.
11. **`cmd/server/node_reach_endpoint_test.go:39-65` — `TestNodeReach_ShapeAndClamp` inserts the node but no observations, so `Importance.RelayObservations`, `Links`, and `DirectObservers` are all zero by construction.** Add an end-to-end attribution test at the HTTP layer: insert transmissions+observations+observers, hit the endpoint, assert non-zero `we_hear`/`they_hear`/`bottleneck` for a known neighbour.
12. **`cmd/server/node_reach_endpoint_test.go` — no test for cache hit/miss, no test for `IsBlacklisted` returning 404.** Both branches are live in production and untested.
13. **`cmd/server/node_reach_bench_test.go:32-37` — bench `tx.Exec` errors are ignored** and the schema omits `transmissions(payload_type)` index used by the real attribution path. Bench result is best-case and silent on insert failures.
14. **`cmd/server/node_reach.go:288-298` — `pathRow` decoding silently discards any row where `json.Unmarshal` fails or `Scan` errors.** Add a counter (or at minimum a debug log) — silently discarding malformed rows hides ingest bugs.
15. **`public/node-reach.js:18-21` — `statCard(label, value, desc)` interpolates `value` raw.** Numbers today, but defense in depth: stringify/escape. Same applies to `imp.degree_rank` / `imp.nodes_with_edges` in `statsHtml`.
16. **`public/node-reach.js:99-115` — every toggle re-renders the Leaflet map** (`renderMap` calls `qmap.remove()` then re-creates). For a node with many links this is wasteful and causes visible map flicker. Update polyline visibility/styles in place instead.
17. **`public/node-reach.js:122-130` — `destroy()` does not abort an in-flight `load()`** (which then mutates `container.innerHTML` on a now-foreign container). Wire an `AbortController` into `api()` or track a request token.
18. **`public/node-reach.js:18-21` — `statCard` HTML template hand-rolled with string concat.** Inconsistent with rest of file using template literals elsewhere; convert for readability.
19. **`public/node-reach.js:74` — `'Loading reach…'` re-flashes on every time-range click**, wiping the already-rendered stats. Either keep the current shell and only re-render the body, or show an in-place spinner.
20. **`public/node-reach.js:130-133` — `destroy()` doesn't null `current`'s previous request, and there's a module-level `current` shared across pages.** Defensive code-smell; either remove `current` (it's unused outside `load`) or scope it inside `init`.
21. **`public/node-reach-map.js:13-17` — `cssColor()` reads `getComputedStyle` per polyline.** For nodes with many links this is unnecessary layout thrash; resolve the three tier colours once at the top of `render()` and pick by tier.
22. **`public/node-reach-map.js:33-37` — fallback `L.tileLayer(...)` hardcodes `tile.openstreetmap.org`** and dodges the user's tile preference if `_applyTilesToNodeMap` is ever unset. Either always require the helper (assert) or remove the fallback so missing-helper is loud, not silent.
23. **`test-node-reach-e2e.js:45-51` — `withIncoming < twoWay` only asserts rows didn't shrink.** If the chosen repeater happens to have ONLY two-way links (very plausible), the toggles assert nothing. Pick a fixture with known asymmetric links, or assert toggles actually add rows when asymmetric links exist (and skip otherwise with a clear message).
24. **`test-node-reach-e2e.js:11-15` — `nodes.nodes.length` check, then unconditional `nodes.nodes[0]`** — fine, but `await page.request.get(...).json()` has no status check. Fails opaquely on a 5xx.
25. **`docs/api-spec.md` (new section) — does not document the cache TTL (5 min), the row-scan upper bound (none today), or the row-cap behaviour when added.** Operators reading the spec will be surprised by sub-30s polling returning identical results.
26. **`cmd/server/routes.go:241` — new route added between `/neighbors` and `/{pubkey}` with no comment.** mux ordering matters here (more-specific before catch-all); add a one-line comment to prevent a future re-order that breaks routing.

### Out-of-scope (file as issues, do not block this PR)

1. The general pattern of `LIKE '%"TOK"%'` over `path_json` is used elsewhere in CoreScope and the underlying attribution model would benefit from a normalised `path_hops(observation_id, position, hop_token)` table with a covering index. This PR's perf is bounded by that pre-existing design choice.
2. There is no global rate-limit middleware on `/api/nodes/{pubkey}/*` endpoints. This PR adds another expensive endpoint into the same unprotected family.
3. The `directObs` SNR aggregation uses a simple arithmetic mean and ignores the existing SNR-quality conventions used elsewhere in the codebase. Worth a follow-up to align with the project-wide SNR display rules.

Once items 1–13 (backend correctness/safety) and 15–17 (frontend correctness) are addressed, this is in good shape — the core directional-attribution logic and the "unique-prefix only" reliability rule are the right design.

efiten:
Review feedback addressed (commit `e89ef764`). Thanks for the thorough pass — went through every item.

**Numbers** (bench, all-matching unless noted; remaining scan allocs are driver-side per-column `Scan`):
- 100k-row scan: `2.2M → 1.3M` allocs, `344ms → 203ms`
- attribution over 100k rows: **102** allocs total (size hints)

### Done

1. **Hard LIMIT** `reachScanRowLimit=200000` on the scan. Preallocate a modest `2048` hint rather than reserving the full cap (avoids reserving 200k `pathRow` for a 50-row result).
2. **No more double alloc** — dropped `json.Unmarshal` entirely (your option c): `parsePathTokens` does a single pass and slices into `path_json` (no parallel slice, no reflection, alloc-free for already-uppercase hops). New unit test covers it.
3. **`LOWER()` in SQL** for `obs.id` + `t.from_pubkey`; per-row `strings.ToLower` removed.
5. **Cache eviction** is now evict-expired-then-evict-oldest (`evictReachLocked`); no more full-map wipe.
6. **`sync.RWMutex`**; `reachCacheGet` takes `RLock`.
7. **singleflight** (`reachSF`, keyed `pubkey|days`) dedups concurrent cold-cache requests. (`golang.org/x/sync` was already an indirect dep; now direct.)
8. **Degree/rank from a 60s snapshot** — the `neighbor_edges` aggregate runs at most once per TTL and is shared; rank is a binary search over the descending degree list.
10. **Memoized resolver** (`newResolver`) — `ToLower` + prefix lookup once per distinct hop token, not per row.
11. **Size-hinted** `we`/`they`/`obs` maps (`len(rows)/8`).
12. **`context.Context`** plumbed handler → `computeNodeReach` → `scanReachRows`/`QueryContext` + `QueryRowContext`.
13. **Bench rewritten**: `ReportAllocs`, 1k/10k/100k sizes, a mixed-payload scan bench, and an isolated `attributeDirections` bench.
14. **Resolved by #8** — degree now comes from the snapshot (no SQL on the hot path), so only the `first_seen` query remains: one round-trip.
15. **Documented** `reachCacheMax` rationale (entry cap; ~2KB/entry → <1MB worst case).

### Pushing back (with reasons)

4. **The `path_json LIKE` scan** — agreed it's the dominant cost, but the deep fix is constrained:
   - An indexed/generated path-token column is a **schema migration**, and per the read/write-separation invariant (#1283) the server is read-only — all schema/migrations live in `cmd/ingestor/`. So that's a separate ingestor PR, not doable from here.
   - Gating by `transmissions.from_pubkey = ?` would **break correctness**: reach needs every observation where the node appears *anywhere* in the path (relayed traffic), not just its own adverts — `from_pubkey` only covers the originator.
   - For now the hard LIMIT bounds the worst case. I've left a comment pointing at the indexed-column follow-up. Happy to open that as an ingestor issue.

9. **`buildNodeInfoMap`** — it builds from the already-cached node slice (`getCachedNodesAndPM`, in-memory ~1.5k nodes) plus a ~48-row `observers` query; it's microseconds, not a hot cost, and isn't rebuilt per row. Left as-is to avoid churn, but I can wrap it in a short-TTL cache if you'd prefer.

### Out-of-scope note
Agreed on the `reliableTokens`/`uniqueResolve` current-prefix-map concern — it predates this PR (same model the rest of the resolver uses). I'll file a follow-up issue rather than widen this PR.

efiten:
Thanks all three — addressed the **backend correctness/safety + test coverage** in commit `3f7fed44` (the perf items from the earlier Carmack pass were in `e89ef764`). Frontend/Tufte items are batched into a dedicated UI/a11y pass next (called out below) so this commit stays reviewable.

## Independent review

**Done**
- #1 hard row LIMIT (`reachScanRowLimit=200000`) — landed in `e89ef764`.
- #2 pubkey validation → **400** before any work (`isHexPubkey`, 64-hex).
- #3 cache full-wipe → evict-expired-then-oldest — `e89ef764`.
- #4 swallowed `QueryRow` errors → logged (first_seen + degree snapshot).
- #5 `json.Marshal` error → log + **500** (no caching a bad body).
- #6 documented `reachCacheGet`'s shared-slice **immutable** invariant.
- #7 `first_seen` now `public_key=?` (pubkey pre-lowercased) — index usable.
- #8 dropped local `advertPayloadType`; use canonical `PayloadADVERT`.
- #9 dropped `minInt` → Go 1.21 `min()`.
- #10 added a wired-server "valid pubkey not in nodes → 404" test (the old one only hit the `store==nil` path).
- #11 added an **HTTP-level attribution test**: seeds tx+obs+observers for `A→N→B`, asserts non-zero `we_hear` (A) / `they_hear` (B).
- #12 added cache-hit + blacklist-404 tests.
- #13 bench now checks `tx.Exec` errors.
- #14 malformed/empty scan rows are counted + logged (no longer silent).
- #26 commented the mux specific-before-catch-all ordering.

**Pushing back**
- #4-deep / out-of-scope #1 (normalised `path_hops` table / indexed token column): agreed it's the right end-state, but it's an **ingestor-side schema migration** — the server is read-only by invariant (#1283), so it can't land here. The LIMIT bounds the worst case; I'll file it as an ingestor follow-up.
- #9 `buildNodeInfoMap`: it builds from the already-cached node slice (in-memory) + a small `observers` read; it's microseconds and not per-row. Left as-is.

**Deferred to the UI/a11y batch** (next commit): #15-#25 (statCard escaping, in-place map update, AbortController in `destroy()`, loading-flash, `cssColor` thrash, tile fallback, e2e status check + tautology).

## Kent Beck gate

**Tests added** (`3f7fed44`): scanReachRows decode coverage (#3), 3-byte reliable-token branch (#7), non-advert first-hop must-not-credit (#8), observer count+SNR aggregation across rows (#9), and the observation-level/HTTP attribution test (#6).

**TDD history (#1, #2):** declining the history rewrite. The branch was rebased onto upstream and the intermediate SHAs were squashed, so "red commit before prod" can't be reconstructed honestly after the fact — fabricating a red→green sequence via force-push would be theatre, not evidence. The behaviour is now covered by the expanded suite above (incl. the previously-missing edges you flagged). Happy to add more cases, but not to rewrite history to stage a narrative.

**#4, #5 (e2e tautologies):** valid — folding these into the UI batch since they need a browser run to verify, not just a diff edit.

## Tufte review

All 22 are legitimate and going into a **dedicated UI/a11y batch** (next): map legend, drop the color+width double-encoding, colorblind glyph/label alongside color, scoped `@media print`, `fieldset/legend` + `for/id` on the toggles, `aria-pressed` on time-range, `aria-live` on the count, lighter table grid, move inline styles/`#ff6b6b` to themed CSS vars, single-source the 680px/`--text-muted` constants. Kept out of this commit so the backend review stays clean.

CI green expected; will confirm.

Issue #1606, opened 2026-06-06
Frontend doesn't paginate /api/nodes after v3.8.3 silently clamped limit from 5000 to 500 — 78% of nodes invisible on the Nodes page for medium+ deployments


MarekWo:
## Summary

PR #1540 (v3.8.3 security train) introduced silent `limit` clamping on list endpoints — `/api/nodes` is now capped at 500 regardless of the client's `?limit=` request. The cap is a sensible DoS-prevention default, but `public/nodes.js:1117` still hardcodes `limit: '5000'` and treats the response as the complete node set ("Fetch all nodes once, filter client-side"). There is no pagination loop, and `response.total` is ignored after the first batch.

Net effect: on a deployment with more than 500 nodes, the Nodes page **silently shows only the top 500 sorted by `last_seen DESC`** — operators see nothing older than ~2h on a busy mesh, with no UI indication that anything was truncated.

## Reproduction (our deployment, `analyzer.marwoj.net`, v3.8.3)

```bash
$ curl -s "https://analyzer.marwoj.net/api/nodes?limit=5000" | jq '{nodes_len: (.nodes | length), total, counts}'
{
  "nodes_len": 500,
  "total": 2313,
  "counts": {
    "companions": 398,
    "repeaters": 1831,
    "rooms": 82,
    "sensors": 2
  }
}
```

`total` correctly reports 2313, but `nodes` is capped at 500. Top 500 sorted by `last_seen DESC` covers only ~2h of activity on a network with ~10k packets/h. The remaining **1813 nodes** (78% of the database) are unreachable through the UI — the user cannot scroll to them, sort to them, or filter to them. They exist in the DB and are queryable via `/api/nodes?offset=500&limit=500` from curl, but the frontend never makes that call.

This compounds with #1598 ("Active nodes disappearing"): the operator workaround for #1598 (raise `infraSilentHours` so older-but-relaying nodes count as Active) has **zero visible effect** when those nodes are below the hardcap cutoff anyway.

## Where it lives

**Single fetch in `public/nodes.js:1113-1128`:**

```javascript
async function loadNodes(refreshOnly) {
  try {
    // Fetch all nodes once, filter client-side
    if (!_allNodes) {
      const params = new URLSearchParams({ limit: '5000' });
      const rp = RegionFilter.getRegionParam();
      if (rp) params.set('region', rp);
      const ap = AreaFilter.getAreaParam();
      if (ap) params.set('area', ap);
[...]

PR #1637, opened 2026-06-09
fix(nodes): paginate /api/nodes across map/live/analytics/packets/area-map (500-row cap)

efiten:
## Summary

The server clamps `/api/nodes` `?limit` to **500** (DoS guard, PR #1540 / v3.8.3) and orders by `last_seen DESC`. Every node-list consumer issued a single big-`?limit` fetch and trusted it as the full set, so on >500-node meshes the top-500-by-advert window silently hid the tail.

Because `nodes.last_seen` is updated **only on self-adverts** (never on relay traffic; `UpsertNode` is called solely from the advert path), a repeater that relays constantly but last advertised hours ago fell outside that window and **vanished from the map and live view** — while still showing "Active" in its detail panel and (since #1606) in the paginated Nodes list.

#1606 fixed only the Nodes page (`nodes.js`). This generalizes that fix to the deferred siblings.

## Changes

- **`public/app.js`** — new shared `fetchAllNodes(extraQuery, opts)`: pages `limit=500` + `offset` until a short page (the server's `total` is unreliable — clamped to the page size and overwritten with the filtered length under area/region filters, so we stop on a short page, not on `total`), dedups by `public_key`, returns the real deduped count as `total`.
- **`public/map.js`**, **`public/live.js`** (keeps the `LIVE_MAP_MAX_NODES` ceiling via `safetyCap`), **`public/analytics.js`** (×2), **`public/packets.js`** now use the helper.
- **`public/area-map.html`** is standalone (cross-origin `baseUrl`, no `app.js`) so it gets an inline copy of the same loop.
- **`.eslintrc.json`** — declare `fetchAllNodes` global (no-undef).

## Tests

- **`test-fetch-all-nodes-pagination.js`** — unit-tests the helper via the real `api()`+`fetch` path: pagination past 500, short-page stop vs. the unreliable server `total`, dedup across a page boundary, counts pass-through, `safetyCap` bound. 5/5.
- **`test-map-nodes-pagination-e2e.js`** — browser E2E (Playwright) proving `map.js` surfaces a 501st node reachable only on page 2 and renders its marker. Verified **red→green**: against the pre-fix single fetch all 3 assertions fail (500 nodes, page-2 node absent, no marker); after the fix all pass. Wired into `deploy.yml`.

## Verification

- unit 5/5, E2E 3/3, `test-frontend-helpers.js` 611/611, `npx eslint public/*.js` → 0 errors.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -209,6 +209,7 @@
     "escapeHtml": "readonly",
     "exports": "readonly",
     "favStar": "readonly",
+    "fetchAllNodes": "readonly",
     "filterPacketsByRoute": "readonly",
     "formatAbsoluteTimestamp": "readonly",
     "formatChartAxisLabel": "readonly",
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -354,6 +354,7 @@ jobs:
           BASE_URL=http://localhost:13581 node test-channel-issue-1087-e2e.js 2>&1 | tee -a e2e-output.txt
           BASE_URL=http://localhost:13581 node test-channel-issue-1111-e2e.js 2>&1 | tee -a e2e-output.txt
           BASE_URL=http://localhost:13581 node test-map-modal-fluid-e2e.js 2>&1 | tee -a e2e-output.txt
+          CHROMIUM_REQUIRE=1 BASE_URL=http://localhost:13581 node test-map-nodes-pagination-e2e.js 2>&1 | tee -a e2e-output.txt
           BASE_URL=http://localhost:13581 node test-observer-iata-1188-e2e.js 2>&1 | tee -a e2e-output.txt
           CHROMIUM_REQUIRE=1 BASE_URL=http://localhost:13581 node test-nav-fluid-1055-e2e.js 2>&1 | tee -a e2e-output.txt
           CHROMIUM_REQUIRE=1 BASE_URL=http://localhost:13581 node test-nav-priority-1102-e2e.js 2>&1 | tee -a e2e-output.txt
--- a/public/analytics.js
+++ b/public/analytics.js
@@ -2148,7 +2148,7 @@
     try {
       const rq = RegionFilter.regionQueryString() + AreaFilter.areaQueryString();
       const [nodesResp, bulkHealth] = await Promise.all([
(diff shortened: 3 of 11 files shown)


efiten:
Thanks for the thorough review — all points addressed in `5bf9bbe9`.

**MAJOR**
1. **`live.js` safetyCap ceiling** — `fetchAllNodes` now treats `safetyCap` as a hard node-count ceiling: it slices the deduped result to `safetyCap`, so `LIVE_MAP_MAX_NODES` is honored exactly (no `pageSize-1` overshoot). Comment reworded to match. New unit test `safetyCap is a hard node-count ceiling (no pageSize overshoot)` asserts `safetyCap:800` → exactly 800.
2. **`area-map.html` duplicate untested + silent `break`** — the inline `fetchAllNodesPaged` now **throws** on a non-OK page instead of returning a truncated, complete-looking list; `loadAreaNodes`' empty `catch` now `console.error`s. Added tests that extract the inline helper from the HTML and assert pagination+dedup **and** the throw-on-non-OK contract, so the two copies can't drift silently.

**MINOR**
3. **Falsy `public_key` dedup** — rows without a `public_key` now get a unique synthetic key instead of collapsing into one (app.js + area-map). Test added.
4. **Red→green** — acknowledged: single commit, but the E2E (`test-map-nodes-pagination-e2e.js`) is red pre-fix (asserts a page-2 node + its marker, both absent on the single-fetch code) and green after. Verified locally.
5. **Loose safetyCap assertion** — tightened to `strictEqual(1000)`, and the new 800-overshoot test would catch an off-by-`pageSize` regression.

**Also corrected (caught locally):** the comments/test docblocks claimed a hard "500-row cap (#1540)". That's stale — the cap is `listLimits.nodesMax` (default **2000**, configurable; #1540's 500 was raised in **#1589**). Pagination is needed because a mesh can exceed `nodesMax`, not because of a 500 cut. The server `total` is a real `COUNT(*)` (overwritten only by area/geo/blacklist handler filtering, not region) — comments fixed accordingly, plus the app.js doc contradiction about `total`. Also gave `packets.js`' empty `catch` in `ensureHopResolver` a `console.warn`.

Verification: unit 11/11, E2E 3/3, `eslint public/*.js` 0 errors.

Issue #1727, opened 2026-06-15
Proposal: opt-in mobile client-RX coverage (crowdsourced RF reach) + /api/nodes/resolve


efiten:
## Summary

I'd like to contribute an **opt-in, crowdsourced RF-coverage** feature, plus a small node-name lookup endpoint. The whole thing is gated behind one config flag and is **disabled by default**, so it has **zero impact on existing deployments** unless an operator turns it on. A clean branch off `master` is ready (`efiten:feat/client-rx-coverage-pr`) and I can open a PR — opening this issue first to check the design is welcome, since the feature has a companion-app dependency worth discussing up front.

## What it adds

**1. Mobile client-RX coverage.** A roaming MeshCore **companion** radio — driven by an open-source Android PWA, [corescope-rx](https://github.com/efiten/corescope-rx) (GPLv3) — reports which nodes it heard *directly*, tagged with the phone's GPS and the packet's SNR/RSSI. CoreScope ingests these into a new `client_receptions` table and renders:
- per-node **hex coverage** as an overlay on the existing Reach page, and
- a standalone **Coverage dashboard** (`#/rx-coverage`) with a top-mobile-observers leaderboard.

**2. `GET /api/nodes/resolve?prefix=<hex>`** — a tiny read-only lookup that resolves a pubkey prefix to `{name, pubkey, ambiguous}`. The companion app uses it to show friendly node names; it's generally useful and not coverage-specific.

## How it works

```
companion ──BLE 0x88 (snr+rssi+raw)──▶ corescope-rx PWA ──▶ MQTT meshcore/client/{pubkey}/packets
                                                                      │
                                          ingestor (gated) ──▶ client_receptions (GPS + SNR + heard-key)
                                                                      │
              server: pure-Go hex grid ──▶ GeoJSON ──▶ Reach hex overlay + Coverage dashboard
```

- **Direct-only capture:** records only what the companion heard itself and directly — a 0-hop advert's pubkey, or `path[last]` (the last forwarder) for FLOOD routes; a ≥2-byte path-hash is required. Upstream hops are discarded.
- **Pure Go, no new deps:** hexbins are a pointy-top grid over Web Mercator (`cmd/server/hexgrid.go`) computed at query time — `CGO_ENABLED=0` / `modernc.org/sqlite` friendly. The frontend draws with the existing Leaflet.
[...]

PR #1728, opened 2026-06-15
feat: opt-in mobile client-RX coverage (crowdsourced RF reach) + /api/nodes/resolve

efiten:
Implements #1727.

## What this adds

**Mobile client-RX coverage** — an opt-in, crowdsourced RF-coverage feature. A roaming MeshCore **companion** radio (driven by the open-source [corescope-rx](https://github.com/efiten/corescope-rx) PWA, GPLv3) reports which nodes it heard directly, tagged with the phone's GPS and the packet's SNR/RSSI. CoreScope ingests these into a new `client_receptions` table and renders per-node **hex coverage** on the Reach page, plus a standalone **Coverage dashboard** (`#/rx-coverage`) with a top-mobile-observers leaderboard.

Also includes **`GET /api/nodes/resolve?prefix=<hex>`** — a read-only node-name lookup by pubkey prefix (`{name, pubkey, ambiguous}`), used by the companion app for friendly names.

## Opt-in — default OFF (zero impact on existing deployments)

The whole feature is gated behind one config flag, **disabled by default**:

```jsonc
"clientRxCoverage": { "enabled": false }
```

When disabled (the default): the ingestor writes **no** `client_receptions`; the three coverage endpoints return a clean **404**; the UI hides the Coverage nav link, the `#/rx-coverage` route, and the Reach-page toggle. `/api/nodes/resolve` is always available (not coverage-specific).

## How it works

```
companion ──BLE 0x88 (snr+rssi+raw)──▶ corescope-rx PWA ──▶ MQTT meshcore/client/{pubkey}/packets
                                                                      │
                                          ingestor (gated) ──▶ client_receptions (GPS + SNR + heard-key)
                                                                      │
              server: pure-Go hex grid ──▶ GeoJSON ──▶ Reach hex overlay + Coverage dashboard
```

- **Direct-only capture:** records only what the companion heard itself and directly — a 0-hop advert's pubkey, or `path[last]` (last forwarder) for FLOOD routes; ≥2-byte path-hash required. Upstream hops discarded.
- **No new deps:** hexbins are a pure-Go pointy-top grid over Web Mercator (`cmd/server/hexgrid.go`) computed at query time (`CGO_ENABLED=0` / `modernc.org/sqlite` friendly); frontend uses the existing Leaflet.
- **Trust:** companion pubkey = identity; an EMQX ACL binds each client to publish only to its own `meshcore/client/{pubkey}/packets` topic. Payload contract in `docs/client-rx-coverage.md`.

## How to enable / try it

1. In `config.json`, set `"clientRxCoverage": { "enabled": true }` and restart server + ingestor.
2. Point an EMQX (or any broker) listener so a client can publish to `meshcore/client/<pubkey>/packets`; the ingestor already subscribes under `meshcore/#`.
3. Run the [corescope-rx](https://github.com/efiten/corescope-rx) PWA on an Android phone paired (BLE) to a MeshCore companion — it captures heard nodes + GPS and publishes.
4. View results: per-node Reach page → toggle **coverage**, or the **Coverage** dashboard at `#/rx-coverage`.

## What's where

- **Ingestor:** `cmd/ingestor/client_reception.go` (ingest), `db.go` (`client_receptions` + `client_observers` schema), `main.go` (gated dispatch), `config.go` (flag).
- **Server:** `cmd/server/rx_coverage.go` + `rx_dashboard.go` (endpoints, self-guard 404 when off), `hexgrid.go` (pure-Go grid), `node_resolve.go` (resolve), `routes.go` / `types.go` / `config.go` (wiring + flag + `/api/config/client` field).
- **Frontend:** `public/rx-coverage.js` (dashboard), `node-reach-coverage.js` + `.css` (overlay), `node-reach.js` (Reach toggle, flag-gated), `roles.js` (reads the flag, hides nav when off).
- **Docs:** `docs/client-rx-coverage.md`.

## Testing

- Go: `cd cmd/server && go test ./...` and `cd cmd/ingestor && go test ./...` — green, including new gate tests (`coverage_gate_test.go` in both: off → no rows / 404, on → works) and the rx-coverage / resolve / hexgrid suites.
- JS: `node test-coverage-gate.js`, `node test-node-reach-coverage.js` (wired into CI). The Playwright `test-node-reach-coverage-e2e.js` is wired into the e2e job and **skips when `clientRxCoverage` is disabled**, so it's safe under the default-off config.

## Notes for reviewers

- The four new routes are registered in `cmd/server/openapi_known_gaps.json` (the existing OpenAPI-completeness ratchet), matching how other not-yet-spec'd routes are tracked. Happy to write full OpenAPI spec entries instead if you prefer.
- Commits are split per layer (ingestor / server endpoints / resolve / frontend / CI) for review.


efiten:
## Review round 1 — all 20 must-fix items addressed

Thanks for the thorough pass. Every numbered item is fixed, each in its own commit with a test that fails without the fix. The opt-in gate (default OFF ⇒ 404 + no ingest + UI hidden) is preserved throughout, and there is no fork bleed (matomo/locode/on8ar/_paq/piwik). Summary by item:

### Security / trust
- **#1 blacklist bypass via client topic** — `handleMessage` now enforces `IsObserverBlacklisted` before the client-coverage dispatch, so a blacklisted operator can't skirt it. (`44fc6ac8`)
- **#2 unvalidated `rxPubkey`** — the companion pubkey from the topic must be lowercase hex (`^[0-9a-f]{2,64}$`) before any write; non-hex is dropped. (`6643280c`)
- **#10 dead `origin_id` fallback** — removed; trusting a payload-supplied id would defeat the ACL model, and with #2 it was unreachable anyway. The identity comes only from the ACL-bound topic. (`6643280c`)
- **#14 HTML injection via `data-rx`** — `escapeHtml` now applied to the pubkey (both `data-rx` and the truncated label), not just the name. (`970b3b70`)
- **#15 `/api/nodes/resolve` enumeration + identity leak** — minimum prefix raised to 4 hex (matches the 2-byte heard-key floor, kills the 256-prefix sweep); blacklisted/hidden-prefix nodes now resolve as not-found (parity with `/api/nodes/search` & `/api/resolve-hops`), applied to coverage tooltips too. Endpoint kept and justified; per-IP rate-limit filed as follow-up. (`4183b14a`)

### Correctness
- **#3 dead `mobileRxStats`** — wired into the per-node response as `mobile_receptions` / `mobile_clients` totals. (`0b5bdfd8`)
- **#4 nil-cfg panic** — `requireClientRxCoverage` guards `s == nil` / `s.cfg == nil`; `ClientRxCoverageEnabled()` is nil-receiver-safe. (`197e5f2e`)
- **#9 mis-named advert test** — renamed to `...RelayedAdvert...` and added a true 0-hop advert test driving the `src='advert'` path. (`7cebe71a`)
- **#20 name flap** — node identity locked to the most-specific (longest) heard_key, order-independent; mixed-resolver test added. (`c7806460`)
- **#8 non-deterministic GeoJSON order** — features sorted by `cell`. (`c7806460`)

### Performance / scale
- **#5 / #18 indexes** — added `(heard_key, heard_keylen, lat, lon)` + `(lat, lon)`. Benchmarking at ~1M rows then showed the **OR/`substr` shape couldn't use the index** (2.28 s, slower than a 0.30 s scan), so the per-node query was rewritten to a sargable `heard_key IN (pubkey, prefix6, prefix4)` list → **0.42 ms (~5400×)**, bounded by node not table size. (`15b20f8a`, `f46f67df`, `67358053`)
- **#11 uncapped per-cell nodes** — capped server-side (25) with `nodes_truncated`. (`22655fc1`)
- **#12 unbounded response** — feature collection capped (5000, densest kept) with top-level `truncated`. (`22655fc1`)
- **#17 high-latitude NaN** — latitude clamped to ±85.05° in `hexCellAt`. (`eac0e511`)

### Frontend UX / robustness
- **#6 redraw storm** — pan/zoom redraws debounced (200 ms). (`102ec7e4`)
- **#7 swallowed errors** — `console.warn` everywhere, plus a one-line in-page message on leaderboard failure. (`102ec7e4`)
- **#13 config-ready race** — both pages now await `MeshConfigReady` before reading the coverage flag. (`50da1416`)
- **#19 inline style on legend** — replaced with an `.is-hidden` class. (`dea1fc74`)

### Docs / config + triage items
- **#16** — documented that `clientRxCoverage.enabled` is a **single flag read by both processes**. (`e1b7d089`)
- **ACL note** — `docs/client-rx-coverage.md` now states plainly that the feature **requires an ACL-capable broker** (without it the topic, GPS and attribution are spoofable); the server/ingestor validation is documented as defense-in-depth, not a replacement. (`e1b7d089`)
- **Retention** — added `retention.clientRxDays`: a reaper bounds `client_receptions`/`client_observers` (the maintainer's unbounded-growth concern). (`6eba9371`)
- **Benchmark (carmack)** — `BenchmarkCoverageQuery` seeds ~1M rows over a metro bbox; numbers above are published in its commit message. (`f46f67df`)

Out-of-scope follow-ups (rate-limiter, H3 build-tag, `/api/stats` field, EMQX ACL assertion test) left for separate issues per the triage. **Not merging — leaving that to you.**

Re-spawn review:

Kpa-clawbot (changes requested):
### Independent review (round 2)

Cold review — verified the 20 round-1 must-fix items are real (read every fix in `gh pr diff`, cross-checked against the noted commits). All accounted for:

- **#1 blacklist on client topic** — `handleMessage` checks `cfg.IsObserverBlacklisted(parts[2])` before dispatch, plus test (`TestClientRxCoverageBlacklistedDropped`). ✓
- **#2/#10 hex pubkey validation** — `clientPubkeyRe` enforced; payload `origin_id` fallback removed; test covers `!@#$`, `companionpk`, `xyz`. ✓
- **#3 mobileRxStats wired in** — per-node response carries `mobile_receptions/mobile_clients`; asserted. ✓
- **#4 nil-safe gate** — `Server`/`Config`/`ClientRxCoverageEnabled` all nil-safe; `TestRequireClientRxCoverageNilSafe` covers it. ✓
- **#5/#18 sargable IN-list + composite index** — `coverageHeardKeyCandidates` + composite `(heard_key, heard_keylen, lat, lon)`; EXPLAIN-QUERY-PLAN test + 1M-row benchmark with numbers. ✓
- **#6/#7 debounced redraws + surfaced errors** — `debounce(refresh,200)` with stable handler, `console.warn` replaces empty `.catch`; in-DOM test fires 6-event burst → 1 fetch. ✓
- **#8/#20 deterministic features + name precedence** — `sort.Slice` by cell, longest-heard-key name wins; order-independent test. ✓
- **#9 true 0-hop advert** — rebuilt `1100 + payload`, asserts `src='advert'/keylen=32` + GPS captured. ✓
- **#11/#12 response bounds** — `coverageCellNodeCap=25`, `coverageFeatureCap=5000`, omitempty flags; tests cover both caps. ✓
- **#13 MeshConfigReady race** — both `rx-coverage.init()` and `node-reach.load()` await; `loadGen` race guarded; controllable-promise test. ✓
- **#14 escape pubkey** — `escapeHtml(o.pubkey)` for `data-rx` + label fallback; sliced-source test asserts no raw `" badge above the map).
2. **`cmd/server/rx_coverage.go:160-176` — `aggregateCoverage` sorts features twice when truncation applies** (densest-desc to slice, then cell-asc unconditionally). One pass after the slice is enough — second `sort.Slice` is cheap but the densest-first sort can be replaced with a partial selection.
3. **`cmd/server/rx_dashboard.go:33-45` — `heardKeyResolver`'s per-request cache is unbounded.** A wide query touching N distinct heard_keys grows the map O(N) without an eviction cap. Bound to, say, 4096 entries or skip caching when over a threshold; the existing tests pass either way.
4. **`cmd/server/config.go:320-325` — server `RetentionConfig` is missing `ClientRxDays`.** The ingestor reads it; the server's JSON struct silently drops the operator's `retention.clientRxDays` key on parse. Even though the server doesn't prune, the field-set mismatch is a future footgun — mirror the field (or add a `json:"-"` placeholder with a comment) for symmetry.
5. **`cmd/server/rx_dashboard.go:107-126,165-183` — `rxLeaderboard` and `queryCoverageFiltered` don't filter blacklisted `rx_pubkey`.** Ingest is blocked now (#1), but pre-blacklist legacy rows still surface a banned companion's name in the leaderboard and let an attacker target `?rx=<blacklisted>`. Defense-in-depth: add a `NOT IN (blacklist)` filter (or a `WHERE NOT EXISTS …`) mirroring how `/api/nodes/resolve` hides identities.
6. **`cmd/server/rx_coverage.go:queryCoverageRows / mobileRxStats` — no guard against empty `pubkey`.** Today the mux route guarantees non-empty, but `coverageHeardKeyCandidates("")` returns `[]` which would yield `WHERE heard_key IN ()` (SQL syntax error). One-line guard (`if pubkey == "" { return nil, nil }`) is cheap insurance against a future caller.

### Out-of-scope

1. Per-IP rate limiting on `/api/nodes/resolve` (acknowledged in author commit message as follow-up #1).
2. Optional companion-signed broker token (called out in docs as future hardening).

Verdict: **request changes** on the six nits above. The 20 round-1 fixes are real and well-tested.

efiten:
## Review round 2 — both must-fixes addressed (`588b4afd`)

Good catch — you're right that round 1 hid the node *name* but not the rest of the identity on two read paths. Both are fixed, each with a test that fails without the gate; the opt-in gate is untouched and the bleed check is clean.

### Must-fix #1 — `handleNodeRxCoverage` leaked geometry/counts for hidden nodes
The per-node coverage endpoint now mirrors `handleNodeReach` at the same `{pubkey}`:
- `isHexPubkey(pubkey)` → **400** on a non-full/non-hex key (parity).
- `s.cfg.IsBlacklisted(pubkey) || s.isPubkeyHidden(pubkey)` → **404** before any query, so the GPS hex bins and `mobile_receptions`/`mobile_clients` are no longer retrievable for a blacklisted or hidden-prefix node.

Test: `TestNodeRxCoverageHidesBlacklistedAndHidden` asserts 404 for both a blacklisted and a `🚫`-hidden node.

### Must-fix #2 — `/api/rx-leaderboard` returned blacklisted observers
`rxLeaderboard` now post-filters the result set (≤100 rows, no perf concern) with the same helpers the rest of the API uses:
- `IsObserverBlacklisted(rx_pubkey)` → row dropped entirely (handles pre-PR rows, blacklist-after-ingest, and config-reload lag).
- `IsBlacklisted(rx_pubkey) || IsNameHidden(name)` → name blanked.

Test: `TestRxLeaderboardHidesBlacklistedAndHidden` covers a normal contributor (kept), an observer-blacklisted one (dropped), and node-blacklisted + hidden-prefix ones (name blanked).

(One test-infra note: `setupTestDBv2`'s `nodes` table gained `foreign_advert` so `GetNodeByPubkey` — which `isPubkeyHidden` calls — works against the test schema.)

### Out-of-scope notes
- `/api/rx-coverage` `?node=` non-hex up-front rejection and the `/api/nodes/resolve` per-IP rate-limit: left as the acknowledged follow-ups (rate-limit = follow-up #1 in the PR body), as you flagged them non-blocking.
- The retention reaper and config-race surfaces you re-verified clean — thanks.

CI green on `588b4afd`. **Still not merging — that's your call.**

Re-spawn review:

efiten:
## Polish review (4-persona) — triaged; most items fixed

Thanks — genuinely useful sweep. Most of it is now addressed; the rest is deferred with rationale. Opt-in/default-OFF gate untouched, bleed-clean, CI green.

### 🚫 BLOCKER — per-observer movement trail
Operator decision: **kept as a documented tradeoff** rather than changing the map — fine resolution is what makes the *aggregate* map useful, and the feature is opt-in / OFF by default. You're right there's no auth; the gap the review most clearly identifies is **informed consent**, which was missing. Added a **"Privacy — contributor location is public"** section to `docs/client-rx-coverage.md` (`e147d8df`): it states the per-observer view reconstructs movements, that a **pseudonymous companion name does not mitigate it** (the *locations* identify; the pubkey links all points), tells operators to warn users and contributors to avoid carried devices, and documents the hardening levers (lower `clientRxDays`, auth proxy, coordinate coarsening / k-anonymity). Happy to implement coarsening or a k-anonymity gate on `?rx=` if you'd rather have code than a consent doc — your call.

### Fixed — backend (carmack)
- **`rx_at` unindexed** — added `idx_client_recept_rxat`; the retention `DELETE … WHERE rx_at < ?` and the leaderboard window now seek it (was a full scan under the writer lock — gap from the reaper). (`2b41dd25`)
- **Redundant `idx_client_recept_rxpk`** dropped (the `UNIQUE(rx_pubkey, …)` index covers it); dead `count==1` branch removed. (`2b41dd25`)
- **`heardKeyResolver` N+1** — replaced with a single batched resolve (`UNION ALL` of per-prefix `LIMIT 2` subqueries, one round-trip per 200-key chunk, bounded 2 rows/key, #15 hiding preserved). (`c11465a0`)

### Fixed — frontend a11y/theming (tufte) (`d6ea742f`)
- **Dark theme:** added `[data-theme="dark"]` `--nq-cov-*` variants.
- **Colour-blind:** added a redundant, monotonic fill-opacity ramp (strong>mid>weak>grey) to both layers, so tiers don't rely on hue alone; per-cell SNR stays in the tooltip.
- **Keyboard/SR:** leaderboard rows are now `role="button"` + `tabindex` + `aria-pressed` + `aria-label` with an Enter/Space handler and a `:focus-visible` outline; day buttons get `aria-pressed`.

### Pushback
- **`mobileRxStats` "unbounded scan"** — it was rewritten earlier (`67358053`) to the sargable `heard_key IN (…)` form, so it's an index seek bounded by the node's own reception count, and retention bounds it in time. Left all-time semantics intentionally (the "heard by N clients" total is more useful unwindowed); can add a `?days=` bound if you'd prefer it track the view window.

### Deferred follow-ups (with reasons)
- **e2e never runs on CI master** (flag off by default) — the fix is a CI workflow change to run the coverage e2e with the toggle on; that's a maintainer infra call (the e2e is already wired to self-skip when disabled), so I left `.github/workflows` alone. Happy to add the job if you want it.
- **Rate-limits** (`/api/nodes/resolve`, per-`rx_pubkey` write) — acknowledged follow-up #1.
- **MINORs:** `client_observers.name` length cap/sanitise, the 12 inline-style sites → CSS, and a behaviour assertion on the perf bench. Low-risk; can fold in on request.

CI green on the latest head. **Not merging — maintainer's call.** Tell me if you want the BLOCKER done in code (coarsening/k-anonymity), the e2e CI job, or any deferred MINOR pulled in.

Re-spawn review:

Kpa-clawbot (approved):
## Kent Beck Gate (round 2) — TDD + test quality 

**Verdict: APPROVED** — all 5 follow-up commits land with appropriate tests; no must-fix issues.

### Per-commit TDD audit

**`588b4afd` fix(coverage): hide blacklisted/hidden nodes on coverage + leaderboard** — ✅
- `TestNodeRxCoverageHidesBlacklistedAndHidden` (rx_coverage_endpoint_test.go) asserts 404 for blacklisted *and* hidden-prefix pubkeys at `/api/nodes/{pk}/rx-coverage`. Reverting the gate in `handleNodeRxCoverage` flips the response from 404→200 → assertion fails. ✓
- `TestRxLeaderboardHidesBlacklistedAndHidden` (rx_dashboard_test.go) covers three distinct behaviors: observer-blacklisted dropped entirely, node-blacklisted name blanked but row kept, hidden-prefix name blanked but row kept. Reverting the post-loop filter loop flips all three → assertions fail. ✓
- Six questions: tests assert observable behavior (HTTP code, map of pubkey→row), not internal call order. Setup is wide (4 rows × 3 hide policies) but justified — it's the matrix the fix is designed to cover.

**`2b41dd25` perf(coverage): index rx_at; drop redundant rx_pubkey index** — ✅
- Perf-exempt per AGENTS.md, but the commit *still* lands a real perf gate: `TestClientReceptionsRetentionUsesRxAtIndex` uses `EXPLAIN QUERY PLAN` and asserts `idx_client_recept_rxat` appears. A future drop of the index, or a query rewrite that bypasses it, fails the test on the assertion (not on perf flake). This is the right shape for a perf regression test. ✓
- Drive-by behavior change in `aggregateCoverage` (drop `count==1` guard, rely on lexical `>= ""`) is invariant-preserving but unasserted by a new test. Out-of-scope nit only.

**`e147d8df` docs(coverage): warn contributor location is public** — ✅ pure-docs exemption.

**`c11465a0` perf(coverage): batch heard_key resolution to kill N+1** — ✅
- Perf-exempt, but `TestBatchResolveHeardKeys` covers the four resolution semantics (unique, ambiguous, unknown, hidden-prefix) of the new batch function with one call. Reverting `batchResolveHeardKeys` back to a per-key loop wouldn't fail this test (it asserts correctness, not query count), so the test gates *semantic regressions in the new batched code path*, not the N+1 itself. Parent flagged this as a "is the query count pinned?" question — answer: not pinned, accepted under perf exemption.
- `resolveHeardKey` is now a thin wrapper over the batch path, so existing single-key resolve tests transitively exercise the batched code. Good consolidation.

**`d6ea742f` fix(coverage): a11y + dark-theme for coverage layers** — ✅
- `test-node-reach-coverage.js` asserts monotonic opacity ramp (`strong > mid > weak > grey`). Reverting `coverageFillOpacity` to constant `0.45` fails the strict-inequality assertion. ✓
- `test-rx-coverage-escape.js` adds three keyboard-a11y assertions on the rendered row (`role="button"`, `tabindex="0"`, `aria-pressed="(true|false)"`). Reverting the `renderBoard` HTML changes fails all three. ✓
- Six questions: tests name the *behavior* (`opacity must ramp strong>mid>weak>grey`, `row must be focusable`), not the implementation. Setup is minimal.

### Six-question pass over test files in this batch

- `cmd/server/rx_coverage_endpoint_test.go`: behavior-named, minimal setup, no tautologies.
- `cmd/server/rx_dashboard_test.go`: three new tests, each with a single distinct invariant. The `BatchResolveHeardKeys` cases map is the smallest test that catches each semantic class.
- `cmd/ingestor/client_reception_test.go`: `EXPLAIN QUERY PLAN` is exactly the right tool for the perf invariant; no flake surface.
- `test-node-reach-coverage.js` / `test-rx-coverage-escape.js`: strict-inequality + regex assertions are tight; no implementation-coupling.

### Out-of-scope (do not block)
1. `cmd/server/rx_dashboard.go` `aggregateCoverage` removed `count==1` guard — behaviorally equivalent (lexical `>=""` always true on first row) but unasserted; consider a one-line test if revisited.
2. Dropping `idx_client_recept_rxpk` is reasoned-about ("UNIQUE leading column"), but no test pins that the `?rx=` filter still hits an index. EXPLAIN-QUERY-PLAN-style test would harden it.

Kpa-clawbot (changes requested):
## Munger Review (round 2) 

Cold re-review of the 5 follow-up commits (`588b4af`, `2b41dd2`, `e147d8d`, `c11465a`, `d6ea742`) against the round-1 findings, applied with inversion + second-order incentives.

### What the round-2 fixes got right

- **588b4af** correctly closes the two identity leaks: `handleNodeRxCoverage` now mirrors `handleNodeReach` (400 on non-hex, 404 on blacklist/hidden) so GPS hex bins and `mobile_*` counts aren't fetchable at a hidden pubkey, and `rxLeaderboard` drops `IsObserverBlacklisted` and blanks names for blacklisted/hidden identities. Test coverage matches both gates. The `setupTestDBv2` schema patch (adding `foreign_advert`) is the right minimal change to make `isPubkeyHidden` real in tests.
- **2b41dd2** is a clean swap: `idx_client_recept_rxat` backs the reaper `DELETE WHERE rx_at < ?`, and `idx_client_recept_rxpk` is genuinely redundant because the `UNIQUE(rx_pubkey, heard_key, rx_at)` constraint already creates an `rx_pubkey`-leading index. Dropping the dead `na.count == 1` guard in `aggregateCoverage` is correct (`""` ≤ any RFC3339).
- **e147d8d** says the quiet part out loud: opt-in + default-OFF + fine resolution + per-observer view = informed-consent territory. The doc is explicit that a pseudonymous companion name does not mitigate the trail, names the further-hardening levers, and links it from the enabling steps.
- **c11465a** kills the per-key round-trip cleanly. Resolver is built once per request from the rows that will be aggregated, so every key the resolver is asked about is in the batch (no fall-through to a slow path). Per-request scoping means no cross-request cache to poison. The `'pfx' AS pfx` echo-back keys the per-prefix aggregator deterministically, and `LIMIT 2` per prefix preserves the unique-vs-ambiguous distinction.
- **d6ea742** addresses both round-1 a11y items (keyboard-reachable role="button" rows with aria-pressed + focus-visible, monotonic opacity ramp as a non-hue SNR cue) and the dark-theme palette gap.

### Must-fix

1. **`cmd/server/rx_dashboard.go` (`batchResolveHeardKeys`, ~lines 78–96): literal-interpolated `LIKE` is one regex change away from SQL injection.** The safety argument ("k is `hexPrefixRe`-validated") lives in a comment, not in the function. Anyone who later widens `hexPrefixRe` in `node_resolve.go` (uppercase, `:`, etc., for whatever reason) silently turns this into an injection sink. Either (a) parameterize: build `?` placeholders for both the `pfx` literal and the `LIKE` operand and pass `args ...interface{}`, or (b) re-assert inside the loop with a function-local stricter check before interpolating. "Show me the incentive": a future contributor optimizes the prefix regex and has no reason to grep this file.

2. **`cmd/server/rx_dashboard.go` (`rxLeaderboard`, ~lines 199–215): drop-after-LIMIT shrinks the leaderboard below what the caller asked for.** The SQL `LIMIT ?` is applied before the Go-side `IsObserverBlacklisted` / `IsBlacklisted` / `IsNameHidden` filter. If N of the top-100 contributors are blacklisted observers, the public leaderboard returns 100−N rows — not 100. Not a leak, but a quality regression on a public surface, and the second-order incentive is bad (a blacklisted operator can occupy a top slot for free, just invisibly). Fix options: push `NOT IN (observer_blacklist)` into the SQL (cheapest), or over-fetch (`LIMIT ?*2`) and re-cap after filtering.

3. **`cmd/ingestor/client_reception_test.go` (`TestClientReceptionsRetentionUsesRxAtIndex`): only the reaper is asserted; the leaderboard plan is not.** The commit message and index comment both promise the new index covers `WHERE rx_at >= ?` in `rxLeaderboard`, but there's no `EXPLAIN QUERY PLAN` test for the leaderboard SELECT. SQLite's planner is fond of choosing the UNIQUE-backed `rx_pubkey`-leading index instead, especially with the `GROUP BY rx_pubkey`. Add a second EXPLAIN test against the actual leaderboard query (or a representative form) so a future schema tweak doesn't quietly regress to a full scan under the writer lock.

4. **`cmd/server/rx_dashboard.go` (`batchResolveHeardKeys`, query error path, ~lines 102–106): silent fallback hides a real bug.** On `db.Query` error, every key in the batch is mapped to `(k, "")` and the function continues. Operationally this presents as "all resolutions ambiguous" — a coverage map with no names — and there's no log line. Either log the error once (already have a logger on `*Server`), or surface it up so the handler can decide. Inversion: how would I notice this is broken? — answer: I wouldn't.

### Out-of-scope (pre-existing or accepted)

- The privacy tradeoff itself (per-observer view = movement trail) is an accepted, documented operator decision.
- `clientRxDays` retention default lives outside this PR.
- The companion-side signed-token follow-up is already tracked.

### Verdict

NEEDS-WORK — 4 must-fix items, all small. The round-1 BLOCKERs are genuinely closed; the remaining items are hardening against the next change rather than today's behavior, which is the right time to fix them.

efiten:
## Round-2 must-fix items addressed (`22065640`)



1. **`batchResolveHeardKeys` LIKE injection surface** — fully parameterized now: bound `?` args for both the `pfx` echo and the `LIKE` operand, no interpolation of the prefix. Safe regardless of how `hexPrefixRe` later evolves.
2. **`rxLeaderboard` drop-after-LIMIT shrink** — over-fetches by the observer-blacklist size (`LIMIT limit + len(observerBlacklist)`) and re-caps to `limit` after the Go-side filter, so blacklisted top contributors can't shrink the public board. New test `TestRxLeaderboardLimitSurvivesBlacklistDrop` (2 blacklisted at the top → still returns exactly `limit` good rows in order).
3. **Leaderboard query plan test** — added `TestRxLeaderboardQueryIsIndexBacked` (EXPLAIN QUERY PLAN). It confirms your prediction: the leaderboard is served by the `UNIQUE(rx_pubkey, heard_key, rx_at)` **covering** index, not `idx_client_recept_rxat`. That's a covering scan (no table-heap access) on a retention-bounded table — acceptable — and the test guards against a regression to a bare table scan. I also corrected the `db.go` comment that wrongly claimed `idx_client_recept_rxat` backs the leaderboard window; that index backs the retention `DELETE` (separately tested).
4. **Silent query-error fallback** — `batchResolveHeardKeys` now `log.Printf`s the query error before falling back to "all ambiguous", so a broken resolver is visible in the logs instead of presenting as a nameless map.

Out-of-scope items left as noted. CI green on the push.

efiten:
### Heads-up: the e2e red here is a pre-existing fixture time-bomb on `master`, not this PR

The Playwright job fails at:

```
test-issue-1630-reach-mobile-e2e.js:40 — no repeater with reach links found in fixture
    at pickRepeaterWithReach
```

`pickRepeaterWithReach` scans `/api/nodes?role=repeater` and needs one with `>= 2` links from `/api/nodes/{pk}/reach?days=30`. It fails at the **data** step (before any page render), and the same failure is happening on **`master`** independent of this PR:

- master run **27637454126** (commit `1476b857`, 2026-06-16) failed with this exact error;
- the **same commit `1476b857`** *passed* on 2026-06-15.

Same code, pass→fail across one day ⇒ the e2e fixture's reach observations have aged past the rolling 30‑day window, so no repeater qualifies. This PR is Go-only on the coverage dashboard (`cmd/server/rx_dashboard.go`, batch resolver / leaderboard) — it doesn't touch `node_reach.go`, `/api/nodes`, the reach page, or the fixture. `Go Build & Test` is green here.

Not asking to fix it in this PR (it's infra, and it blocks `master` too) — flagging so the e2e red isn't attributed to the coverage change. Likely fixes on your side: regenerate the e2e fixture with now-relative reach timestamps, or have `pickRepeaterWithReach` widen the window / seed a guaranteed reach repeater.

efiten:
## Incremental review (frontier leaderboard) — both MAJORs fixed

Thanks. Both production-data MAJORs are fixed in `cmd/server/rx_dashboard.go`, with the requested regression test.

### MAJOR 1 — unbounded leaderboard scan
`rxLeaderboard` now:
- takes a `context.Context` and uses `QueryContext`, checking `ctx.Err()` in batches (every 2048 rows) so a client disconnect/timeout aborts the scan;
- caps the scan with a hard SQL `LIMIT leaderboardScanCap` (500k), `ORDER BY rx_at DESC` so a truncated window keeps the most-recent receptions, and logs `[rx-leaderboard] scan hit cap …` when truncated.

`handleRxLeaderboard` passes `r.Context()`.

### MAJOR 2 — score dilution by blacklisted observers
The per-cell denominator is now computed from a `cellCount` map that **excludes** observer-blacklisted and node-blacklisted pubkeys, so an operator of a blacklisted node parked in a cell can no longer silently lower everyone else's frontier weight. Name-hidden (but not blacklisted) observers are legitimate contributors and still count.

### Test (kent-beck)
`TestRxLeaderboardScoreNotDilutedByBlacklisted`: a blacklisted observer sharing a legit observer's only cell — asserts the legit score stays **1.0** (it's **0.5** without the fix) and the blacklisted observer is absent. (Point taken on the red→green split for ranking changes; noted for future.)

### NITs (not done — flagging for your call)
- Dutch score/cells tooltips: left as-is — they're the author's intentional copy, not mine to change unilaterally; happy to translate to English for codebase consistency if you want.
- Score `.toFixed(1)` tie display, `hexCellAt` `fmt.Sprintf` in the hot path (precompute int64 cell ids), `sortBoard()` on every render, and a `deploy-*.sh.example`/`docs/DEPLOY.md` breadcrumb — all low-risk; say the word and I'll fold any in.

### CI status
`Go Build & Test` ✅ and every coverage/leaderboard e2e ✅. Playwright is red **solely** on `test-issue-1630-reach-mobile-e2e.js:40` (`pickRepeaterWithReach`) — confirmed not a #1728 regression: the latest master run (`27637454126`) fails on the identical line, and that same master SHA (`1476b857`) passed on 06-15 then failed on 06-16, so it's an environmental/flaky failure independent of this branch. Docker is "skipped" only because it gates on the Playwright job. Nothing in this PR touches the reach-mobile repeater-selection path that test exercises.

Not merging — maintainer's call.

Re-spawn review:

efiten:
Heads-up on CI: the `Playwright E2E` red here is **not** from this PR — it's the pre-existing `test-issue-1630-reach-mobile-e2e.js` failure that's been failing on `master` since ~2026-06-16 (the e2e fixture's observations aged out of the 30-day reach window).

Fixed separately in **#1747** (a one-line `tools/freshen-fixture.sh` change, no app code). Once #1747 lands on master and this branch is brought up to date, this PR's Playwright will go green too. Deliberately kept out of this PR to keep its scope to the coverage feature.

`Go Build & Test` is green here; the Docker job only shows "skipped" because it gates on the Playwright job.

PR #1737, opened 2026-06-16
live: cap animation-canvas DPR at 1.5 and redraw at ~60fps

ArcanConsulting:
The live-map animation overlay re-clears and re-draws its full backing store
every animation frame. Two unbounded multipliers make that expensive:

1. devicePixelRatio is uncapped in updateAnimCanvas(). The canvas is already
   ~1.4x the screen area (20% pad per side), so at DPR 2-3 it allocates and
   fills 5-12x the screen's pixels per frame. Cap at 1.5 — lines stay crisp,
   per-frame fill cost drops up to ~4x on hi-DPI displays.

2. renderAnimations() reschedules via rAF with no rate limit, so on 120/144Hz
   displays it does 2-2.4x the work for no visible gain. Add a ~60fps guard.
   Progress is time-based (tickDt, itself capped at 32ms), so skipping frames
   preserves motion exactly. Paused frames fall through to the existing sleep.

No behavior change on a standard 60Hz / 1x-DPI display. Existing animation
tests (test-live-dt-cap-1524, test-live-anims) unaffected.

### DIFF
--- a/public/live.js
+++ b/public/live.js
@@ -34,6 +34,10 @@
   let packetCount = 0;
   let activeAnims = 0;
   const MAX_CONCURRENT_ANIMS = 20;
+  // Perf caps for the animation overlay (see updateAnimCanvas / renderAnimations).
+  const ANIM_MAX_DPR = 1.5;       // cap backing-store pixels on hi-DPI displays
+  const ANIM_MIN_FRAME_MS = 15;   // don't redraw faster than ~60fps (high-refresh guard)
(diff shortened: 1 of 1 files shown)


efiten:
Heads-up: this PR's Playwright failure was the stale-fixture reach bug, fixed in #1747 (merged to master). I updated this branch to pick up the fix and CI is re-running — no code review action needed from the failure itself. Please verify the re-run is green before merging.

PR #1754, opened 2026-06-20
fix: don't re-render node dots when scrubbing the Live timeline

djkazic:
## What

Scrubbing the Live page timeline no longer re-renders all node dots.

## Why

`vcrReplayFromTs()` ran `clearNodeMarkers()` (wiping `nodesLayer` and `nodeMarkers`) and then `loadNodes()` rebuilt every marker from scratch. A single scrub click destroyed and recreated the entire node layer; visible flicker plus unnecessary DOM work (even though `addNodeMarker()` already no-ops nodes that still exist).

## How

- `vcrReplayFromTs()` now clears only the transient animation/path layers.
- The time-scoped branch of `loadNodes()` reconciles against the existing markers: removes only nodes absent at the target time, adds genuinely new ones, leaves shared dots untouched.

## Testing

- `test-live.js`: 95/95 pass
- `node --check public/live.js` clean

### DIFF
--- a/public/live.js
+++ b/public/live.js
@@ -416,8 +416,13 @@
     var gen = VCR.replayGen;
     vcrSetMode('REPLAY');
 
-    // Reload map nodes to match the replay time
-    clearNodeMarkers();
+    // Refresh map nodes to match the replay time. Don't tear every dot down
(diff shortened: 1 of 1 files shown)


djkazic:
Fixed the must-fix.

PR #1762, opened 2026-06-21
feat(server): complete the #672 4-axis repeater usefulness score

ArcanConsulting:
Adds Coverage (harmonic reach) + Redundancy (Tarjan articulation) axes + composite & grade. Closes #672.
**TDD note (BLOCKER-1):** Community PR delivered as a single squashed commit, so there is no separate pre-fix failing-test commit — please accept as a community-PR exemption. The tests are *gating*, not just thorough: each axis test pins a specific topology outcome (coverage on line/star/disconnected/weight-sensitive; redundancy online/triangle/star/bridged-cliques), and an end-to-end `/api/nodes` surface test drives the whole pipeline and asserts the composite diverges from the Traffic axis. Inverting the `1/weight` distance, dropping the NaN/Inf reject, removing the `redundancyMinWeight` floor, or aliasing `usefulness_score` back onto `traffic_share_score` each break a specific assertion. The axis functions are pure (no hidden state), so the suite fully characterises the behavior without the red anchor.

Issue #672, opened 2026-04-08
feat: repeater usefulness score — is this repeater actually adding value to the mesh?


Kpa-clawbot:
## Origin

Operator feedback from AC3T: *"Is this repeater actually useful (in its current position and configuration)? Looking at paths, neighbors, coverage, etc."*

## User Story

As a mesh operator, I want to see a **usefulness score** for each repeater so I can identify redundant nodes, optimize placement, and make informed decisions about where to add or remove hardware.

## What Makes a Repeater "Useful"?

A repeater earns its place by doing one or more of these:

| Signal | What it means | High value | Low value |
|--------|--------------|------------|-----------|
| **Bridge** | Only path between mesh segments | Sole connector between clusters | All neighbors already reach each other |
| **Coverage** | Extends reach to unique areas | Connects observer pairs nobody else does | 5 other repeaters cover the same area |
| **Traffic** | Carries actual packets | High relay count | Zero or near-zero relay traffic |
| **Resilience** | Removing it hurts the network | Avg hop count increases significantly | Network topology unchanged without it |
| **Quality** | Provides faster/better paths | Higher SNR or fewer hops than alternatives | Every path through it has a better alternative |

### Inputs (all from existing CoreScope data)

1. **Bridge score** — betweenness centrality from `neighbor_edges` graph
   - Build undirected graph from neighbor edges
   - Compute betweenness centrality per node
   - Normalize to 0–1 range

2. **Coverage score** — unique observer-pair connectivity
   - For each repeater, count observer pairs that it connects (appears in `resolved_path`)
   - Subtract pairs that are ALSO connected through alternate paths not involving this repeater
   - Remaining = unique coverage contribution
   - Normalize to 0–1

3. **Traffic score** — relay volume from packet data
   - Count packets where this repeater appears in `resolved_path` (relay count)
   - Normalize against network median relay count
   - Cap at 1.0

4. **Redundancy score** — simulated removal impact
   - Remove node from neighbor graph
   - Recompute connected components — did the graph split?
   - Recompute average shortest path — did it increase?
   - Score = weighted combination of partition risk + path length increase
   - Normalize to 0–1

### Composite

```
usefulness = w_bridge * bridge + w_coverage * coverage + w_traffic * traffic + w_redundancy * redundancy
```

Default weights (tunable via customizer later):
- `w_bridge = 0.30` — structural importance
- `w_coverage = 0.25` — reach contribution
- `w_traffic = 0.20` — actual utilization
[...]

PR #1760, opened 2026-06-21
feat(analytics): repeater metric scatter tab

ArcanConsulting:
Repeater metric scatter tab. Closes #1763.

PR #1771, opened 2026-06-21
feat(map): Important Links overlay — B-weighted top routes

ArcanConsulting:
Public map overlay drawing the most important affinity links, weighted by the #672 usefulness axes (rank-by selector + top-N slider). Frontend-only. Pairs with #1762.


efiten:
Closing and immediately reopening this to get a CI result. Nothing is wrong with your PR and no action is needed from you.

This one never actually failed: its last pipeline was CANCELLED, so there has been no verdict on it at all. Meanwhile master has moved (#1924 fixed #1923, an E2E bug that was failing PRs which never touched the affected page), and a plain re-run would reuse the merge commit from the original run rather than testing against current master. A close/reopen is the way to get a fresh one.

Apologies for the notification noise.

efiten:
Correction to what I just wrote: I said the last pipeline was CANCELLED. The run is recorded as `failure`, not cancelled. The substance holds and is actually better than I put it: **no test failed**. On run 28999632816, `✅ Go Build & Test` and `🎭 Playwright E2E Tests` both came back **success**, and only `🏗️ Build & Publish Docker Image` was cancelled, which is what sank the run's overall conclusion. The CANCELLED I quoted came from the PR's status-check rollup, which reflects the cancelled downstream job rather than the test jobs.

So this PR has never had a real test failure. The fresh run will confirm whether that still holds against current master.

efiten:

**Verdict: approve the feature, request a change to how it is tested.** CI was CANCELLED, not
failed, so re-run it first.

The overlay is sensibly built: a pure ranking core (`computeTopRouteEdges`) separated from the
Leaflet drawing, importance as edge affinity weighted by the mean of the endpoints' #672 axis
score, a debounced slider, `--accent` cached and invalidated on `theme-refresh`, `escapeHtml` in
popups, and an empty-state hint that names the two reasons nothing rendered.

**The test approach is the thing to push back on.** `test-top-routes-overlay.js` extracts the
ranking core by `indexOf`-slicing `public/map.js` between the literals `const TOP_ROUTES_AXES` and
`function clearTopRoutes`, then `new Function`s the result. The author added a guard assertion for
the rename case, which is thoughtful, but the approach still breaks on any reordering of map.js
and it tests a string, not the module. Two other PRs in this same queue do it properly and are
worth pointing at: #1821 exports `applyObserverFilter` through `_packetsTestAPI`, and #1912 puts
`hashPrefixInfo` on `window`. Ask for the same here.

Smaller notes: modifying the existing accordion regression test
(`test-issue-1329-map-controls-accordion-e2e.js`) to skip the new hidden controls is justified and
the comment correctly explains why the accordion check still bites (that collapse is class-based,
not inline style), but it deserves a second reviewer's eye on principle. And
`/analytics/neighbor-graph?min_count=1&min_score=0` pulls the entire unfiltered graph to filter
client-side; that matches this project's bulk-fetch convention, but the payload grows with the
mesh.

---

**Update 2026-09-02.** Correction to the line above: this PR's run was recorded as `failure`, not CANCELLED. It still was not a test failure. On run 28999632816 both `Go Build & Test` and `Playwright E2E Tests` passed and only `Build & Publish Docker Image` was cancelled, which sank the run's overall conclusion. After a fresh run against master it is **fully green**. Only the test-extraction question above is still open.

efiten:
Rebased and continued as #1928. **Your commits, your authorship, unchanged.**

This went CONFLICTING through no fault of yours: #1760 landed first in the #1922 queue run and both PRs append a line to `test-all.sh` at the same spot. That was the whole conflict, and I resolved it by keeping **both** test lines rather than either.

Nothing else was touched. `public/map.js` and `test-issue-1329-map-controls-accordion-e2e.js` are byte-for-byte as you wrote them. Verified on the rebased tree: your `test-top-routes-overlay.js` 20 passed, #1760's `test-repeater-metric-scatter.js` 31 passed, `test-frontend-helpers.js` 627 passed, 0 failed.

I opened it as a new PR rather than force-pushing to your branch. If you would rather #1771 stayed the vehicle, say so and I will push the rebase here and close #1928 instead.

Also worth knowing: this PR's run was recorded as CANCELLED for a long time, which is part of why it sat. Its tests were never the problem.

efiten:
Merged via #1928, which is your two commits rebased onto master with your authorship intact. Closing this as the vehicle rather than the work.

The only thing I changed was the `test-all.sh` conflict, and I kept both test lines rather than choosing between yours and #1760's. `public/map.js` and the accordion E2E are byte-for-byte as you wrote them.

The one review point still open is the test-extraction approach — `test-top-routes-overlay.js` slices `map.js` by `indexOf` and `new Function`s the result, so it breaks on any reordering of that file. #1821 and #1912 both export the function under test instead. Happy to take that as a follow-up if you would rather not.

Issue #881, opened 2026-04-21
schema: store per-observation raw_hex (each observer sees different bytes on air)


Kpa-clawbot:
## Problem
`observations` table stores `path_json` per observer but NOT `raw_hex`. All observations for a given transmission point at the same `transmissions.raw_hex`. This is wrong.

Each observer receives a physically distinct over-the-air transmission:
- Closer observers see fewer path bytes (fewer forwarders have prepended their hash)
- More distant observers see more path bytes (whole forwarder chain)
- Flags byte (hops remaining) differs
- In principle signature/payload should match but transient corruption / decode differences exist

Currently the `raw_hex` pane on the packet detail page cannot update when switching observations because there's only one raw_hex per transmission stored.

## Fix
- Schema: `ALTER TABLE observations ADD COLUMN raw_hex TEXT` (nullable for historical rows)
- Ingestor: store the MQTT-reported raw_hex with the observation, not just dedup-aggregate it into transmissions
- `packets_v` view: expose observation-level raw_hex
- Server: `ObservationResp.RawHex` populated from observation row, not transmission
- Frontend: hex pane + byte breakdown already read from `effectivePkt.raw_hex` — will "just work" once server returns per-observation bytes
- Backfill: for historical rows, leave NULL; UI falls back to transmission raw_hex with a subtle note

## Out of scope
Reconstructing per-observer path bytes from `path_json` alone is impossible because we don't know the pre-path payload boundary reliably for all packet types. Need to store the bytes as received.

## Related
- #866 (partial frontend fix in #880 handles Observer/Path/Timestamp/Counter per-observation — hex pane remains stuck until this issue lands)

PR #1773, opened 2026-06-21
feat(server): store memory diagnostics + drop redundant obs.RawHex

ArcanConsulting:
Drops the redundant per-observation RawHex (~98MB on a live store; reader already falls back to tx.RawHex #881) and adds an opt-in /api/perf?mem=1 memory breakdown (flood-forward share + per-component bytes). Profiled against a live instance.

  **Savings substantiation:** live-instance profiling shows ~1.66M observations in the store, each previously carrying its own per-observation `raw_hex` (avg ≈118 hex chars ≈59 bytes) that exactly duplicates the parent transmission's `raw_hex`. Dropping the duplicate on every load/ingest path eliminates ≈98 MB of redundant in-memory
  storage plus ~1.66M string allocations, with no data loss — the read path (`enrichObs`) already falls back to `tx.RawHex` when `obs.RawHex` is empty (verified by the new safety-gate test). The patched build cannot be run against the live instance here; instead the new opt-in `/api/perf?mem=1` diagnostic lets operators measure the
  real before/after (`trackedMB` and the per-component breakdown) directly after deploy.


ArcanConsulting:
On the two remaining items:

- **TDD ordering (MAJOR):** As a community PR, please accept this as an exception — the safety-gate test `TestObsRawHexNotRetainedOnLoad` loads a fixture through the real `LoadChunked` path and proves `obs.RawHex==""` while the `tx.RawHex` fallback still serves `raw_hex`; it fails the moment any of the `RawHex:` assignments is 
restored (verified by hand).
- **Staging citation (MINOR):** The patched build cannot be run against the live instance here. The "~98 MB" figure is derived from live-instance profiling (~1.66M observations × ≈59 B of redundant raw_hex); the new `/api/perf?mem=1` diagnostic lets operators measure the real before/after directly after deploy.

The round-2 nits (flood-bytes assertion + `memBreakdownNote` const) are fixed in the current push.

PR #1781, opened 2026-06-23
feat(live): optional "Multibyte only" view filter (#1780)

### LINKED ISSUE #1780: Live view: optional "Multibyte only" filter to hide unreliable single-byte path-hash packets
## Problem

The live map renders all traffic, mixing single-byte and multibyte path-hash packets. Single-byte (1-byte) path hashes collide heavily at scale (~8-way at ~2k relays), so the paths they produce are unreliable — they add visual noise and can mislead anyone reading propagation on the live view.

As deployments grow and more repeaters advertise multibyte path hashing, it's useful to be able to look at **only** the trustworthy multibyte traffic.

## Proposal

Add an opt-in **"Multibyte only"** toggle to the live map controls (alongside Favorites / Realistic / Color-by-hash):

- When ON, packets whose path hash size is `< 2` bytes (single-byte, or unresolvable) are excluded from the entire live view — feed, map polylines/rain, and the packet counter — in both LIVE and REPLAY modes.
- **Default OFF** — no behavior change for existing users; the choice is remembered (localStorage).

A packet's path hash size is uniform across its whole path and is encoded in the path-length byte of `raw_hex` (`hashSize = (pathByte >> 6) + 1`), so classification is a cheap O(1) per-packet check. This is distinct from the existing global "hide 1-byte path hops" toggle, which filters individual hops within a path at every render site; this one filters whole packets, on the live view only.

## Scope

Frontend only. A small shared classifier in `hop-filter.js`, two render-time filter sites in `live.js`, the toggle UI + handler, unit + Playwright E2E tests, and a user-guide note.

Happy to open a PR (already implemented and tested).

efiten:
Closes #1780.

## What

Adds an opt-in **"Multibyte only"** toggle to the live map controls. When ON, packets whose path hash size is `< 2` bytes (single-byte, or unresolvable) are excluded from the entire live view — feed, map polylines/rain, and the packet counter — in both LIVE and REPLAY modes.

- **Default OFF** — no behavior change for existing users.
- Persisted in `localStorage` under `live-multibyte-only`.
- Distinct from the existing global "hide 1-byte path hops" toggle: that filters individual hops within a path at every render site; this filters whole packets, on the live view only. They share no state.

## How

- **`public/hop-filter.js`** — new pure, dependency-free classifier `MC_packetHashSize(rawHex, routeType)` returning `1|2|3`, or `0` when unresolvable. Reads the path-length byte from `raw_hex` (`(pathByte >> 6) + 1`), offset `5` for transport routes (route_type 0/3) else `1` — mirroring the existing `getPathLenOffset`/`computeBreakdownRanges` logic in `app.js`. Lives next to the existing `hopByteLen`/`MC_*` family; `app.js` is untouched (no duplication of the byte math).
- **`public/live.js`** — `groupIsMultibyte(packets)` consumes that helper; applied at two render-time sites: the top of `renderPacketTree` (above the counter increment, so the counter reflects multibyte-only) and inside the `rebuildFeedList` group loop (so toggling re-filters the buffered feed). Toggle markup + change handler mirror the existing `liveFavoritesToggle` pattern.

## Why read from `raw_hex` and not the path hops

The hash size is a property of the whole packet and is present even for zero-hop packets (where there are no hops to inspect), so reading the path-length byte is correct in all cases. Unresolvable size is treated as single-byte (excluded when ON) — we only show packets we can positively confirm are multibyte.

## Performance (hot path)

The filter runs in the packet-render hot path, so: classification is **O(1) per packet group** — it reads the first resolvable observation's `raw_hex` (a short hex string, single `parseInt` of one byte) and short-circuits. No per-packet API calls, no allocation in the loop, no added O(n²). When the toggle is OFF (default) the check is a single boolean guard and does nothing else. The buffered-feed re-filter reuses the existing `rebuildFeedList` pass — no extra traversal.

## Tests

- **Unit** (`test-live-multibyte-filter.js`, 9 cases): single/2-byte/3-byte classification, transport-route offset, missing/short/garbage `raw_hex` → 0, whitespace tolerance.
- **E2E** (`test-live-multibyte-only-e2e.js`, Playwright): toggle present and defaults OFF; ON hides a single-byte packet while a multibyte one renders; OFF restores it; setting persists across reload. Registered in the CI live-E2E block in `deploy.yml`.

## Docs

User-guide entry added in `docs/user-guide/live.md`.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -410,6 +410,7 @@ jobs:
           BASE_URL=http://localhost:13581 node test-issue-1128-packets-layout-e2e.js 2>&1 | tee -a e2e-output.txt
           BASE_URL=http://localhost:13581 node test-issue-1128-multi-viewport-e2e.js 2>&1 | tee -a e2e-output.txt
           BASE_URL=http://localhost:13581 node test-issue-1136-live-region-e2e.js 2>&1 | tee -a e2e-output.txt
+          BASE_URL=http://localhost:13581 node test-live-multibyte-only-e2e.js 2>&1 | tee -a e2e-output.txt
           BASE_URL=http://localhost:13581 node test-issue-1150-404-state-e2e.js 2>&1 | tee -a e2e-output.txt
           BASE_URL=http://localhost:13581 node test-issue-1146-path-link-contrast-e2e.js 2>&1 | tee -a e2e-output.txt
           CHROMIUM_REQUIRE=1 node test-issue-1705-subpath-contrast-e2e.js 2>&1 | tee -a e2e-output.txt
--- a/docs/user-guide/live.md
+++ b/docs/user-guide/live.md
@@ -67,6 +67,7 @@ Each packet type has a color and icon:
 ## Controls
 
 - **Favorites only** — show only packets from your claimed nodes
+- **Multibyte only** — show only multibyte (≥2-byte path-hash) packets and hide single-byte traffic. Single-byte path hashes collide heavily at scale, so their paths are unreliable; enable this for a cleaner, more trustworthy live picture. Off by default; the choice is remembered.
 - **Matrix mode** — visual effect overlay (just for fun)
 
 ## Tips
--- a/public/hop-filter.js
+++ b/public/hop-filter.js
@@ -40,6 +40,23 @@
     return s.length >> 1;
   }
 
+  // Packet-level path-hash size (1|2|3), or 0 when unresolvable.
+  // Reads the path-length byte from raw_hex; its top two bits encode
+  // hashSize-1. Offset is 5 for transport routes (route_type 0/3, which
+  // carry 4 next/last-hop code bytes before path-length) else 1 — mirrors
+  // getPathLenOffset()/computeBreakdownRanges() in public/app.js. Reading
+  // from raw_hex (not path hops) is correct even for zero-hop packets.
+  function packetHashSize(rawHex, routeType) {
+    if (!rawHex || typeof rawHex !== 'string') return 0;
+    var clean = rawHex.replace(/\s+/g, '');
+    var bytes = clean.length >> 1;
+    var offset = (routeType === 0 || routeType === 3) ? 5 : 1;
+    if (bytes < offset + 1) return 0;
+    var pathByte = parseInt(clean.slice(offset * 2, offset * 2 + 2), 16);
+    if (isNaN(pathByte)) return 0;
+    return (pathByte >> 6) + 1;
+  }
+
   // Render-time predicate. opts may be omitted — if so, falls back to the
   // current localStorage value. The hop is hidden only when:
   //   - opts.hide1ByteHops === true AND
@@ -77,6 +94,7 @@
     window.MC_isVisibleHop = isVisibleHop;
     window.MC_filterPathHops = filterPathHops;
     window.MC_hopByteLen = hopByteLen;
+    window.MC_packetHashSize = packetHashSize;
   }
 
   if (typeof module !== 'undefined' && module.exports) {
@@ -86,6 +104,7 @@
       isVisibleHop: isVisibleHop,
       filterPathHops: filterPathHops,
       hopByteLen: hopByteLen,
+      packetHashSize: packetHashSize,
       _STORAGE_KEY: STORAGE_KEY
     };
   }
--- a/public/live.js
+++ b/public/live.js
@@ -39,6 +39,7 @@
   let showGhostHops = localStorage.getItem('live-ghost-hops') !== 'false';
   let realisticPropagation = localStorage.getItem('live-realistic-propagation') === 'true';
   let showOnlyFavorites = localStorage.getItem('live-favorites-only') === 'true';
+  let multibyteOnly = localStorage.getItem('live-multibyte-only') === 'true';
   let matrixMode = localStorage.getItem('live-matrix-mode') === 'true';
   let matrixRain = localStorage.getItem('live-matrix-rain') === 'true';
   let colorByHash = localStorage.getItem('meshcore-color-packets-by-hash') !== 'false';
@@ -1107,6 +1108,8 @@
             <span id="audioDesc" class="sr-only">Sonify packets — turn raw bytes into generative music</span>
             <label><input type="checkbox" id="liveFavoritesToggle" aria-describedby="favDesc"> ⭐ Favorites</label>
             <span id="favDesc" class="sr-only">Show only favorited and claimed nodes</span>
+            <label><input type="checkbox" id="liveMultibyteToggle" aria-describedby="multibyteDesc"> Multibyte only</label>
+            <span id="multibyteDesc" class="sr-only">Show only multibyte (≥2-byte path-hash) packets; hide unreliable single-byte traffic</span>
             <label id="liveGeoFilterLabel" style="display:none"><input type="checkbox" id="liveGeoFilterToggle"> Mesh live area</label>
             </div>
             <div class="live-toggles">
@@ -1571,6 +1574,14 @@
       showOnlyFavorites = e.target.checked;
       localStorage.setItem('live-favorites-only', showOnlyFavorites);
       applyFavoritesFilter();
+    });
+
+    const multibyteToggle = document.getElementById('liveMultibyteToggle');
+    multibyteToggle.checked = multibyteOnly;
+    multibyteToggle.addEventListener('change', (e) => {
+      multibyteOnly = e.target.checked;
+      localStorage.setItem('live-multibyte-only', multibyteOnly);
+      rebuildFeedList();
     });
 
     // Region filter (#1045): dropdown of observer IATA regions
@@ -2794,6 +2805,7 @@
     const sorted = [...byHash.values()].sort((a, b) => b.latestTs - a.latestTs).slice(0, 25);
 
     for (const group of sorted) {
+      if (multibyteOnly && !groupIsMultibyte(group.packets)) continue;
       const pkt = Object.assign({}, group.latestPkt, { observation_count: group.count });
       const decoded = pkt.decoded || {};
       const header = decoded.header || {};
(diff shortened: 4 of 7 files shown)


Kpa-clawbot (changes requested):

### Must-fix

1. **`public/live.js` — `groupIsMultibyte` uses `window.MC_packetHashSize` but unit test doesn't cover this integration path.** The function silently returns `false` when `MC_packetHashSize` is undefined (e.g., if `hop-filter.js` fails to load). This is defensive, but the E2E test only exercises the happy path where the script loaded. Consider: at minimum add one unit-test case for `groupIsMultibyte` with `MC_packetHashSize` absent → returns `false`. Currently `groupIsMultibyte` is untested in isolation (only exercised via E2E).

2. **`test-live-multibyte-only-e2e.js:87` — `makePkt` is defined at module scope but references `singleHash`/`multiHash` that only exist at call-site.** Actually this is fine (it takes `hash` as arg). However, the E2E calls `makePkt(singleHash, SINGLE_HEX)` inside `page.evaluate` (line 87-88). `makePkt` is a Node function — it isn't available inside the browser page context. This should throw `ReferenceError: makePkt is not defined` unless the test framework serializes it. Looking closer: the second arg to `page.evaluate` is `{ single: makePkt(...), multi: makePkt(...) }` — `makePkt` runs in Node and the result is serialized into the page. This is actually correct. Withdrawn.

3. **`test-live-multibyte-only-e2e.js:93-94` — `.catch(() => {})` swallows timeout without failing the test.** If the multibyte packet never renders (real bug), `waitForFunction` times out, the catch swallows it, then the next `evaluate` checks the DOM — which WILL fail the assertion. So functionally it works, but the error message will be the assertion failure ("multibyte packet should render...") rather than "timed out waiting for selector". This is a minor DX issue — the `.catch(() => {})` is a code smell that obscures the actual failure mode. Remove the `.catch(() => {})` or replace with `.catch(() => null)` and check its return, so timeout gives a clearer diagnostic.

4. **`test-live-multibyte-only-e2e.js:102` — same `.catch(() => {})` pattern on the OFF→single-appears step.** Same issue as #3.

5. **`public/live.js` — `rebuildFeedList` filter but no map/rain filter in the same toggle handler.** The PR description says "excluded from the entire live view — feed, map polylines/rain, and the packet counter" but the toggle change handler only calls `rebuildFeedList()`. The map/rain exclusion happens via `renderPacketTree` (which early-returns for non-multibyte). However, already-rendered polylines/rain from before the toggle was enabled remain on the map — `rebuildFeedList()` does not clear map layers. This means toggling ON hides new arrivals but does NOT remove already-painted polylines for single-byte packets from the current map view. The PR body claims full exclusion; the implementation only excludes going-forward. Either the claim should be softened in docs, or the toggle handler should also clear/re-render map layers.

6. **`docs/user-guide/live.md` — doc line is very long (single line, ~280 chars).** Not a functional issue, but the rest of the file wraps at ~80 chars. Wrap this entry for consistency with the surrounding prose style.

7. **`public/hop-filter.js` — `packetHashSize` doesn't validate `routeType` is a number.** If called with `routeType = undefined` or a string, the `=== 0` / `=== 3` checks will fail and it falls to offset 1. This is *probably fine* in practice (defensive default), but differs from how `hopByteLen` validates its inputs. A brief inline comment ("non-numeric routeType treated as non-transport") would clarify intent.

8. **`test-live-multibyte-filter.js` — no test for `routeType = undefined` or non-numeric.** Since `groupIsMultibyte` passes `p.route_type` which may be `undefined` for some packet shapes, this edge case should have coverage to document the fallback behavior.

### Out-of-scope

1. The existing "hide 1-byte path hops" toggle shares conceptual overlap with this filter — potential UX confusion. Would benefit from a tooltip distinguishing the two, but that's a separate enhancement.
2. Map-layer clearing on toggle (if deemed necessary) is a larger refactor touching map render state.

Kpa-clawbot (commented):

### TDD Verdict: EXEMPT-net-new-UI ✅

**Exemption applied:** Net-new UI surface. `MC_packetHashSize` and the "Multibyte only" toggle did not exist prior to this PR — there is no pre-existing behavior to break. Per AGENTS.md: "test must land in the SAME PR but doesn't need to be the FIRST commit."

**Would tests fail against pre-PR state?** Yes — `packetHashSize` did not exist in `hop-filter.js` on master. The unit test's `load()` would fail to find the export. The E2E's `#liveMultibyteToggle` selector would throw. Both tests gate the new behavior.

Commit 1 bundles production (`public/hop-filter.js`) + test (`test-live-multibyte-filter.js`) in one commit. Under strict red→green this would be a violation, but the net-new-UI exemption explicitly permits this.

---

### Six Questions — `test-live-multibyte-filter.js` (9 cases)

| # | Question | Assessment |
|---|----------|-----------|
| a | Fails on revert? | ✅ Function wouldn't exist; `load()` returns undefined → all 9 cases throw. |
| b | Smallest test? | ✅ Each case is a single `strictEqual` on one dimension (offset, bit-shift, edge). Minimal. |
| c | Wrong impl passes? | ✅ No — tests pin offset-1 vs offset-5 (transport diverges), all three bit-shift outputs (1/2/3), and the 0-return family (null/short/garbage). A hardcoded constant fails ≥6 cases. |
| d | Untested edges? | ⚠️ Mixed-case hex (`'10Ff'`) not tested — `parseInt` handles it, but the assumption is implicit. route_type 2/4 implicitly falls into the non-transport path (offset 1) — no explicit assertion. |
| e | Names? | ✅ Behavioral: "non-transport single-byte", "transport route uses offset 5", "garbage hex byte → 0". |
| f | Setup > test? | ✅ `load()` vm sandbox is 8 lines; each test is 1 assertion. Ratio is fine. |

### Six Questions — `test-live-multibyte-only-e2e.js` (4 steps)

| # | Question | Assessment |
|---|----------|-----------|
| a | Fails on revert? | ✅ `#liveMultibyteToggle` wouldn't exist → step 1 throws. `_liveBufferPacket` wouldn't call `groupIsMultibyte` → single-byte packet would render → step 2 fails. |
| b | Smallest test? | ✅ Integration toggle + render pipeline is the right granularity for E2E. |
| c | Wrong impl passes? | ✅ "Always hide all" fails (multibyte must render). "Never hide" fails (single-byte must be hidden). "Always show all" fails step 2. |
| d | Untested edges? | ⚠️ VCR/replay mode not exercised. Counter exclusion not asserted (PR body claims it). Map polyline/rain filtering not asserted. |
| e | Names? | ✅ Descriptive step strings. |
| f | Setup > test? | ✅ Playwright boilerplate is standard; assertions are clear. |

---

### Severity

**Must-fix:** 0

**Out-of-scope (not blocking):**
1. Mixed-case hex not unit-tested (low risk — `parseInt` is case-insensitive, but explicit coverage is better documentation).
2. E2E doesn't assert counter exclusion or map-polyline filtering — these are claims in the PR body that lack test evidence. Consider adding in a follow-up.
3. `route_type` values other than 0/1/3 have no explicit unit case (they fall into the else-branch with offset 1, same as route_type 1 — implicitly covered but not documented by test name).

---

### Final Verdict: **APPROVED**

Tests are well-designed, behaviorally named, anti-tautological, and would fail when the implementation is reverted. The net-new-UI exemption is cleanly justified. No must-fixes.

Kpa-clawbot (changes requested):
## Tufte Review (round 1) — PR #1781 "Multibyte only" live filter

### Summary

This PR adds one checkbox toggle to the live view controls bar. The implementation is clean — O(1) filter, mirrors existing `⭐ Favorites` pattern, proper localStorage persistence. The core data-ink concern: **does this control help the viewer think about the data, or does it add cognitive overhead?**

Verdict: the feature itself is sound (filtering unreliable single-byte collisions improves signal-to-noise), but the **label and description leak protocol internals** to end-users who don't know what "path-hash bytes" are.

---

### Must-fix

**1. Label jargon: "Multibyte only" is meaningless to non-protocol users**
`public/live.js:1110` — The label `Multibyte only` communicates nothing about *why* the user would want this. Compare the adjacent `⭐ Favorites` — immediately clear. A Tufte-aligned label communicates the *data consequence*, not the implementation mechanism.

Suggested: **"Reliable paths only"** or **"Hide noisy paths"** — these describe the *information quality* outcome, not the byte-width of a hash.

**2. aria-describedby text repeats the jargon**
`public/live.js:1111` — `"Show only multibyte (≥2-byte path-hash) packets; hide unreliable single-byte traffic"` — screen-reader users get protocol jargon *before* the useful part. Lead with the user benefit: "Hide packets with unreliable single-byte path hashes that collide at scale."

**3. User-guide doc line buries the "why" after the "what"**
`docs/user-guide/live.md:70` — The doc line front-loads "multibyte (≥2-byte path-hash)" — protocol detail — before explaining the user-facing consequence (collision unreliability). Invert: lead with the benefit, parenthetical the mechanism.

Suggested: `- **Reliable paths only** — hide single-byte path-hash packets whose paths collide heavily at scale; keeps only the trustworthy ≥2-byte traffic. Off by default; choice is remembered.`

---

### Out-of-scope

- The controls panel has ~10 toggles already visible (Heat, Inferred Hops, Realistic, Color by hash, Matrix, Rain, Audio, Favorites, Geo filter, Region). Adding one more is acceptable for now, but the panel is approaching the threshold where grouping/progressive disclosure would improve scanability. Not this PR's problem.
- No emoji/icon on the new toggle (Favorites has ⭐). Consider a small icon for visual scanability in a future pass — but not required here since most toggles lack one.

---

### CDP Evidence

Loaded `http://analyzer-stg.00id.net/#/live` in headless Chromium. Screenshot confirms the BEFORE state of the controls bar. All toggle labels extracted via `querySelectorAll('.live-toggles label')` — rects are 0×0 (panel is in a collapsed/overflow state on initial load with no packets flowing), so I **cannot validate spacing/overlap visually**. The new toggle's DOM position (between Favorites and Geo filter) is confirmed from the diff; no overlap concern given it follows the identical `<label><input type="checkbox">` pattern as its siblings.

---

### Data-ink assessment

- **Data-ink ratio impact**: +1 checkbox, +2 words visible. Acceptable — it's a filter control, not chartjunk. When ON, it *removes* low-quality data from the display, actually improving the data-ink ratio of the live feed itself.
- **No chartjunk introduced**: no decorative elements, no gratuitous animation, no tooltip hover-only information hiding.
- **Information hiding concern**: the toggle defaults OFF, so no data is hidden by default. Good. The user opts in to a higher-signal view. This is the correct default.

PR #1814, opened 2026-07-01
fix(release): report correct version on fast-path retagged images (#1807)

### LINKED ISSUE #1807: v3.9.2: /api/stats reports version: "edge" instead of v3.9.2 — UI hides version display
**Container tag:** ghcr.io/kpa-clawbot/corescope:v3.9.2
**Commit:** 3c440c0 (matches the v3.9.2 release commit)

## Problem

After pulling and running `v3.9.2`, `/api/stats` returns:

```json
"version":"edge","commit":"3c440c0","buildTime":"2026-06-13T04:44:00Z"
```

Since `version` is `"edge"` rather than `"v3.9.2"`, the Performance dashboard's
`renderVersionCard()` (public/perf.js) ends up linking to a broken
`/releases/tag/vedge` URL — or in practice the version display looks wrong/missing
to operators who expect to see `v3.9.2`.

## Likely cause

This looks related to the new release fast-path in #1680
("Release fast-path: re-tag :edge → :vX.Y.Z when SHA matches"). When the
:edge image is re-tagged to :vX.Y.Z without a full rebuild, the version
string baked into the binary via ldflags at build time isn't updated to
match the new tag — so the binary still reports itself as "edge" even
though the container is now labeled v3.9.2.

## Suggested fix

When the fast-path re-tags :edge → :vX.Y.Z, either:
- re-inject the correct version via ldflags and do a (fast) rebuild, or
- have the fast-path also patch/override the version string reported by
  /api/stats and /api/healthz to match the new tag.

## Repro

1. Pull ghcr.io/kpa-clawbot/corescope:v3.9.2
2. curl http://localhost:3000/api/stats
3. Observe "version":"edge" instead of "version":"3.9.2" / "v3.9.2"

Saarlandpower:
Fixes #1807.

Implements the fix path from the triage (env → image-version file → baked version, zero rebuild cost):

## Changes

**`cmd/server/main.go` — `resolveVersion()` fallback chain**
1. `CORESCOPE_VERSION` env (operator override)
2. `.image-version` file in the working dir (`/app` in the container) — mirrors the existing `.git-commit` pattern in `resolveCommit()`
3. ldflags-baked `Version`
4. `"unknown"`

Edge builds are unaffected: no env, no file → baked `"edge"` as before.

**`.github/workflows/release-fast-path.yml` — retag step**
Instead of a plain `crane tag :edge → :vX.Y.Z`, the fast path now runs `crane mutate` on `:edge` with:
- `--append` of a deterministic one-file layer containing `/app/.image-version` = `vX.Y.Z`
- `--label org.opencontainers.image.version=vX.Y.Z`
- `--tag :vX.Y.Z`

`vX.Y`, `vX` and `latest` are then pointed at the mutated image. Still no rebuild — the mutation is a manifest + single ~100-byte layer operation.

## Notes
- The release tags no longer share the exact digest with `:edge` (they carry one extra layer); the fallback SHA check is unaffected since it compares the `org.opencontainers.image.revision` label against `github.sha`.
- The layer tar uses `--owner=0 --group=0 --mtime='UTC 2020-01-01'` for reproducibility.
- Operators can also fix existing deployments immediately with `-e CORESCOPE_VERSION=v3.9.2`, no image change needed.

## Testing
- `go build ./cmd/server` + `go vet` clean (golang:1.24)
- Workflow YAML validated
- Reporter context: running the affected v3.9.2 fast-path image in production (live.saarmesh.de), happy to verify the next tagged release end-to-end.


Saarlandpower:
Review feedback addressed (commits `7eace7bd`, `d89ff2d2`):

1. **MAJOR — TestResolveVersion coverage:** Added `TestResolveVersionFallbackChain` covering all three branches of the new `resolveVersion()` fallback chain: (a) `CORESCOPE_VERSION` env override, (b) `.image-version` file, (c) baked `Version` fallback.
2. **Build fix:** Missing `"os"` import added to `coverage_test.go` — caught by running `go test ./...` locally on bold-wright before pushing. Both tests pass cleanly.

Local validation: `go test -run TestResolveVersion -v -timeout 60s .` → PASS on Go 1.23.4 / linux/amd64.

Saarlandpower:
**Browser validation (AGENTS.md Rule 2):** Not applicable for this change — pure Go backend fix in `cmd/server/main.go` and `coverage_test.go`. No frontend files touched, no UI behavior changed. The version string will appear correctly in `/api/stats` once the fast-path workflow sets the env/file; visual verification on `/perf` is deferred to the next tagged release.

Saarlandpower:
Bot review has been green for a while now — this looks ready whenever a maintainer has a moment to approve the CI run (action_required gate). Let me know if anything else is needed on my end. Thanks!

Kpa-clawbot (commented):
## Independent review (round 1) — parallel personas

Consolidated from carmack / munger / dijkstra / kent-beck on `gh pr diff` at head df96a8a. Verdict: **APPROVE (comment-only)** — fix is correct and minimal; findings below are all non-blocking polish notes for a follow-up.

### Findings

**MAJOR — [kent-beck] Post-hoc test (TDD)**
Commit order: b9f2909 lands `resolveVersion()` production code with no test; 7eace7b adds `TestResolveVersionFallbackChain` after bot review. Per repo `AGENTS.md` TDD rule (red→green with revert-failing assertion first) this is a violation. Revert-fail check is fine (removing the env branch fails case (a)), so the test does its job now — flagging for process, not blocking a community fix.

**MINOR — [munger] `"unknown"` sentinel asymmetry** (`cmd/server/main.go:55-58`)
`.image-version` file value `"unknown"` is filtered and falls through to baked `Version`, but `CORESCOPE_VERSION=unknown` is returned verbatim. Consider applying the same filter to the env branch for consistency with `resolveCommit()`'s pattern.

**MINOR — [kent-beck] Missing negative cases** (`cmd/server/coverage_test.go:1046-1075`)
No test covers: `.image-version` empty / whitespace-only / literal `"unknown"` → all should fall through. Case (b) only exercises the happy path; a fourth subcase would pin the `TrimSpace + != "unknown"` guards.

**MINOR — [dijkstra] Errors ignored** (`coverage_test.go:1062-1063,1072`)
`os.Getwd()` and both `os.Chdir()` returns unchecked. If `Getwd` errors, `origDir=""` and the deferred `Chdir("")` fails silently, potentially leaving other tests in `t.TempDir()`. Use `t.Chdir(tmp)` (Go 1.24+, already the toolchain here) — it auto-restores and eliminates the manual dance.

**MINOR — [kent-beck] Monolithic subtests**
Three cases in one function without `t.Run` — failures report as one line and case ordering matters. `t.Run("env wins", ...)` etc. would isolate state and improve failure surface.

**NIT** — trailing blank line at `coverage_test.go:4745`.

### CI / out-of-scope

`gh pr checks 1814` → `no checks reported` (release-fast-path.yml is `release`-triggered). Recommend `go test ./cmd/server -run TestResolveVersion` locally before merge, plus end-to-end verification on the next tagged release (reporter offered to do so). The YAML path cannot be exercised until then — accepted risk.

efiten:

efiten:

**Verdict: approve.** The mechanism is sound and I verified the part I doubted: `Dockerfile:62`
sets `WORKDIR /app`, `.git-commit` is written there (Dockerfile:73) and read back with the same
relative `os.ReadFile(".git-commit")` at `cmd/server/main.go:34`. So a `/app/.image-version`
layer plus `os.ReadFile(".image-version")` follows an established, working pattern.

Three notes, none blocking:

1. **The fast path's own premise is now stale in the same file it edits.** The header comment
   in `release-fast-path.yml` still says "Skips ~30 min ... because the bytes are identical —
   only the manifest name changes". After `crane mutate --append`, the release image is no
   longer bit-identical to `:edge`: it carries an extra layer and a different digest. The retag
   loop handles this correctly by re-tagging from `${IMAGE}:${TAG}`, but the comment should be
   updated with it.
2. The test calls `os.Chdir`, which is process-global. Go runs tests in a package sequentially
   unless `t.Parallel()` is used, so it works today, but it will break the first time anything
   in this package goes parallel.
3. A stray trailing blank line is appended to the end of `coverage_test.go`.

Issue #1598, opened 2026-06-05
Active nodes are disappearing from the map


liquidraver:
After the latest release, active nodes are disappearing from the live view and the map too.
After adverting, they come back, but after a while (more than half a day?) they completely disappear and only appear as ghosts in paths. (there is constant traffic through the nodes all day)
We have adverts set to 47 hours.
Corescope is set to retain 10 days of data.

Kpa-clawbot:
**Severity:** P2 — active relay-only nodes drop off the map after ~24h despite forwarding traffic.

**Claim:** likely confirmed — but two distinct mechanisms could cause it; need a small clarification to pick.

**Code today:**
- `cmd/server/store.go:1410-1434` (`touchRelayLastSeen`) updates `last_seen` for resolved relay pubkeys, but is debounced to 5 min and only runs when `resolvedPubkeys` is non-empty (i.e., disambiguator resolved that hop). 1-byte hash collisions or new repeaters with no advert yet produce empty resolved sets → no touch.
- `public/roles.js:354-360` (`getNodeStatus`): a node is `stale` when `Date.now() - last_seen > nodeSilentMs (24h)` or `infraSilentMs (72h)` for repeaters/rooms. With adverts at 47h, a non-repeater node stops advertising for >24h ⇒ map flips it to stale even when relayed by others.
- `public/map.js:1520-1525`: `statusFilter='active'` (default for many users via localStorage) actively **hides** stale markers — they "disappear" rather than just dim.

**Findings:**
- (meshcore): adverts every 47h is longer than the `nodeSilentMs=24h` companion threshold (`public/roles.js:341`). Companions only re-advertise on user-initiated app open; for a 47h advert cadence, a healthy companion *will* fall into stale and (if filter=Active) vanish. This is a threshold-vs-reality mismatch, not a regression.
- (carmack): the recent #708 fix (`indexByNode` from `resolved_path`) only covers in-memory liveness; the persisted `nodes.last_seen` only moves via ADVERTs or `touchRelayLastSeen`. Relay-only repeaters get touched; non-repeater nodes that only show as packet sources without re-advertising do not.

**Fix path:** two layers — (a) make `HEALTH_THRESHOLDS.nodeSilentMs` config-derived so deployments with long advert intervals can raise it; (b) consider touching `last_seen` on any observed `srcPubKey` (not just resolved path) to capture activity. `public/roles.js:337` + `cmd/server/store.go:1410`. Effort: **M**.

[...]

PR #1815, opened 2026-07-01
fix(frontend): relay-aware staleness for infra nodes + dim-not-delete (#1598, PR A)

Saarlandpower:
Implements **PR A** from the #1598 triage fix path (r6 — all three operator decisions locked: confidence ≥0.75 [PR B], **dim-not-delete**, **in-place `getNodeStatus` signature extension**).

## Changes

**`public/roles.js` — `getNodeStatus()`**
- Now accepts a full node object (preferred); the legacy `(role, lastSeenMs)` signature keeps working unchanged.
- For infra roles (repeater/room), freshness = `max(advert-based timestamp, last_relayed)`. Freshness precedence mirrors existing call sites: `_liveSeen` > `_lastHeard` > `last_heard` > `last_seen`.
- `last_relayed` is only consulted for infra — companions keep pure advert/heard-based staleness (per @liquidraver's collision caveat; the ≥0.75-confidence `_liveSeen` refresh is PR B).

**`public/live.js` — `pruneStaleNodes()`**
- Repeater/room markers are **dimmed, never deleted**, regardless of `_fromAPI` origin. WS-only non-infra nodes are still removed to prevent unbounded memory growth.

**Call sites** — all seven (`nodes.js` ×3, `map.js` ×3, `live.js` ×1) now pass the node object. The Nodes-page status explanation shows "Last relayed …" when relay participation is the fresher signal, so an Active badge next to an old "last heard" isn't confusing.

**Tests** — 16 new `getNodeStatus` unit tests incl. the triage's backbone-repeater fixture (`last_seen`=25h, `last_relayed`=5min → `active`); `pruneStaleNodes` tests updated for dim-not-delete plus a new relay-aware prune test.

## Test results
`node test-frontend-helpers.js`: **641 passed, 2 failed** — the 2 failures (favStar ★ assertions) are pre-existing on current master (verified on a clean upstream clone).

## Validation offer
live.saarmesh.de currently has **160 infra nodes past `infraSilentMs`=72h while actively relaying** (incl. KatS disaster-relief repeaters) — a ready-made test population. Happy to run this branch there and report before/after node visibility.

Refs #1598 (PR A of two; PR B = `_liveSeen` refresh on `resolved_path` ≥0.75 confidence).


efiten:

efiten:

**Verdict: approve the staleness half. One question on the pruning half.**

The premise checks out. `last_relayed` is a real, populated field, not something the PR assumes
into existence: it is set at `cmd/server/routes.go:1413` and `:1610`, documented in
`openapi.go:191`, computed by `relayMetrics` at `store.go:4218`, and already consumed by
`analytics.js:2518` and `nodes.js:643`. So making `getNodeStatus` consider it for infra roles is
building on something that works. Note this is independent of the #1855 problem: that one is about
`nodes.last_seen`, this uses `last_relayed`, which is computed in memory and was never affected.

The dual signature (`getNodeStatus(node)` preferred, `getNodeStatus(role, ms)` still accepted) keeps
every existing caller working, and fifteen tests cover both forms including the "companion does not
get relay-based freshness" case. The odd-looking `if (!(effectiveMs >= relayedMs))` is deliberate,
not a bug: the negated form assigns when `effectiveMs` is NaN, which a plain `<` would not.

**The question is the pruning change, which is a bigger behaviour change than the title suggests.**
`pruneStaleNodes` now exempts every infra node from deletion, not just API-loaded ones. The
existing comment says pruning exists "to prevent unbounded memory growth", and the PR's own
replacement comment concedes it by narrowing to "remaining WS-only **non-infra** nodes are
removed". On a large mesh, a long-lived Live tab accumulates one permanently-retained entry per
repeater or room ever seen via ADVERT, with nothing bounding it.

Dimming rather than deleting is the right call for the user-visible behaviour, so ask for a bound
rather than a revert: evict infra after a much longer window than the staleness threshold, or cap
the retained marker count.

Issue #1558, opened 2026-06-04
Per-node stats (packetsToday, totalTransmissions, observers) reset on container restart — byNode index appears not populated by background backfill


MarekWo:
## Summary

After a container restart, per-node statistics shown on the **Home** page (`Packets today`, `Observers`, `Avg hops`, `SNR`) drop dramatically and slowly recover as new packets arrive. The in-memory store reports successful backfill of all packets, but the `byNode` index — which `/api/nodes/{pubkey}/health` and `/api/nodes` use to compute per-node stats — seems to only contain packets from the **hot startup window** + packets received **since restart**, not the full `retentionHours` window.

## Environment

- **Version:** `v3.8.2` (commit `93b2f4b`, build `2026-05-29T03:08:02Z`)
- **Image:** `ghcr.io/kpa-clawbot/corescope:latest`
- **Config:**
  ```json
  "packetStore": {
    "maxMemoryMB": 1024,
    "retentionHours": 168,
    "hotStartupHours": 24
  },
  "retention": {
    "nodeDays": 30,
    "observerDays": 30,
    "packetDays": 60
  }
  ```
- 3 MQTT sources, ~10k packets/h, ~210k packets/24h, 1700+ nodes, 68 observers, ~25k transmissions in DB.

## Reproduction

1. Run CoreScope for ~24h+ so per-node stats accumulate.
2. Open the **Home** page; observe a tracked node, e.g. `Packets today: 753`, `Observers: 23`, `Avg hops: 18.0`.
3. `docker compose restart corescope`.
4. Wait for healthcheck to pass and `[store] background load complete` in logs.
5. Hard-refresh the Home page (Ctrl+F5) to clear browser cache + WebSocket re-init.
6. **Expected:** stats roughly the same (`Packets today` near 753, observers near 23).
7. **Actual:** stats drop drastically — `Packets today: 8`, `Observers: 9`, `Avg hops: 8.0`. They slowly recover as new MQTT packets arrive, but never reach the pre-restart numbers until enough time passes for the in-memory store to re-accumulate from live ingest.

## What the logs say (NOT a backfill failure)

The in-memory store successfully loads everything from the DB after restart:

```
[ingestor] Loaded 296 channel keys from /app/channel-rainbow.json
[ingestor] Loaded 318 channel keys for GRP_TXT decryption
[...]

PR #1820, opened 2026-07-04
fix(clock-skew): restrict per-node skew to self-originated adverts (#1816, #1818)

### LINKED ISSUE #1816: Clock-skew engine attributes relayed adverts to relay-hop nodes: false no_clock/bimodal fleet-wide, bit-identical skew clusters
# Clock-skew engine attributes relayed adverts to relay-hop nodes — fleet-wide false `no_clock`/`bimodal_clock` classifications and bit-identical skew "clusters"

## Summary

`getNodeClockSkewLocked` (`cmd/server/clock_skew.go`, `txs := s.byNode[pubkey]`) treats `byNode[pubkey]` as "ADVERT transmissions **originated by** this node". It is not that. `byNode` is deliberately an *involvement* index: it also contains every transmission where the node was resolved as a **relay hop** from the flood path (added for the activity timeline in #1558 / #1352), plus `destPubKey`/`srcPubKey` matches.

Consequence: a node's clock-skew sample stream contains adverts it merely **forwarded**. Each such sample carries the *originator's* clock error, charged to the relay. One broken-clock node's advert, flooded through N relays, injects the identical bad skew sample into all N relays' streams simultaneously.

This is not cosmetic. On the production instance (map.meshcore.hu, snapshot 2026-07-02 16:51 UTC, 936 nodes / 725 repeaters) it produces:

1. **Bit-identical "wrong clock clusters"**: groups of 5–15 unrelated repeaters showing the exact same `recentMedianSkewSec` to the decimal (e.g. 12 repeaters at exactly +30040.0, 8 at exactly −60791177.1, 6 at exactly −67220194.8). Operators reading the dashboard conclude whole regions share a synchronized wrong clock. In every case investigated, the "cluster" is one wrong originator plus its flood relays.
2. **False severity ratings at scale**: 274 of 725 repeaters are rated `no_clock`. At least **91** of those share a bit-exact `recentMedianSkewSec` value with ≥2 other nodes — provably the artifact (that is a *lower bound*: pollution with ≥2 distinct foreign samples in the window produces a non-shared median and is not counted by this heuristic). The 75 `bimodal_clock` repeaters are largely healthy relays that regularly forward adverts from one broken neighbor — the "flaky RTC / mixed good+bad samples" interpretation (#845) is wrong for most of them.
3. **Polluted derived statistics**: `goodFraction`, `recentBadSampleCount`/`recentBadSamples` (#1094), drift estimation inputs, RTC-reset badge counts, and any fleet-level analytics built on per-node skew (good-clock fraction per neighborhood, etc.).

The per-transmission math is correct — the bug is purely the join from transmissions to nodes.

## Confirmed end-to-end example (production data, public API, reproducible)

Fleet snapshot `GET https://map.meshcore.hu/api/nodes/clock-skew` on 2026-07-02 showed 12 repeaters spanning Vienna, Lower Austria, Styria, and Slovakia — different owners, different hardware — with:

- identical `recentMedianSkewSec = 30040.0`
- identical `lastAdvertTS = 1783039071` (2026-07-03 00:37:51 UTC — ~8h in the *future* at observation time)
- identical `lastObservedTS = 1783009052`

Twelve distinct physical nodes cannot advertise in the same second and be observed in the same second. Per-node evidence (`GET /api/nodes/{pubkey}/clock-skew` → `recentHashEvidence`) for two of them (`432178f3…` AT-W-XTZ-Huttengasse and `1686e201…` SK-ZV-Slatina) shows the +30040 value comes from **the same transmission hash** in both: `6bf340e548188992…`.

`GET /api/packets/6bf340e548188992` decodes that packet as:

```json
{"type":"ADVERT",
 "pubKey":"51003e7d1d1a92182a9bb2299acae27d236fb196a2ab45808e0a557d098b69a5",
 "timestamp":1783039071, ...}
```

with a 17-hop `_parsedPath`. The originator is `51003e7d…` (AT-WB-NPStuberl-RIT), whose clock genuinely is ~+8.3 h fast (its own fleet entry: `recentMedian=30040, drift=0, goodFraction=0`, 266 samples). The other 11 "cluster members" are the resolved relay hops of that one flood. Their own sample streams are healthy (~−20 s) between foreign spikes; e.g. SK-ZV-Slatina's recent samples: `-16.6, -65565999.8, 100756, -23.8, -24.3, -30.8, -19.8, -63403679.8, 100584, 30040` — a healthy clock interleaved with the skews of every broken node whose adverts it relays.

Why bit-identical across all 12: severity uses the recent window = min(last `recentSkewWindowCount=5` samples, samples within `recentSkewWindowSec=3600`). A relayed advert that is the only sample in the last hour becomes the window's sole member for **every relay on its path**, so `recentMedianSkewSec` equals that one foreign value, bit-exactly, for all of them.

Every other investigated "cluster" resolved the same way to a single wrong originator:

| Apparent cluster | Actual originator (own-advert verified) | Originator's real clock state |
|---|---|---|
| 12 nodes at exactly +30040.0 | AT-WB-NPStuberl-RIT (`51003e7d…`) | +8.3 h fast |
| 12 nodes at −76739.3/−76739.8 (two values, two relay sets) | HU-PE-5AVZ-MOBIL (`1915bccb…`), two adverts sent 2 min apart | −21.3 h slow |
| 8 nodes at exactly −60791177.1 | SK-SC-Most-1EPT relayed advert, advertTS 2024-07-29 | firmware default epoch (2024-05-15) + 75 d uptime |
| 6 nodes at exactly −67220194.8 | Rick-T-Deck (`962568a3…`), advertTS 2024-05-15 | never-set clock at MeshCore's `VolatileRTCClock` default epoch 1715770351 |
| 5 nodes at exactly −66591266.8 | AT-W02-MASH (`23490db1…`), advertTS 2024-05-22 | rebooted to default epoch + 7 d, never re-synced (lifetime median was −32 s) |

Also affected at band level (not bit-identical, but same mechanism): ~54 nodes shown in a +100.5k band and ~17 in a +201k band reduce to ~6 and ~5 genuinely wrong originators respectively; the rest are their relays.

## Root cause — code walk

All references are at commit `750b8742`.

1. **`byNode` is an involvement index by design.** It is populated from three source classes:
   - `indexByNode` (`cmd/server/store.go:1711-1731`): decoded JSON fields `pubKey`, `destPubKey`, `srcPubKey`.
   - `indexResolvedPathHops` (`store.go:1696-1705`) → `addToByNode` for every **resolved relay-hop pubkey** — live ingest, `Load`, `loadChunk` (`store.go:952`, `store.go:1411`, `chunked_load.go:615`; contract documented at `store.go:1680-1695`, issue #1558).
   
   This is intentional and correct for its consumers ("byNode is what the node-analytics activity timeline reads", `store.go:946-947`). Do **not** change `byNode` itself.

2. **The clock-skew engine misuses it.** `getNodeClockSkewLocked` (`clock_skew.go:489` ff.):

   ```go
   txs := s.byNode[pubkey]
   ...
   for _, tx := range txs {
       if tx.PayloadType == nil || *tx.PayloadType != PayloadADVERT {
           continue
       }
       cs, ok := s.clockSkew.nodeSkew[tx.Hash]
       ...
   ```

   The only filter is payload type. An ADVERT relayed by (or path-resolved to) `pubkey` passes, and its per-hash skew (which measures the **originator's** clock vs observers) is appended to `pubkey`'s `allSkews`/`tsSkews`. Everything downstream — recent window, `goodFraction`, bimodal/no_clock classification (`clock_skew.go:582-645`), drift input, sparkline `Samples`, `recentBadSamples`, `RecentHashEvidence` — inherits the pollution. The same misuse exists in the evidence-hash loop later in the function (`for _, tx := range txs` at the `RecentHashEvidence` build, ~`clock_skew.go:670`).

3. **Phases 1–2 are NOT affected.** `collectSamples`/`calibrateObservers`/`computeNodeSkew` operate per transmission hash and never touch `byNode`; per-hash skews and observer calibration are correct. The bug is exclusively the hash→pubkey join in `getNodeClockSkewLocked`.

4. **A secondary effect hides the evidence.** `hashEvidenceMedian` filters |skew| > 24 h and returns the median of the remainder (0 if none survive) — so for relayed epoch-reset adverts the evidence panel shows `medianCorrectedSkewSec: 0` while the raw per-observer skews inside the same entry read −66.5M s. This made the artifact harder to diagnose from the UI and API; the per-observer `rawSkewSec` values are the ground truth.

## Prior issues built on polluted data

#789 (recent-window severity), #845 (bimodal detection), #1094 (per-bad-sample evidence), #1285 (RTC-reset outlier exclusion) all added machinery to cope with "nodes that mix good and nonsense timestamps". A large fraction of that mixing is this attribution bug, not node behavior. The features remain useful (genuine reboot-to-epoch nodes exist), but their calibration and the fleet-level counts they produce should be re-validated after the fix.

## Suggested fix

At the top of `getNodeClockSkewLocked`, restrict to **self-originated** adverts: an ADVERT's decoded `pubKey` identifies its signer/originator, so keep a tx only if `decoded.pubKey == pubkey` (case per stored convention — pubkeys appear lowercase-hex in the store).

Sketch:

```go
txs := s.byNode[pubkey]
...
for _, tx := range txs {
    if tx.PayloadType == nil || *tx.PayloadType != PayloadADVERT {
        continue
    }
    if !txOriginatedBy(tx, pubkey) {   // NEW
        continue
    }
    ...
}

// txOriginatedBy reports whether the decoded payload's pubKey field equals
// the given node pubkey. ADVERT payloads are self-signed, so pubKey is the
// originator; relay-hop membership in byNode must not contribute skew.
func txOriginatedBy(tx *StoreTx, pubkey string) bool {
    decoded := tx.ParsedDecoded()   // cached parse
    if decoded == nil {
        return false
    }
    v, _ := decoded["pubKey"].(string)
    return strings.EqualFold(v, pubkey)
}
```

Apply the same guard to the second `txs` loop that builds `RecentHashEvidence`, or hoist the filtered slice and reuse it. Notes:

- `ParsedDecoded()` is already called for every advert in `collectSamples`, so the parse is cached; if profiling still shows cost on the fleet path (`computeFleetClockSkewForArea` calls `getNodeClockSkewLocked` per node), an alternative is a dedicated `advertsByOriginator map[string][]*StoreTx` maintained in `indexByNode` where the field is `pubKey` and payload type is ADVERT — that also removes the O(all-involved-tx) scan for relay-heavy hubs.
- Leave `byNode` untouched — activity timeline, home-page stats, and paths-through depend on relay attribution (#1558, #1352).
- Consider whether GetFleetClockSkew's steady-state recomputer snapshot (#1265) needs invalidation on deploy so stale polluted results don't persist until the next recompute.
- After deploying, fleet severity counts will shift substantially (expect `no_clock`/`bimodal_clock` repeater counts to drop on the order of one-third or more); if anything alerts on those counts, adjust expectations.

## Regression tests

1. **Relay pollution excluded**: build a store with one ADVERT tx (decoded `pubKey = "AAAA…"`, timestamp 8 h ahead of its observation) indexed under `"AAAA…"` via `indexByNode` AND under `"BBBB…"` via `addToByNode` (simulating relay-hop resolution, as `resolved_index_test.go` does). Assert `GetNodeClockSkew("BBBB…")` either returns nil (no self-originated adverts) or excludes the foreign sample; assert `GetNodeClockSkew("AAAA…")` still reports the +8 h skew.
2. **Bit-identical cluster reproduction (pre-fix failing test)**: one bad-clock originator + 5 relays, each relay also having several healthy self-adverts older than 1 h; pre-fix all 5 relays report `recentMedianSkewSec` equal to the originator's skew; post-fix they report their own healthy values.
3. **Evidence loop**: assert `RecentHashEvidence` for a relay contains only hashes of its own adverts.

## How to verify against production after the fix

- `GET /api/nodes/clock-skew`: group by exact `recentMedianSkewSec` for |skew| > 1 h — the ≥3-node bit-identical groups (91+ nodes in the 2026-07-02 snapshot) should disappear entirely; each genuine bad clock keeps a unique value.
- Spot-check `432178f3061bc59ac9acbc32e7201a1bb6e5bd8f12864ca62bade97ae68352d3` (AT-W-XTZ-Huttengasse) and `1686e2013f03774addbbe5c10e013024e24690d9b88d1eebe45b30f15e01938e` (SK-ZV-Slatina): both should flip to `ok` with recent medians near −20 s.
- `51003e7d1d1a…` (AT-WB-NPStuberl-RIT) should remain genuinely bad (~+30k s) — the fix must not mask real offenders.

### LINKED ISSUE #1818: Bug: Observer node in an island of "me, myself, and I" reports clock is off by nearly ten days.
Observer node is https://corescope.chicagolandmesh.org/#/observers/55C5891FDA8775E941A40B26657B87EB1D4F17195CFC7D40A486EBC6BEE98745

It is running the ESP32 native MQTT firmware, which syncs clock via PTP.  The self adverts it sends (visible with `direction: tx`) have correct time.  But for some reason, CoreScope perpetually thinks its clock is wildly off.

Saarlandpower:
Closes #1816. Closes #1818 (confirmed duplicate of #1816 by the triage bot).

## Root cause

`byNode` is an involvement index (`indexResolvedPathHops`, `store.go:1696-1705`, #1558/#1352): a transmission is indexed under every relay-hop pubkey found in an observation's `resolved_path`, not just its originator. `getNodeClockSkewLocked` (`clock_skew.go:489`) iterated every ADVERT transaction under a pubkey without checking who actually signed it, so a relay inherited the clock skew of every broken-clock node it forwarded as if it were its own.

This produced:
- Fleet-wide false `no_clock`/`bimodal_clock` classifications on healthy relays whose only "bad" samples were adverts they merely relayed.
- Bit-identical `RecentMedianSkewSec` "clusters" across unrelated relays that all forwarded the same broken-clock originator.
- Single relays showing a multi-day skew even though their own self-adverts are healthy, because 1-2 relayed adverts from a broken originator landed in the tail of their small recent-window sample (the #1818 "island" repro from @cwichura).

## Fix

Add `txOriginatedBy(tx, pubkey)`: ADVERTs are self-signed, so `decoded["pubKey"]` is the originator per protocol (case-insensitive compare as a defensive measure). Apply it as a guard in both the main skew-aggregation loop and the per-hash evidence loop in `getNodeClockSkewLocked`. `byNode` itself is untouched — #1558/#1352 still rely on the broader involvement index for other consumers.

## Tests

- Existing `clock_skew_test.go` / `clock_skew_issue1094_test.go` / `clock_skew_issue1285_test.go` fixtures built synthetic ADVERT transactions without a `pubKey` field and seeded `s.byNode` directly, bypassing the normal `indexByNode` path where every real ADVERT carries `pubKey`. Added `pubKey` to each fixture so it reflects a self-originated advert, which is what these tests already intended to represent. All pre-existing tests pass unchanged in behavior.
- New `clock_skew_issue1816_test.go`:
  - `TestTxOriginatedBy` — unit coverage of the new guard (self, foreign, missing pubKey, case-insensitivity).
  - `TestIssue1816_RelayDoesNotInheritOriginatorSkew` — a relay with healthy self-adverts plus relayed adverts from a broken-clock originator (matching the report's +100.5k s band) must report `ok` severity based only on its own adverts.
  - `TestIssue1816_PureRelaysReportNoSkew_NoBitIdenticalCluster` — five relay pubkeys that only ever forward a broken originator's advert (never self-advert) must report `nil`, not a bit-identical copy of the originator's skew.
  - `TestIssue1818_TwoForeignAdvertsDoNotPoisonIslandNode` — reproduces the cwichura island scenario: 8 healthy self-adverts + 2 foreign adverts at ~10 days skew must not flip severity or pollute `RecentMedianSkewSec`.

Full suite: `go test ./...` passes (one pre-existing, unrelated flaky test — `TestHandleNodePaths_PrefixCollision_1352`, an index-loading race — reproduces intermittently on unmodified `master` too).

Operator context: running CoreScope for SaarMesh (SaarLorLux, DE/FR/LU, 800+ nodes); this bug was surfacing as fleet-wide clock-skew false positives on our infra nodes.

### DIFF
--- a/cmd/server/clock_skew.go
+++ b/cmd/server/clock_skew.go
@@ -3,6 +3,7 @@ package main
 import (
 	"math"
 	"sort"
+	"strings"
 	"sync"
 	"time"
 )
@@ -484,6 +485,34 @@ func (s *PacketStore) GetNodeClockSkew(pubkey string) *NodeClockSkew {
 	return s.getNodeClockSkewLocked(pubkey)
 }
 
+// txOriginatedBy reports whether tx is an ADVERT self-signed by pubkey,
+// as opposed to a transmission merely indexed under pubkey because it was
+// resolved as a relay hop. byNode is an involvement index that
+// intentionally includes relay-hop transmissions for the activity
+// timeline (store.go:1696-1705, indexResolvedPathHops, #1558/#1352).
+// Clock-skew computation must restrict to self-originated adverts only:
+// ADVERTs are self-signed, so decoded["pubKey"] is the originator per
+// protocol (safe key, does not need path-hop resolution). Without this
+// guard, every relay that forwards a broken-clock node's advert inherits
+// that node's skew, producing fleet-wide false no_clock/bimodal
+// classifications and bit-identical skew "clusters" across unrelated
+// relays (#1816), including cases where a node's own adverts are healthy
+// but a couple of relayed adverts from a broken-clock originator dominate
+// its small recent-window sample (#1818). Comparison is case-insensitive
+// as a defensive measure; pubkeys are lowercase hex today (decoder.go),
+// but nothing enforces that at this boundary.
+func txOriginatedBy(tx *StoreTx, pubkey string) bool {
+	decoded := tx.ParsedDecoded()
+	if decoded == nil {
+		return false
+	}
+	pk, ok := decoded["pubKey"].(string)
+	if !ok || pk == "" {
+		return false
+	}
+	return strings.EqualFold(pk, pubkey)
+}
+
 // getNodeClockSkewLocked returns clock skew for a node.
 // Must be called with s.mu held (at least RLock).
 func (s *PacketStore) getNodeClockSkewLocked(pubkey string) *NodeClockSkew {
@@ -506,6 +535,12 @@ func (s *PacketStore) getNodeClockSkewLocked(pubkey string) *NodeClockSkew {
 
 	for _, tx := range txs {
 		if tx.PayloadType == nil || *tx.PayloadType != PayloadADVERT {
+			continue
+		}
+		if !txOriginatedBy(tx, pubkey) {
+			// Skip adverts this node merely relayed (byNode is an
+			// involvement index, not an originator index) — see
+			// txOriginatedBy and #1816/#1818.
 			continue
 		}
 		cs, ok := s.clockSkew.nodeSkew[tx.Hash]
@@ -670,6 +705,11 @@ func (s *PacketStore) getNodeClockSkewLocked(pubkey string) *NodeClockSkew {
 	var evidenceHashes []hashMeta
 	for _, tx := range txs {
 		if tx.PayloadType == nil || *tx.PayloadType != PayloadADVERT {
+			continue
+		}
+		if !txOriginatedBy(tx, pubkey) {
+			// Keep evidence consistent with the self-only skew stream
+			// above — don't show relayed adverts as this node's evidence.
 			continue
 		}
 		ev, ok := s.clockSkew.hashEvidence[tx.Hash]
--- a/cmd/server/clock_skew_issue1094_test.go
+++ b/cmd/server/clock_skew_issue1094_test.go
@@ -36,7 +36,7 @@ func seedIssue1094Repro(t *testing.T) (*PacketStore, []string, []int64) {
 		txs = append(txs, &StoreTx{
 			Hash:        "healthy-1094-" + formatInt64(int64(i)),
 			PayloadType: &pt,
-			DecodedJSON: `{"payload":{"timestamp":` + formatInt64(advTS) + `}}`,
+			DecodedJSON: `{"payload":{"timestamp":` + formatInt64(advTS) + `},"pubKey":"BADTS1094"}`,
 			Observations: []*StoreObs{
 				{ObserverID: "obs1", Timestamp: time.Unix(obsTS, 0).UTC().Format(time.RFC3339)},
 			},
@@ -52,7 +52,7 @@ func seedIssue1094Repro(t *testing.T) (*PacketStore, []string, []int64) {
 		txs = append(txs, &StoreTx{
 			Hash:        hash,
 			PayloadType: &pt,
-			DecodedJSON: `{"payload":{"timestamp":` + formatInt64(advTS) + `}}`,
+			DecodedJSON: `{"payload":{"timestamp":` + formatInt64(advTS) + `},"pubKey":"BADTS1094"}`,
 			Observations: []*StoreObs{
 				{ObserverID: "obs1", Timestamp: time.Unix(obsTS, 0).UTC().Format(time.RFC3339)},
 			},
--- a/cmd/server/clock_skew_issue1285_test.go
+++ b/cmd/server/clock_skew_issue1285_test.go
@@ -41,7 +41,7 @@ func seedIssue1285Repro(t *testing.T) *PacketStore {
 		tx := &StoreTx{
 			Hash:        "healthy-" + formatInt64(int64(i)),
 			PayloadType: &pt,
-			DecodedJSON: `{"payload":{"timestamp":` + formatInt64(advTS) + `}}`,
+			DecodedJSON: `{"payload":{"timestamp":` + formatInt64(advTS) + `},"pubKey":"RTCRESET"}`,
 			Observations: []*StoreObs{
 				{ObserverID: "obs1", Timestamp: time.Unix(obsTS, 0).UTC().Format(time.RFC3339)},
 			},
@@ -56,7 +56,7 @@ func seedIssue1285Repro(t *testing.T) *PacketStore {
 	rtcTx := &StoreTx{
 		Hash:        "rtc-reset-0001",
 		PayloadType: &pt,
-		DecodedJSON: `{"payload":{"timestamp":` + formatInt64(rtcResetAdv) + `}}`,
+		DecodedJSON: `{"payload":{"timestamp":` + formatInt64(rtcResetAdv) + `},"pubKey":"RTCRESET"}`,
 		Observations: []*StoreObs{
 			{ObserverID: "obs1", Timestamp: time.Unix(rtcResetObs, 0).UTC().Format(time.RFC3339)},
 		},
--- a/cmd/server/clock_skew_issue1816_test.go
+++ b/cmd/server/clock_skew_issue1816_test.go
@@ -1,0 +1,287 @@
+package main
+
+// Regression tests for #1816 / #1818:
+//
+// getNodeClockSkewLocked() iterated every ADVERT transaction indexed under
+// a pubkey in s.byNode WITHOUT checking whether the node actually
+// originated (self-signed) that advert. byNode is an involvement index
+// (store.go:1696-1705, indexResolvedPathHops, #1558/#1352): a transmission
+// is also indexed under every relay-hop pubkey extracted from an
+// observation's resolved_path. So a relay that merely forwarded a
+// broken-clock node's advert inherited that node's skew as if it were its
+// own — producing:
+//
+//   - Fleet-wide false no_clock / bimodal_clock classifications on healthy
+//     relays whose only "bad" samples are adverts they relayed (#1816).
+//   - Bit-identical RecentMedianSkewSec "clusters" across unrelated relays
+//     that all forwarded the same broken-clock originator (#1816).
+//   - A single relay showing a nonsense multi-day skew even though its own
+//     self-adverts are perfectly healthy, because 1-2 relayed adverts from
+//     a broken originator landed in the tail of its small recent-window
+//     sample (#1818 — the "island" repro from cwichura).
+//
+// The fix (txOriginatedBy in clock_skew.go) restricts the ADVERT filter in
+// getNodeClockSkewLocked to self-originated adverts only (decoded pubKey
+// == pubkey), leaving byNode itself untouched (#1558/#1352 still depend on
+// the broader involvement index for other consumers).
+
+import (
+	"testing"
+	"time"
+)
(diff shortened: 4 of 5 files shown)


efiten:

efiten:

**Verdict: approve. A real, well-diagnosed bug with fleet-wide blast radius.**

`byNode` is an *involvement* index: `indexResolvedPathHops` (#1558/#1352) also indexes a
transmission under every relay-hop pubkey in the resolved path. `getNodeClockSkewLocked` iterated
every ADVERT in `byNode[pubkey]` without checking who signed it, so **every relay that forwarded a
broken-clock node's advert inherited that node's skew.** That explains all three reported symptoms:
false `no_clock`/`bimodal` on healthy relays, bit-identical skew "clusters" across unrelated
relays that forwarded the same originator, and a single relay showing multi-day skew because one
or two relayed adverts dominated its small recent-window sample.

The fix is correctly scoped: `txOriginatedBy` gates the two ADVERT loops on
`decoded["pubKey"] == pubkey`, and `byNode` itself is left alone because other consumers depend on
the broader index. Using ADVERT's self-signed pubKey as the originator is right per protocol and
needs no path-hop resolution. `strings.EqualFold` is a documented defensive choice.

The most telling evidence is in the diff itself: two pre-existing test fixtures
(`clock_skew_issue1094_test.go`, `clock_skew_issue1285_test.go`) had to gain a `"pubKey"` field to
keep passing, because until now nothing checked origination at all.

Two things to raise:

1. **An ADVERT with no decoded `pubKey` is now excluded from its own node's skew.** That is the
   right call (origination cannot be proven), but it silently shrinks the sample rather than
   erring. If any older `decoded_json` rows in the retention window lack `pubKey`, those nodes
   lose skew data on upgrade. Worth asking whether that case exists in practice.
2. It adds another `ParsedDecoded()` caller. Read-only here, so safe, but see the #1871 finding:
   that method hands out a shared cached map and is becoming load-bearing across the codebase.

PR #1821, opened 2026-07-04
fix(packets): observer filter hides multi-observer rows in grouped mode (#1748)

### LINKED ISSUE #1748: Filter by observer is broken
Some packets do not showed in packet history if filtered by observer.
For example filter "hash == NNNN" shows correct packet, but if we also filter by observer "hash == NNNN && observer_id == PUBKEY" (this observer listed in previous query) no packet is shown, unless filtering by observer which first logged this packet.
Both observer, observer_id results in empty list for such packets, the same happens if filtering only by observer button and scolling down the history.

So if operator only wants to see packets logged by specific observer, some may be not shown.

Saarlandpower:
Closes #1748.

## Root cause

`renderTableRows()` in `public/packets.js` re-filtered the already server-filtered `/api/packets` (grouped) results client-side, comparing `filters.observer` against each row's `observer_id`. But that field is only the **representative** observer chosen for display — `QueryGroupedPackets` (`cmd/server/db.go:554-606`) picks the observation with the longest observed path via a `LEFT JOIN ... ORDER BY length(path_json) DESC LIMIT 1`, purely for display purposes.

The server-side filter (`buildTransmissionWhere`, `cmd/server/db.go:725-790`) is already correct: it uses an `EXISTS` subquery over **all** observations of a transmission, so any row it returns was genuinely seen by at least one selected observer.

The client then discarded rows *again*, checking only the representative's `observer_id`, with a fallback to `p._children` — which is `undefined` on initial page load (only fetched lazily on row-expand or when the observer-sort dropdown changes, see the `obsSortSel` change handler). Net effect: a multi-observer transmission stayed visible under an observer filter only when the filtered observer happened to also be the representative (longest-path) observer — which matches the reported "works only for whichever observer logged it first" behavior in dense meshes, where longest-path and earliest-seen correlate.

## Fix

Skip the client-side observer re-filter entirely when `groupByHash` is active — the server's `EXISTS` filter is authoritative for grouped rows and needs no client-side correction. The flat/expanded-mode path (single-observation rows, each with its own exact `observer_id` from `buildPacketWhere`) keeps the existing children-aware filter unchanged, which already had test coverage under #537 for the case where `_children` is already populated.

## Tests

Added 6 cases in `test-frontend-helpers.js` covering the specific gap #537's tests didn't reach — grouped mode with `_children` still `undefined` (the actual initial-load state that triggers this bug). New tests confirm:
- A multi-observer row whose representative doesn't match the filter is kept in grouped mode (the core bug).
- Grouped mode never re-filters client-side (trusts the server).
- Flat mode behavior is unchanged (matches by own `observer_id`, falls back to already-loaded `_children`).

Full run: `test-frontend-helpers.js` 631 passed / 2 failed (same 2 pre-existing `favStar` failures reproduce identically on unmodified `master` — unrelated). `test-packet-filter.js` 92/92. `test-aging.js` 18/18.

Operator context: running CoreScope for SaarMesh (SaarLorLux, DE/FR/LU, 800+ nodes, 14 observers) — this was hiding a large share of traffic whenever filtering by a non-primary observer in our dense mesh.

### DIFF
--- a/public/packets.js
+++ b/public/packets.js
@@ -2719,6 +2719,47 @@
     });
   }
 
+  // applyObserverFilter decides which already-loaded packets remain visible
+  // under the current observer filter. Extracted into its own function
+  // (rather than left inline in renderTableRows) specifically so tests can
+  // exercise the real production logic instead of a hand-copied
+  // reimplementation — see #1748 PR review (kent-beck): a test that only
+  // checks a copy of this logic doesn't fail if this function regresses.
+  //
+  // #1748: In grouped mode, the server already filters transmissions
+  // correctly (buildTransmissionWhere emits an EXISTS subquery over ALL
+  // observations of the transmission, not just the displayed one — see
+  // cmd/server/db.go). Each row's `observer_id` here is only the
+  // *representative* observer chosen for display (longest observed path),
+  // which may legitimately differ from the observer that satisfied the
+  // filter. Re-filtering client-side against that single representative —
+  // with `_children` still unpopulated at this point (only fetched lazily
+  // on row-expand or observer-sort-change) — hid every multi-observer
+  // transmission whose representative happened not to be one of the
+  // selected observers. In practice this meant a transmission only stayed
+  // visible when the filtered observer was also the one with the longest
+  // path (which is why the report described it as "works only for
+  // whichever observer logged it first" in dense meshes, where
+  // longest-path and earliest-seen correlate). The server-side EXISTS
+  // filter is authoritative for grouped rows, so no client-side
+  // re-filtering is needed or correct here.
+  //
+  // Flat/expanded mode (groupByHash === false) has no such
+  // representative-vs-actual mismatch — buildPacketWhere filters each
+  // observation row by its own exact observer_id — but we keep the
+  // defensive re-filter for that path since it costs nothing and guards
+  // against any future flat-mode server change.
+  function applyObserverFilter(displayPackets, filters, groupByHash, hashOnly) {
+    if (hashOnly || !filters.observer) return displayPackets;
+    if (groupByHash) return displayPackets;
+    const obsIds = new Set(filters.observer.split(','));
+    return displayPackets.filter(p => {
+      if (obsIds.has(p.observer_id)) return true;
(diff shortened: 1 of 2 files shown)


Saarlandpower:
Addressed the MAJOR from the round-1 review (tests mirrored the fix instead of exercising it):

- Extracted the observer-filter block out of `renderTableRows()` into a named `applyObserverFilter(displayPackets, filters, groupByHash, hashOnly)`, added to the existing `_packetsTestAPI` export (same pattern as `_calcVisibleRange`).
- Tests now load the real `packets.js` via sandbox and call `_packetsTestAPI.applyObserverFilter` directly — no reimplementation left in the test file.
- Manually verified fail-on-revert: temporarily short-circuiting the `groupByHash` early-return (simulating the pre-#1748 bug) turns the 3 grouped-mode tests red; restoring the fix turns them green again.
- Added 2 more cases while refactoring: `hashOnly` bypass and the no-filter passthrough.

`test-frontend-helpers.js`: 633 passed / 2 failed (same 2 pre-existing `favStar` failures, unrelated).

efiten:

efiten:

**Verdict: approve.** The diagnosis is right: in grouped mode each row's `observer_id` is only
the *representative* observer (longest observed path), while the server's `buildTransmissionWhere`
already filtered with an EXISTS subquery over all observations of the transmission. Re-filtering
client-side against the representative, with `_children` still unfetched, dropped every
multi-observer transmission whose representative was not one of the selected observers. Dropping
the grouped-mode re-filter is the correct fix, and extracting `applyObserverFilter` so the tests
exercise the real function rather than a copy is the right call.

**Concern I raised and then cleared, recorded so nobody re-raises it:** with the grouped-mode
filter gone, I expected live WebSocket packets to slip through unfiltered, because
`packets.unshift(newGroup)` at `public/packets.js:1311` prepends live rows into the same array
`renderTableRows` reads. There is already a guard on the insert path (`packets.js` ~1249:
`if (filters.observer) { ... return false; }`), and for a live packet `p.observer_id` is the
actual observing observer, not a representative, so that guard is correct. No leak.

Issue #1777, opened 2026-06-22
feat: Treat SRC-ID as Hop


damn-simple-scripts:
Currently detecting neighbours is done by analyzing paths.
Adverts as exception add the originator of the advert + first hop as Edge.

Although ANON_REQ similar to ADV contains the full senders pubkey, they are not used for nieghbour detection.

According to Meshcore documentation "Returned path, request, response, and plain text message" contains a src and a dst 1 byte ID.

SUGGESTION: Similar to Adverts we should handle ANON_REQ as they contain the senders pubkey and therefore are very reliable information.
SUGGESTION: Similar to Adverts we could treat the src as hop, so we could process the addition edge src+first hop.

Failing to do so, results in a situation where companions are rarely listed as neighbours because companions usually are sender or receiver of messages but do not appear in the path. Currently only companions that are an observer or advert a lot are listed as neighbours.
Failing to implement handling of RESP could lead to situation in dense networks where we miss neighbourhood relationships.

Failing to list them as neighbours hinders analysis of the mesh (e.g. who is sending out REQ?)

Kpa-clawbot:
**Severity:** P3 — protocol feature gap, no breakage.

**Claim:** partial. Reporter conflates two distinct proposals; one is clean, the other is dubious.

**Code today:**
- `cmd/ingestor/neighbor_builder.go:209` — `isAdvert := payloadType == 0x04`; only ADVERTs (payload 0x04) emit the `originator↔path[0]` edge.
- `cmd/server/neighbor_graph.go:318-323` (in-memory mirror) — same ADVERT-only gate.
- ANON_REQ (0x07) IS decoded with the sender's full Ed25519 pubkey: `cmd/server/decoder.go:471` exposes `ephemeralPubKey` (32 B at offset 1), matching firmware `firmware/docs/payloads.md:177-185`.

**Findings:**
- (meshcore): Suggestion #1 (treat ANON_REQ like ADVERT) is protocol-sound — the wire format DOES carry the sender pubkey, and our decoder already extracts it. The builder + graph just don't consume it. Edge `ephemeralPubKey ↔ path[0]` is structurally identical to the ADVERT case.
[...]

Issue #1143, opened 2026-05-07
bug: pubkey attribution via LIKE-on-JSON is unsound — adversarial spoofing + same-name false positives + perf


Kpa-clawbot:
# bug: pubkey attribution via LIKE-on-JSON is unsound — adversarial spoofing + same-name false positives + perf

> **MIGRATION CONSTRAINT (added 2026-05-07):** Production databases have **>10M observations** and 100K+ transmissions. Sync migration is NOT acceptable — would block startup for minutes and starve MQTT ingest. **Backfill MUST follow the existing async pattern**, modeled on `BackfillPathJSONAsync()` (`cmd/ingestor/db.go:1101-1180`) — background goroutine, batched (`batchSize = 1000`), throttled with yield between batches, idempotent via `_migrations` table marker, panic-recovered, progress-logged every N batches.

### Hole 1 — name LIKE
`cmd/server/db.go:910`: query is `WHERE decoded_json LIKE '%pubkey%' OR decoded_json LIKE '%name%'`. Any other node sharing the operator-chosen display name gets its transmissions attributed.

### Hole 2 — pubkey LIKE itself is unsound

`LIKE '%pubkey%'` is also wrong:

1. **Adversarial:** an operator can name their node `"f7181c468dfe7c55..."` (someone else's full pubkey). Their transmissions then show up in the *victim's* recentAdverts. Trivial audit-trail spoofing.
2. **Accidental:** any free-text field in `decoded_json` (channel names, GRP_TXT message bodies, path strings, observer IDs in path arrays) that happens to contain a 64-char hex substring matching a real pubkey produces a false positive.
3. **Performance:** `LIKE '%substring%'` defeats index usage. Full table scan per call. At 100K+ transmissions and thousands of node-detail page loads/day, this is wasteful.

## Affected call sites

```
cmd/server/db.go:582      decoded_json LIKE ?  — packet search (verify)
cmd/server/db.go:626      t.decoded_json LIKE ?  — packet listing filter (verify)
cmd/server/db.go:910      t.decoded_json LIKE ? OR t.decoded_json LIKE ?  — recentAdverts (in scope)
cmd/server/db.go:1785     t.decoded_json LIKE ?  — multi-pubkey OR query (in scope if pubkey-attribution)
cmd/ingestor/db.go:327    t.decoded_json LIKE '%' || nodes.public_key || '%'  — backfill (in scope)
```

[...]

PR #1822, opened 2026-07-04
feat(neighbor-graph): treat ANON_REQ as an originator↔path[0] edge source (#1777)

Saarlandpower:
Addresses #1777 — implements **suggestion #1 only** (extend ADVERT-only originator-edge logic to also accept ANON_REQ). Suggestion #2 (treat REQ/RESP/PATH/TXT's 1-byte src/dst hash as a hop) was explicitly rejected in the issue thread by both the meshcore-reviewer and the reporter themselves — ~1/256 collision odds, would manufacture false edges.

## What

ANON_REQ (payload type `0x07`) carries the sender's full Ed25519 ephemeral pubkey (`decoder.go`'s `EphemeralPubKey` / `ephemeralPubKey` JSON field) — the same trust level as an ADVERT's `pubKey`. Both places that build the originator↔path[0] edge only recognized ADVERT:

- `cmd/ingestor/neighbor_builder.go:209` (persisted `neighbor_edges`, read by the server)
- `cmd/server/neighbor_graph.go:296` / `extractFromNode` (in-memory mirror, `BuildFromStore`)

Both now also accept ANON_REQ via a new `hasFullOriginator := isAdvert || isAnonReq` gate. REQ/RESP/PATH/TXT remain excluded from *originator*-edge creation — their existing observer↔path[last] edge (unaffected by this change, already applies to all packet types) still applies.

## Changes

- `cmd/ingestor/maintenance.go` — new `extractPubkeyFromAnonReqJSON`, mirroring the existing `extractPubkeyFromAdvertJSON`.
- `cmd/ingestor/neighbor_builder.go` — `payloadAnonReq` const; falls back to the new JSON helper when `from_pubkey` is empty (that column is only populated for ADVERT at write time, per `db.go`'s #1143 comment — left untouched to keep this change scoped to the two call sites triage identified).
- `cmd/server/neighbor_graph.go` — `extractFromNode` also checks `"ephemeralPubKey"` (safe unconditionally: the field only appears in ANON_REQ's `decoded_json`, via the shared `Payload` struct's `omitempty`).

## Tests

- `neighbor_graph_test.go`: `TestBuildNeighborGraph_AnonReqSingleHopPath` (ANON_REQ produces the edge) + `TestBuildNeighborGraph_ReqRespStillExcluded` (REQ still doesn't).
- `neighbor_builder_test.go`: same two cases against the persisted-DB path (`TestNeighborEdgesBuilderUpsertsFromAnonReqEphemeralPubKey`, `TestNeighborEdgesBuilderExcludesOtherNonAdvertTypes`), exercising the `from_pubkey`-empty JSON fallback specifically.
- All existing neighbor-graph/-builder tests pass unchanged, including `TestBuildNeighborGraph_ADVERTOnlyConstraint`.
- `cmd/server` full suite: pass, except 2 pre-existing flaky tests (`TestHandleNodePaths_*_1352`, an index-loading race unrelated to this change — reproduces intermittently on unmodified `master` too, confirmed separately).
- `cmd/ingestor` full suite: pass.

Operator context: SaarMesh (SaarLorLux, DE/FR/LU, 800+ nodes) — more originator edges from ANON_REQ traffic should sharpen the neighbor graph / topology views for us.


Saarlandpower:
Addressed both nits from the round-1 review:

1. **(munger) Dead code:** `extractEdgesFromObs`/`edgeCandidate` in `neighbor_persist.go` had zero callers anywhere (verified via grep, including tests) — its doc comment claiming BuildFromStore calls it was already stale. Deleted rather than duplicated the ANON_REQ fix into unreferenced code, per your drift-risk concern.
2. **(carmack) Field-presence safety:** `extractFromNode`'s `ephemeralPubKey` lookup is now gated on `tx.PayloadType == PayloadANON_REQ` instead of relying solely on the JSON key being present, so it stays correct even if a future decoder change reuses that field name elsewhere.

No behavior change — all existing + #1777 tests pass unchanged. Full `cmd/server` suite clean on 2 runs.

PR #1840, opened 2026-07-10
feat(#1784): wire consumers to pathTrust.minHashBytesForMapping config

Bjorkan:
## Summary

Step 3 of the #1784 multi-PR project. It wires `packetpath.MeetsPathTrust` into neighbor-graph construction and path inspection, exposes the effective threshold through `/api/config/client`, and provides shared frontend helpers.

**Depends on:** #1824. This branch carries the foundation files for compilation and must be rebased after #1824 merges.

## Default behavior

The default `pathTrust.minHashBytesForMapping` is **2**. Installations without a `pathTrust` block exclude 1-byte prefixes from derived mapping and topology evidence, reducing collision-driven false positives. Raw packets and paths are never filtered.

Set the value to `1` to retain the legacy trust-all behavior, or `3` to require only 3-byte evidence.

```json
"pathTrust": { "minHashBytesForMapping": 1 }
```

## Changes

- Neighbor-graph construction skips observations below the effective trust threshold.
- Path inspection exposes per-hop `trusted` evidence and reports the active threshold.
- `/api/config/client` serializes the effective `pathTrust` value for frontend consumers.
- Shared frontend helpers evaluate hop and path trust consistently.
- The customizer explains the active trust policy.

## Testing

- Unit tests cover threshold clamping, nil safety, and default behavior.
- Neighbor-graph tests cover 1-byte exclusion at the default and explicit threshold boundaries.
- Frontend tests cover trust helper behavior.

PR #1841, opened 2026-07-10
feat(#1784): wire display consumers to path trust threshold

### LINKED ISSUE #1784: Add configurable trust threshold for 1-byte/2-byte path-hash observations
## Problem

CoreScope currently has several useful tools for visualising and inferring MeshCore paths, neighbour relationships, route patterns, and repeater capability from observed packet paths. However, path-hop observations based on short hash prefixes can create false positives:

- 1-byte path hashes have only 256 possible values, so collisions are expected in larger meshes.
- 2-byte path hashes are much better, but they are still prefix-based observations rather than full identity proof.
- A path-derived observation can therefore accidentally map traffic to the wrong repeater, inflate topology edges, or make CoreScope appear more confident than the evidence really supports.

There is already a display-side option to hide 1-byte path hops, but that does not fully solve cases where the operator wants CoreScope to avoid *trusting* short-prefix observations for topology, route, map, or confidence inference.

## Proposal

Add an operator-configurable setting for the minimum path-hash size CoreScope is allowed to use when building trusted topology/mapping data.

Suggested setting:

```json
{
  "pathTrust": {
    "minHashBytesForMapping": 1
  }
}
```

Suggested UI wording:

**Path trust threshold**

Options:

- **Trust 1-byte, 2-byte, and 3-byte path observations**  
  Default / current behaviour.
- **Ignore 1-byte path observations for mapping/trust**  
  Keep the packets visible and stored, but do not use 1-byte path hops to infer trusted map paths, neighbour edges, route patterns, or repeater capability.
- **Ignore 1-byte and 2-byte path observations for mapping/trust**  
  Only use 3-byte path observations for trusted topology inference. This is conservative, but useful for operators who prefer fewer false positives over broader coverage.

## Important behaviour

This should not drop packets, mutate stored packet data, or change MeshCore firmware behaviour.

Raw observations should still be available in packet detail views. The setting should only control whether short-prefix path observations are used as trusted evidence for derived views and analytics.

## Candidate affected areas

The threshold should ideally apply to:

- Map route/path drawing where CoreScope resolves hash prefixes into node/repeater identities.
- Neighbour graph / inferred topology edges.
- Route pattern analytics.
- Repeater multi-byte capability inference where evidence comes only from path observations.
- Path inspector candidate scoring or “trusted/speculative” classification.
- Confidence indicators that currently aggregate observations by hash mode.

## Acceptance criteria

- Default behaviour remains backward-compatible.
- Operators can choose a minimum trusted path-hash byte length: 1, 2, or 3.
- Ignored observations remain stored and visible in raw packet views.
- Derived views clearly indicate when a trust threshold is active.
- Confidence/tooltips explain that lower-byte observations were excluded from trusted mapping.
- Tests cover at least:
  - default trust-all behaviour;
  - 1-byte excluded, 2/3-byte still used;
  - 1/2-byte excluded, only 3-byte used;
  - raw packet data remains unchanged;
  - affected map/topology/analytics consumers actually respect the threshold.

## Why this matters

For dense or regional MeshCore networks, avoiding false positives can be more important than showing every possible inferred path. A trust threshold would let operators choose the right tradeoff for their deployment: maximum visibility by default, or more conservative mapping where short-prefix collisions are a known problem.

Bjorkan:
Fixes #1784

## Summary

Step 4 of the #1784 multi-PR project — wires all frontend display consumers to respect the configured `pathTrust.minHashBytesForMapping` value.

**Depends on:** #1840 (must be merged first — provides `MC_meetsPathTrust` / `MC_pathBelowTrust` / `MC_getPathTrustThreshold` helpers and `PATH_TRUST` global).

### `map.js` — trust-gated route display
- `drawPacketRoute` checks `MC_pathBelowTrust` before drawing polylines
- When all path hops are below the configured threshold, shows a "Route not displayed" message with config guidance instead of speculative polylines
- Cleans up the trust message control when a new route is drawn

### `analytics.js` — subpath trust filtering
- `renderTable` in `renderSubpaths` filters route patterns whose hops are below trust threshold
- Combined filter logic: both the 1-byte hide toggle AND trust threshold are applied together
- Info line shows which filters are active ("1-byte hide + 2-byte trust", etc.)
- No-data message explains which filters caused exclusion

### `route-view.js` — speculative path annotation
- Path picker groups tagged with `belowTrust` flag when any hop doesn't meet the configured threshold
- Speculative paths show `(speculative, <N-byte hops)` annotation with tooltip explaining the trust threshold
- Tooltip includes guidance on changing `pathTrust.minHashBytesForMapping` in config.json

### `live.js` — Paths Through widget
- `_pathHopsBelowTrust()` helper checks whether all hops in a path are below trust threshold
- Widget fallback message distinguishes between "1-byte filtered" (display toggle) and "N-byte trust threshold" (server config)
- Message includes the config key for operators to adjust

### `nodes.js` — confidence weight adjustment
- `modeWeight` initialization considers the trust threshold
- Hash modes below threshold get zero confidence weight
- Bucket-0 (legacy/unknown) excluded at threshold >= 2 per #1784 bucket-0 policy

## Testing

`test-issue-1633-hide-1byte-hops.js`:
- 5 new source-grep guards verifying each consumer references the trust threshold helpers
- All 26 tests passing (21 existing + 5 new)

## Files changed (6 files, +136/-6)

```
public/analytics.js                | 28 ++++++++++++++++++---
public/live.js                     | 19 ++++++++++++++-
public/map.js                      | 21 ++++++++++++++++
public/nodes.js                    |  8 ++++++
public/route-view.js               | 16 +++++++++++-
test-issue-1633-hide-1byte-hops.js | 50 ++++++++++++++++++++++++++++++
```

---

**Depends on:** #1840
**Written by:** DeepSeek V4 Pro in Max Mode

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+- `/api/paths/inspect` now marks a candidate as `speculative` when any hop
+  falls below `pathTrust.minHashBytesForMapping`; consumers can distinguish
+  this from score-based speculation through `evidence.perHop[].trusted`.
+
 ## [3.9.1] — 2026-06-12
 
 Patch release on top of v3.9.0 — v3.9.0's container image never published (Playwright flake gated Docker build). See [docs/release-notes/v3.9.1.md](docs/release-notes/v3.9.1.md).
--- a/cmd/ingestor/config.go
+++ b/cmd/ingestor/config.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/meshcore-analyzer/dbconfig"
 	"github.com/meshcore-analyzer/geofilter"
+	"github.com/meshcore-analyzer/packetpath"
 )
 
 // MQTTSource represents a single MQTT broker connection.
@@ -56,6 +57,7 @@ type Config struct {
 	Runtime            *RuntimeConfig          `json:"runtime,omitempty"`
 	ClientRxCoverage   *ClientRxCoverageConfig `json:"clientRxCoverage,omitempty"`
 	GeoFilter          *GeoFilterConfig        `json:"geo_filter,omitempty"`
+	PathTrust          *PathTrustConfig        `json:"pathTrust,omitempty"`
 	ForeignAdverts     *ForeignAdvertConfig    `json:"foreignAdverts,omitempty"`
 	ValidateSignatures *bool                   `json:"validateSignatures,omitempty"`
 	DB                 *DBConfig               `json:"db,omitempty"`
@@ -111,6 +113,10 @@ func (c *Config) IngestBufferSizeOrDefault() int {
 
 // GeoFilterConfig is an alias for the shared geofilter.Config type.
 type GeoFilterConfig = geofilter.Config
+
+// PathTrustConfig is an alias for the shared packetpath.TrustConfig type
+// (issue #1784). See packetpath.TrustConfig for the full doc comment.
+type PathTrustConfig = packetpath.TrustConfig
 
 // ForeignAdvertConfig controls how the ingestor handles ADVERTs whose GPS lies
 // outside the configured geofilter polygon (#730). Modes:
@@ -236,6 +242,15 @@ func (c *Config) ObserverDaysOrDefault() int {
 		return c.Retention.ObserverDays
 	}
 	return 14
+}
+
+// GetPathTrust returns the effective path-trust config, applying
+// DefaultMinHashBytesForMapping when unset (issue #1784).
+func (c *Config) GetPathTrust() PathTrustConfig {
+	if c != nil && c.PathTrust != nil {
+		return *c.PathTrust
+	}
+	return PathTrustConfig{MinHashBytesForMapping: packetpath.DefaultMinHashBytesForMapping}
 }
 
 // IsObserverBlacklisted returns true if the given observer ID is in the observerBlacklist.
--- a/cmd/ingestor/config.go
+++ b/cmd/ingestor/config.go
@@ -11,7 +11,6 @@ import (
 
 	"github.com/meshcore-analyzer/dbconfig"
 	"github.com/meshcore-analyzer/geofilter"
-	"github.com/meshcore-analyzer/packetpath"
 )
 
 // MQTTSource represents a single MQTT broker connection.
@@ -57,7 +56,6 @@ type Config struct {
 	Runtime            *RuntimeConfig          `json:"runtime,omitempty"`
 	ClientRxCoverage   *ClientRxCoverageConfig `json:"clientRxCoverage,omitempty"`
 	GeoFilter          *GeoFilterConfig        `json:"geo_filter,omitempty"`
-	PathTrust          *PathTrustConfig        `json:"pathTrust,omitempty"`
 	ForeignAdverts     *ForeignAdvertConfig    `json:"foreignAdverts,omitempty"`
 	ValidateSignatures *bool                   `json:"validateSignatures,omitempty"`
 	DB                 *DBConfig               `json:"db,omitempty"`
@@ -113,10 +111,6 @@ func (c *Config) IngestBufferSizeOrDefault() int {
 
 // GeoFilterConfig is an alias for the shared geofilter.Config type.
 type GeoFilterConfig = geofilter.Config
-
-// PathTrustConfig is an alias for the shared packetpath.TrustConfig type
-// (issue #1784). See packetpath.TrustConfig for the full doc comment.
-type PathTrustConfig = packetpath.TrustConfig
 
 // ForeignAdvertConfig controls how the ingestor handles ADVERTs whose GPS lies
 // outside the configured geofilter polygon (#730). Modes:
@@ -242,15 +236,6 @@ func (c *Config) ObserverDaysOrDefault() int {
 		return c.Retention.ObserverDays
 	}
 	return 14
-}
-
-// GetPathTrust returns the effective path-trust config, applying
-// DefaultMinHashBytesForMapping when unset (issue #1784).
-func (c *Config) GetPathTrust() PathTrustConfig {
-	if c != nil && c.PathTrust != nil {
-		return *c.PathTrust
-	}
-	return PathTrustConfig{MinHashBytesForMapping: packetpath.DefaultMinHashBytesForMapping}
 }
 
 // IsObserverBlacklisted returns true if the given observer ID is in the observerBlacklist.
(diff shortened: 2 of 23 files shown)


Bjorkan:
Review feedback addressed (commits `de71d2b`, `368211a`):

**MAJOR 1 (dijkstra):** Unified all inline fallbacks for MC_getPathTrustThreshold to 1 across map.js, live.js, route-view.js (was inconsistent: 3 consumers used 2, 2 used 1). The production initialization path (roles.js) sets the real default (2); inline fallbacks are dead-code safety nets.

**MAJOR 2 (kent-beck):** Added anti-tautology behavioral test (#1784: anti-tautology — MC_pathBelowTrust returns correct boolean for given hops). Sets PATH_TRUST=2, verifies 2-byte hop is NOT below trust, 1-byte hop IS below trust, mixed path works, and threshold 1 returns false. A stub returning hardcoded true would fail this test.

**MINOR 3 (carmack):** Removed dead code in nodes.js — duplicate modeWeight[0]=0 assignment (loop already zeroes it).

**MINOR 4 (tufte):** Replaced hardcoded #1e293b / #e2e8f0 fallbacks with pure CSS vars (--input-bg / --text) in map.js trust-message control.

**Additional fixes from subagent review:**
- map.js: _hb now computes max byte length across ALL hopKeys (not just hopKeys[0]), so trust-message correctly reports worst-case
- analytics.js: updated stale comment claiming default=1

**Chain:** Rebased on updated feat/path-trust-consumers-1784 (includes DefaultMinHashBytesForMapping=2, roles.js fallback fix, routes.go GetPathTrust fix). All 37 JS tests pass.

Bjorkan:
Review feedback addressed (commit `418d327`)

1. Remove the prior map trust-warning control before evaluating every new route, preventing a stale "Route not displayed" message after a trusted route is selected.
2. Moved trust-warning presentation into themed CSS and kept the config guidance consistent with other consumers.
3. Added a regression guard that verifies control cleanup occurs before the trust gate.
4. Restored the path-inspection algorithm comments, reuse the resolved trust config, and documented that `speculative` now includes below-trust hops. The changelog points API consumers to `evidence.perHop[].trusted` for the reason.
5. Documented intentional strict handling of legacy bucket 0 confidence.

Validation: `node test-issue-1633-hide-1byte-hops.js` passes (37/37); `node test-packet-filter.js` passes (92/92); `node test-aging.js` passes (18/18). Go tests could not run locally because the Go toolchain is unavailable, and browser validation could not run because no local server is available.

Bjorkan:
Validation update (commit `d878d73`)

1. Carried the explicit legacy one-byte test setup through the dependent branch.
2. Passed: `go test ./...` in `cmd/server` (15.9s).
3. Passed: `node test-issue-1633-hide-1byte-hops.js` (37/37).

The prior validation also passed `go test ./...` in `cmd/ingestor` and `internal/packetpath`, plus the packet-filter and aging suites.

efiten:

efiten:

**This is the finding that matters most about these three, and it changes the merge plan I gave
earlier.** I read the commit lists rather than assuming the ordering from the titles:

- **#1824** is `beae2c1c` plus two merges from master.
- **#1863** starts from those same three commits and adds three of its own. So **#1863 is genuinely
  stacked on #1824** (ingestor side: `neighbor_builder.go`, `maintenance.go`, `main.go`,
  `neighbor_persist.go`).
- **#1840** does *not* contain `beae2c1c` at all. It starts from `9baf2b16` and **re-implements
  #1824's base independently** — its diff re-adds `internal/packetpath/trust.go`, the
  `config.example.json` pathTrust block and `GetPathTrust` in both `config.go` files. One of its
  own commits is even titled "sync default to 2", i.e. re-deriving the base by hand.
- **#1841** shares `9baf2b16`, `3f6f77c2` and `1b536286` with #1840 and adds four more, so
  **#1841 contains #1840 entirely** (display side: `analytics.js`, `live.js`, `map.js`,
  `nodes.js`, `route-view.js`, `style.css`).

So the real shape is two parallel branches that both carry the same base:

```
#1824 ──> #1863            (ingestor consumers)
#1840 ──> #1841            (server + display consumers, base re-implemented)
```

**Consequences for merging, in order of how much pain they save:**

1. **#1841 supersedes both #1840 and #1824.** Merging #1841 gets the base and the server and
   display consumers in one go. #1840 and #1824 should then be closed as superseded, not merged.
2. **#1841 and #1863 will conflict with each other**, because each adds `packetpath/trust.go`, the
   `config.example.json` block and `GetPathTrust` from a different commit. Both report
   `MERGEABLE` today only because both are measured against master, not against each other.
   **Do not merge them back to back.** Merge one, then rebase the other so it drops its duplicate
   base, then merge that.
3. The earlier advice to merge "1824 → 1840 → 1841 → 1863 in order" is wrong and should not be
   followed. It came from the PR titles; the commit graph says otherwise.

On the code itself: all three use `packetpath.MeetsPathTrust` rather than reimplementing the
prefix-length test per call site, which is what the #1784 audit asked for (27 uses in #1840 and
#1841, 22 in #1863). The behaviour concern is the one recorded under #1824: these are the PRs that
make the default flip real.

efiten:
Merging this as the pathTrust vehicle in the #1922 queue run, ahead of #1824 and #1840.

Reason, from reading the commit graph rather than the titles: #1841 contains #1840 entirely (they share `9baf2b16`, `3f6f77c2` and `1b536286`), and #1840 re-implements #1824's base by hand rather than building on it. So merging #1841 brings the base, the server consumers and the display consumers in one go, and #1824 and #1840 become redundant rather than pending. I am closing both as superseded once this lands, with the reasoning on each.

**#1863 will need a rebase after this.** It is stacked on #1824 and carries the same base from a different commit, so the two share `internal/packetpath/trust.go`, `trust_test.go`, `cmd/server/config.go`, `neighbor_graph.go`, `neighbor_graph_test.go` and `config.example.json`. Both report MERGEABLE today only because each is measured against master and not against the other. I will do that rebase.

**The default is still the open question and it is not blocked by this merge.** `DefaultMinHashBytesForMapping = 2` means every instance that never touches config stops counting 1-byte prefix hops as mapping evidence, and legacy persisted neighbour edges (`prefixBytes == 0`) drop out with them. Measured on a live network: 116,923 of 208,707 path-hop observations are 1-byte (56%), and 645 of 1,577 repeaters use a 1-byte hash (41%). Issue #1784's own first acceptance criterion is "Default behaviour remains backward-compatible" and its example config shows `minHashBytesForMapping: 1`. That is worth a release-note headline at minimum, and arguably worth shipping 1 for one release first. @Bjorkan, your call as the issue author.

PR #1852, opened 2026-07-18
Region-scope visibility: Channels, Packets, and a new Scopes analytics tab

dborup:
## Summary

Adds end-to-end visibility into MeshCore transport-region scopes (`#dk`, `#dk-oj`, etc.) across the UI — from individual messages/packets up to a full analytics dashboard — plus operational tooling (config hot-reload) and three correctness fixes found by cross-referencing real data against the actual MeshCore firmware source.

### User-facing: scope on messages and packets
- Channel messages (REST + live WebSocket) and the Packets-tab transport badge now show which region scope a message/packet carried, or "unknown" when transport-eligible but unresolved.
- Packets-tab badge shows the resolved scope name directly in the label (not just on hover), with a max-width + ellipsis so long region names don't blow out the column.
- Packets tab also gets a dedicated, sortable **Scope** column (between Type and Observer) so scoped traffic can be sorted/scanned directly instead of only spotted via the inline badge — wired into the existing column visibility toggle and mobile column-hiding. Returning visitors with a saved column-visibility preference (predating this column) now get new columns shown by default instead of silently missing them, a general fix to the column-prefs load path.

### New: Scopes analytics tab

A full "Scopes" tab under Analytics, built out incrementally:

| Section | What it answers |
|---|---|
| **Scope Statistics** | Existing summary/byRegion/time-series, now more accurate (see fixes below) |
| **Channel Messages** | Of channel chat specifically (not all traffic), how much is scoped vs unscoped vs unknown |
| **Region Utilization** | All-time diff between configured `hashRegions` and regions that have ever actually matched — surfaces how much of the region list is dead weight |
| **Repeaters by Region** | Which repeaters have relayed traffic carrying each region scope |
| **Bridge Repeaters** | Repeaters relaying for *more than one* region — the mesh's literal backbone nodes connecting separate regional communities |
| **Nodes Running This Region** | Nodes whose OWN `default_scope` is a given region — as distinct from repeaters that merely relay it |
| **Activity by Hour of Day** | Per-region heatmap of when during a typical day (UTC) each region is active |
| **Scope Adoption by Channel** | Of the top 30 channels by volume, which ones actually use region scoping vs never |

### New: reload `hashChannels`/`hashRegions` without restarting (SIGHUP)

Adding a channel or region to `config.json` previously required a full ingestor restart to take effect — which also reset in-memory analytics state (Repeaters by Region, Bridge Repeaters, etc.) that takes real time to rebuild from live traffic. The ingestor now reloads `hashChannels`/`hashRegions` on `SIGHUP` (the standard Unix "reload config" convention) via an atomic key swap (`cmd/ingestor/hot_reload.go`), so config changes take effect without losing warm-up state. A malformed `config.json` during a live reload leaves the previous keys in place instead of blanking them. Documented in `docs/deployment.md` under "Reloading config changes without a restart (SIGHUP)".

### Correctness fixes (found via real data on stg.meshview.dk, ~1100 configured `hashRegions`)

1. **HMAC collision on scope matching.** `matchScope` (ingestor) returned the first configured region whose HMAC happened to match a packet's 16-bit `transport_code_1`, iterating a Go map in randomized order. With enough configured regions, an unrelated region can coincidentally collide (~`len(regionKeys)/65534` per packet — birthday paradox). The same packet could resolve to different, sometimes-wrong region names across ingestor restarts. Now scans every configured region and only returns a name on an unambiguous single match; ambiguous matches report unknown-scoped rather than guessing.

2. **False repeater-scope attribution via ambiguous hash-prefix bucket.** `TransportedScopes` (both the bulk `computeRepeaterRelayInfoMap` and per-node `GetRepeaterRelayInfo`) folded a full pubkey's path-hop packets together with its 1-byte raw-prefix fallback bucket — an existing, intentional approximation for relay-*activity* counts, which tolerates the ambiguity of a 1-byte hash matching many nodes. Applying that same fold to *region-scope attribution* let a repeater get credited with a scope from a packet it never actually carried — confirmed directly against the MeshCore firmware source (`examples/simple_repeater/MyMesh.cpp` `allowPacketForward` + `RegionMap::findMatch`): a repeater only relays a `TRANSPORT_FLOOD`/`TRANSPORT_DIRECT` packet when its own locally configured region matches, so the fold was asserting something the protocol itself wouldn't allow. Scope accumulation is now gated to the exact-key (unambiguous) match only; relay-activity counts keep the existing fallback.

3. **`byPathHop` bucket keys surfaced as fake repeaters.** The 1-byte raw-prefix bucket keys (e.g. `"b3"`) never correspond to a real `nodes.public_key`, but the first cut of "Repeaters by Region" iterated the relay-info map without filtering, so these bucket keys leaked into the UI as phantom "repeaters". Fixed by requiring `role IN ('repeater','room')` existence in the same lookup used to resolve display names.

### Housekeeping
- `docs/deployment.md`: brought in a pending restructuring of the deployment guide alongside the SIGHUP section.
- Removed a stale duplicate git-tracked entry (`docs/DEPLOYMENT.md` vs `docs/deployment.md`) — two case-variant paths tracked as separate blobs, which collide into one file on case-insensitive filesystems (macOS/Windows), silently going stale on one side.
- Gitignored `cmd/server/prune-requests/`, a runtime job-queue directory that shouldn't have been trackable in the first place.

## Test plan
- [x] `go test ./...` (cmd/server, cmd/ingestor) — all green, including new regression tests for each fix (HMAC collision, prefix-bucket scope attribution, bucket-key filtering), each new analytics section, and the SIGHUP hot-reload path (`hot_reload_test.go`, including an end-to-end test sending a real `SIGHUP` to the test process)
- [x] Relevant frontend unit tests (`test-channels-merge-1498-unit.js`, `test-channel-live-decrypt.js`, `test-issue-1849-trace-hashbytes.js`) — all green
- [x] `eslint public/*.js` — 0 errors (pre-existing unused-var warnings only)
- [x] Verified live against real mesh data on stg.meshview.dk for every section, including manually recomputing HMAC matches to confirm the collision/false-attribution fixes against concrete real-world examples (e.g. a hyper-local scope showing repeaters hundreds of km away, traced back to the prefix-bucket bug and confirmed fixed)
- [x] Verified SIGHUP reload live: added a region to `config.json` on stg, sent `kill -HUP` to the ingestor, confirmed the new region matched without a restart
- [x] Verified the new Packets Scope column live: sortable, toggle-able, and confirmed it appears automatically for a browser with a pre-existing (pre-this-PR) saved column-visibility preference
- [ ] Playwright e2e suite — not run locally (no Playwright in this environment); will run in CI


dborup:
Thanks for the concrete numbers, this is exactly the kind of data point a mesh our size can't produce. Fixed: added `minInteriorEdgeHashBytes = 2` and gated the interior loop on it (both hops must be >= 2 bytes / 4 hex chars, or the pair is skipped entirely) — same default #1824 already settled on, inlined locally rather than waiting on that config to land and get wired in. 1-byte prefixes now produce zero interior edges regardless of whether they'd currently resolve unambiguously, so the "unique today, wrong tomorrow" case you described can't happen through this path. New test (`TestNeighborEdgesBuilderInteriorHopEdges_ExcludesOneByte`) pins it down.

Once #1824's `packetpath.MeetsPathTrust` helper is actually wired into call sites, happy to swap this local constant for that shared config instead of keeping a second parallel threshold around.

Good heads-up on #1822 too — noted, will rebase `neighbor_builder.go` around the `hasFullOriginator` block if that lands first.

Appreciate the offer on the larger dataset — may take you up on it if a similar question comes up for the multi-byte capability work.

dborup:
Closing this out — it's grown to track the full diff between this fork's `master` and upstream (131 files at this point) rather than staying a focused, mergeable change, so a formal PR isn't really the right shape for it anymore.

Everything in here is still available directly from `dborup/CoreScope` on `master` if anything's useful to pull from — happy to help with anything specific if it'd be easier as a smaller, scoped PR instead.

Thanks for the reviews along the way, especially the concrete data point on the interior-hop edge case — genuinely useful feedback.

PR #1871, opened 2026-07-28
perf: use cached ParsedDecoded() instead of repeated json.Unmarshal

Joel-Claw:
## Problem

`StoreTx.ParsedDecoded()` already caches the result of `json.Unmarshal` on first call via `sync.Once`. However, 13 call sites in `store.go` and 1 in `routes.go` were independently unmarshaling `DecodedJSON` into local `map[string]interface{}` variables on every access, completely ignoring the cache.

## Impact

For a store with 50k+ transmissions, each analytics endpoint that iterates all packets re-parses 50k JSON strings per request. With multiple endpoints, this means hundreds of thousands of redundant `json.Unmarshal` calls per page load, each allocating new maps and slices.

## Fix

Replace each `var d map[string]interface{}; json.Unmarshal([]byte(tx.DecodedJSON), &d)` with `d := tx.ParsedDecoded()`, which returns the cached parse result (parsed once, reused forever).

### Call sites changed (14 total):
- `untrackAdvertPubkey` — advert PK extraction during eviction
- Ingestion path — payload field for API responses
- `evictStaleInternal` — node cleanup during eviction
- `GetAnalyticsTopology` — node PK extraction
- `GetAnalyticsHashCollisions` — advert PK extraction
- `GetAnalyticsDistance` — region node PK building
- `resolveAreaNodes` — node PK extraction
- `GetRecentPackets` — payload field for API response
- `routes.go` byType grouping — type field extraction

### Left unchanged (3 sites):
Three call sites that unmarshal into typed structs (`grpDec`, `decodedMsg`, `decodedGrp`) cannot use `ParsedDecoded()` since they need specific struct types.

## Testing
- `go build` passes
- No behavior change — same data, same logic, just avoids redundant parses

### DIFF
--- a/cmd/server/routes.go
+++ b/cmd/server/routes.go
@@ -2954,8 +2954,7 @@ func (s *Server) handleAudioLabBuckets(w http.ResponseWriter, r *http.Request) {
 			}
 			typeName := "UNKNOWN"
 			if tx.DecodedJSON != "" {
-				var d map[string]interface{}
-				if err := json.Unmarshal([]byte(tx.DecodedJSON), &d); err == nil {
+				if d := tx.ParsedDecoded(); d != nil {
 					if t, ok := d["type"].(string); ok && t != "" {
 						typeName = t
 					}
--- a/cmd/server/store.go
+++ b/cmd/server/store.go
@@ -1798,8 +1798,8 @@ func (s *PacketStore) untrackAdvertPubkey(tx *StoreTx) {
 	if tx.PayloadType == nil || *tx.PayloadType != PayloadADVERT || tx.DecodedJSON == "" {
 		return
 	}
-	var d map[string]interface{}
-	if json.Unmarshal([]byte(tx.DecodedJSON), &d) != nil {
+	d := tx.ParsedDecoded()
+	if d == nil {
 		return
 	}
 	pk := ""
@@ -2916,8 +2916,7 @@ func (s *PacketStore) IngestNewFromDB(sinceID, limit int) ([]map[string]interfac
 			},
 		}
 		if tx.DecodedJSON != "" {
-			var payload map[string]interface{}
-			if json.Unmarshal([]byte(tx.DecodedJSON), &payload) == nil {
+			if payload := tx.ParsedDecoded(); payload != nil {
 				decoded["payload"] = payload
 			}
 		}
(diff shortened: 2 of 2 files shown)


efiten:
Your red pipeline here is not caused by this PR, and I have filed it separately as #1925 so it does not sit against your name.

The failing test is `test-issue-1758-ng-filter-rerenders-e2e.js`, which mocks the API completely (`page.route('**/api/analytics/neighbor-graph*', ...)` at line 86) and builds its own graph in the browser. No server response reaches the code under test, so a Go-only change cannot affect its outcome. I reproduced it on clean master under CPU contention: 3 failures out of 6 parallel runs, 0 out of 10 sequential.

Two separate notes on the PR itself, from reading it rather than from CI:

- Seven of the nine call sites only read fields locally and look fine. The two in `IngestNewFromDB` (store.go:2916) and `IngestNewObservations` (store.go:3188) do `decoded["payload"] = payload`, which now puts the store's shared cached map into a response object that gets broadcast, where the old `json.Unmarshal` handed each caller a private map. A consumer that mutates it would corrupt the cache for every later reader of that transmission, and a concurrent read while another goroutine writes is a data race. Would you consider keeping `ParsedDecoded()` for the seven read-only sites and copying (or keeping the local unmarshal) for those two?
- Minor, and possibly an improvement: the old `if json.Unmarshal(...) == nil` succeeded for a `DecodedJSON` of `"null"` and set `payload` to a nil map; `payload != nil` now skips the key instead.

Please do not rebase for the CI on my account. Once #1925 is sorted this should go green on its own.

Joel-Claw:
Thanks for the thorough review, @efiten. I've pushed a fix that addresses both points:

**Data race on broadcast sites:** `IngestNewFromDB` and `IngestNewObservations` now shallow-copy the `ParsedDecoded()` map before placing it in the broadcast response. This prevents consumers from mutating the shared cache or causing a data race. The 7 read-only call sites are unchanged.

**`"null"` DecodedJSON behavior:** You're right that the old `json.Unmarshal` would succeed for `"null"` and set `payload` to a nil map, while `ParsedDecoded()` returns nil and the `payload != nil` check now skips the key entirely. I considered preserving the old behavior but decided the new one is arguably an improvement, since setting `decoded["payload"] = nil` would just add a null field to the broadcast that the frontend would have to handle anyway. Happy to change it if you prefer parity.

I did not rebase for the flaky CI (#1925) per your suggestion.

efiten:
Merging as part of the #1922 queue run. Both review findings are addressed by `44c6f3ac` and CI is green. I verified the shallow copy is sufficient rather than taking it on trust: the only consumers are `cmd/server/websocket.go:285-296`, which hand the map straight to `hub.Broadcast` for JSON encoding and never mutate it, and `ParsedDecoded`'s map is written once under `sync.Once` and never again. No reachable writer, so no deep copy needed. Thanks for the quick turnaround.

Issue #1616, opened 2026-06-06
hard-fix: slide-over close path must not transition through focused-but-hidden state (supersedes #1172)


Kpa-clawbot:
## Summary

Hard-fix replacement for the slide-over focus-restore CI flake family (formerly #1172). Three production-side patches and selector hardening were tried on #1168; all green locally, all flaky in CI Chromium-headless. Recently the `focus-restore@800: Escape returns focus to originating row` assertion joined the flaky set, blocking ~8 unrelated PRs at a time (#1577, #1578, #1592, #1593, #1594, #1595, #1596, #1601 as of this writing). It is no longer a softened-and-deferred test — Escape was hard-asserted and is now actively failing on master ~50% of runs.

The component is the problem, not the test.

## Root cause (best hypothesis after #1172 evidence)

Chromium headless has non-deterministic focus/blur ordering when `panel.hidden = true` happens in the same microtask as a delegated table re-render. The current implementation:

1. Calls `tr.focus()` (or stores a row ref and refocuses after the panel hides)
2. Sets `panel.hidden = true`
3. The table's React-style delegated re-render fires (depending on row data refresh, websocket update, or just layout invalidation)

In that window Chromium can land `document.activeElement` on `<body>` rather than the originating `<tr>`. The four prior fixes (#1172 commit list) each papered over one variant but did not remove the "hidden-but-was-focused" intermediate state Chromium reasons about non-deterministically.

## The fix (architectural, not test softening)

Do all three:

1. **Detach the panel on close, don't `hidden = true`.** Either `panel.remove()` (caching the node for re-mount) or pivot to a single shared panel that's always in the DOM and toggles its `data-state="open|closed"` (with empty content when closed). Either way, eliminate the focused-but-hidden window.
2. **Restore focus by data-value lookup AT restore time, with a MutationObserver fallback.** Cache only `data-value` (already done per `7891b70`), then on close:
   - Synchronously look up `tr[data-value="..."]` and call `.focus()` BEFORE the detach
   - If `document.activeElement` is not the target after the call, attach a one-shot MutationObserver to the table tbody; when the matching row re-attaches, call `.focus()` and disconnect.
3. **Verify with `--repeat-each=20` in the exact CI Chromium-headless config.** Every prior attempt passed locally; the only valid green signal is 20 consecutive headless runs in CI.

## Acceptance criteria

[...]

PR #1872, opened 2026-07-28
perf: prune lastSeenTouched map during eviction, remove dead relayTimes field

Joel-Claw:
## Problem

Two memory issues in `PacketStore`:

1. **`lastSeenTouched` grows unbounded**: This map debounces `last_seen` DB writes for relay nodes (1 entry per unique pubkey). Entries are never removed, even after those nodes are evicted from the store. Over months of operation with thousands of relay nodes, this map accumulates thousands of stale entries.

2. **`relayTimes` is dead code**: The field is declared (`map[string][]int64`) and initialized (`make(map[string][]int64)`) but never populated or read in production code. The `addTxToRelayTimeIndex` function and its tests operate on locally-passed maps, not the struct field.

## Fix

1. **Prune `lastSeenTouched` during eviction**: After each eviction pass, remove entries older than the 5-minute debounce window. This bounds the map to only recently-active relay nodes.

2. **Remove `relayTimes` field**: Delete the declaration and initialization. The `addTxToRelayTimeIndex` function and tests are unaffected (they use their own local maps).

## Testing
- `go build` passes
- No behavior change — `relayTimes` was never used, `lastSeenTouched` pruning only removes entries that are past the debounce window (no functional impact)

### DIFF
--- a/cmd/server/store.go
+++ b/cmd/server/store.go
@@ -177,7 +177,6 @@ type PacketStore struct {
 	byNode        map[string][]*StoreTx      // pubkey → transmissions
 	nodeHashes    map[string]map[string]bool // pubkey → Set<hash>
 	byPathHop     map[string][]*StoreTx      // lowercase hop/pubkey → transmissions with that hop in path
-	relayTimes    map[string][]int64         // lowercase pubkey → sorted unix-millis of relay events (full pubkeys only)
 	byPayloadType map[int][]*StoreTx         // payload_type → transmissions
 	loaded        bool
 	totalObs      int
@@ -644,7 +643,6 @@ func NewPacketStore(db *DB, cfg *PacketStoreConfig, cacheTTLs ...map[string]inte
 		byObserver:    make(map[string][]*StoreObs),
 		byNode:        make(map[string][]*StoreTx),
 		byPathHop:     make(map[string][]*StoreTx),
-		relayTimes:    make(map[string][]int64),
 		nodeHashes:    make(map[string]map[string]bool),
 		byPayloadType: make(map[int][]*StoreTx),
 		rfCache:       make(map[string]*cachedResult),
@@ -4791,6 +4789,19 @@ func (s *PacketStore) evictStaleInternal(rpBatch map[int][]string) int {
 	// Compact resolved pubkey index after eviction sweep
 	s.CompactResolvedPubkeyIndex()
 	s.CheckResolvedPubkeyIndexSize()
+
+	// Prune lastSeenTouched: remove entries older than the debounce window
+	// (5 minutes). This map grows unbounded as new relay pubkeys are seen;
+	// without pruning, a long-running instance accumulates one entry per
+	// unique pubkey ever observed, even after those nodes are evicted.
+	if len(s.lastSeenTouched) > 0 {
+		debounceCutoff := time.Now().Add(-5 * time.Minute)
+		for pk, ts := range s.lastSeenTouched {
+			if ts.Before(debounceCutoff) {
+				delete(s.lastSeenTouched, pk)
+			}
+		}
+	}
 
 	return evictCount
 }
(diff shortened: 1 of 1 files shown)


efiten:
Closing and immediately reopening this to force a fresh CI run. Nothing is wrong with your PR and no action is needed from you.

Your red pipeline was #1923: the packets page defaults to a 15-minute time window, CI freshens the fixture once at the start of the job, and the `#1616` slide-over flake-gate runs about 20 minutes later. By then the packets table is empty, so `packets@800: page renders + first row exists` times out and the two tests that need a row to click fail with it. One expired window, three reported failures, on PRs that never touched the packets page.

That is fixed and merged (#1924). A plain "re-run jobs" would not help, because a re-run reuses the merge commit from the original run and would not contain the fix, so a close/reopen is the way to get a run against current master.

Apologies for the notification noise.

efiten:

**Verdict: approve on the merits, but the CI is red and I did not diagnose why.**

Both halves check out:

1. **`relayTimes` really is dead.** Its only references are the declaration
   (`store.go:182`) and the `make` in `NewPacketStore` (`store.go:649`). Nothing writes to it
   and nothing reads it. `relay_liveness_test.go` looks like a user at a glance but it builds
   its own local `idx := make(map[string][]int64)` and passes that to
   `addTxToRelayTimeIndex`; the string "relayTimes" there is only inside a `t.Error` message.
2. **The 5-minute prune cutoff is the right number.** It matches
   `const debounceInterval = 5 * time.Minute` at `store.go:1758` inside `touchRelayLastSeen`,
   the only place `lastSeenTouched` is read or written (store.go:1767, 1771).

Which also means the prune is **semantically a no-op**: an entry older than the debounce window
would already fail the `now.Sub(last) < debounceInterval` check, so deleting it changes no
behaviour, only memory. That is a good property and worth saying in the review, because it makes
this safe to merge once CI is green.

Ask for: hoist `debounceInterval` to package scope rather than repeating the literal in a second
function, so the two cannot drift apart.

---

**Update 2026-09-02.** The "CI is red and I did not diagnose why" above is resolved: it was #1923, an E2E bug unrelated to this PR, now fixed and merged. After a fresh run against master this PR is **fully green**. The two points that still stand are the request to hoist `debounceInterval` to package scope, and the collision with #1855 — that PR deletes `lastSeenTouched` entirely, so once it lands the prune half here will not compile. Suggest dropping the prune and keeping the dead `relayTimes` removal.

Joel-Claw:
Good catch on the drift risk. Pushed `145ead2d`: hoisted the literal to package scope as `relayLastSeenDebounceInterval` and updated both the debounce check in `touchRelayLastSeen` and the prune cutoff in the eviction path to reference it. Both sites now share one const, so they cannot diverge.

efiten:
#1855 has now merged, so the collision I flagged in the review above is real rather than predicted. **This PR no longer compiles against master.**

`lastSeenTouched` has **zero occurrences** in `cmd/server/store.go` on master: #1855 moved the relay `last_seen` touch to the ingestor and deleted the field, the `touchRelayLastSeen` method and the `debounceInterval` const along with it. Everything in this PR that names `lastSeenTouched` now references something that does not exist. GitHub may still report the PR as mergeable, because the textual merge can succeed while the build cannot.

**What is still valid, and worth keeping:** the `relayTimes` removal. That field is still on master at `store.go:180` and `:644`, still written by nothing and read by nothing. I verified that again just now:

```
relayTimes    map[string][]int64   // store.go:180  (declaration)
relayTimes:   make(map[string][]int64),  // store.go:644  (init)
```

`relay_liveness_test.go` looks like a user at a glance but builds its own local `idx := make(map[string][]int64)`; the string "relayTimes" there is only inside a `t.Error` message.

**Suggested reduction:** drop every hunk touching `lastSeenTouched` and the `relayLastSeenDebounceInterval` hoist, keep the two-line `relayTimes` removal. That leaves a small, obviously-correct PR.

For what it is worth, the hoist was my suggestion in the review and it was the right call at the time. #1855 simply took the whole mechanism somewhere else.

Joel-Claw:
Closing in favour of a clean replacement. After #1855 merged, this PR no longer compiles against master: lastSeenTouched, touchRelayLastSeen, and the debounceInterval const were all removed. The only surviving valid change is the relayTimes removal, which is still on master. New PR incoming with just that two-line deletion.

PR #1873, opened 2026-07-28
perf: pre-allocate filter slices, reuse ctx buffer, cache ReadMemStats

Joel-Claw:
## Problem

Three hot-path inefficiencies causing excess CPU and memory allocations:

### 1. `filterTxSlice` starts with nil slice
`filterTxSlice` is called on the full `s.packets` slice (50k+ packets) for every query that doesn't hit a fast-path index. Starting with `var result []*StoreTx` means Go's append does ~15 growth+copy cycles (1→2→4→8→...→32768→65536) before reaching steady state.

### 2. `resolvePathForObs` allocates per hop
Each hop in the path resolution loop allocates a new `ctx` slice (`make([]string, len(contextPKs), len(contextPKs)+2)`). For a 5-hop path, that's 5 allocations per observation. With 500+ observations per ingest batch, that's 2500+ small allocations.

### 3. `estimatedMemoryMB` calls `runtime.ReadMemStats` without caching
`runtime.ReadMemStats()` triggers a STW (stop-the-world) pause. It's called from stats/debug endpoints (`GetStoreStats`, `GetPerfStoreStats`) that may be polled frequently. The routes.go layer already caches this with a 5s TTL, but the store layer doesn't.

## Fix

1. **Pre-allocate `filterTxSlice`**: `make([]*StoreTx, 0, n/2)` — the 2x over-allocation is cheaper than repeated growth+copy.

2. **Reuse ctx buffer**: Allocate one `ctx` buffer before the hop loop, reset to base length each iteration with `ctx = ctx[:ctxLen]`.

3. **Cache `ReadMemStats`**: 5-second TTL cache matching the routes.go pattern. Uses a package-level mutex (not on `PacketStore`) to avoid adding a field.

## Testing
- `go build` passes
- No behavior change — same results, fewer allocations

### DIFF
--- a/cmd/server/neighbor_persist.go
+++ b/cmd/server/neighbor_persist.go
@@ -132,9 +132,14 @@ func resolvePathForObs(pathJSON, observerID string, tx *StoreTx, pm *prefixMap, 
 		contextPKs = append(contextPKs, strings.ToLower(fromNode))
 	}
 	resolved := make([]*string, len(hops))
+	// Reuse a single ctx buffer across hops instead of allocating per hop.
+	// Capacity +2 for the previous-hop resolved PK that may be appended.
+	ctx := make([]string, len(contextPKs), len(contextPKs)+2)
+	copy(ctx, contextPKs)
+	ctxLen := len(ctx)
 	for i, hop := range hops {
-		ctx := make([]string, len(contextPKs), len(contextPKs)+2)
-		copy(ctx, contextPKs)
+		// Reset to base context (trim any previously appended resolved PK)
+		ctx = ctx[:ctxLen]
 		if i > 0 && resolved[i-1] != nil {
 			ctx = append(ctx, *resolved[i-1])
 		}
--- a/cmd/server/store.go
+++ b/cmd/server/store.go
@@ -480,6 +480,13 @@ type PacketStore struct {
 	trackedBytes     int64          // running total of estimated packet store memory
 	memoryEstimator  func() float64 // injectable for tests; nil = use runtime.ReadMemStats (stats only)
 
+	// Per-store ReadMemStats cache (5s TTL). Fields (not package-level vars) so
+	// that test helpers constructing &PacketStore{...} directly get independent
+	// cache state, avoiding order-dependent test failures.
+	estMemMu  sync.Mutex
+	estMemVal float64
+	estMemAt  time.Time
+
 	// Short-lived cache for the observations aggregate in GetStoreStats (30s TTL).
 	// Avoids a per-/api/stats full-table scan; values accurate to ~30s which is
 	// sufficient for dashboard display.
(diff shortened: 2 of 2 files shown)


efiten:

**Verdict: request changes.** Three unrelated optimisations in one PR, none with a benchmark.

1. ~~gofmt~~ **Withdrawn.** The new var block is misaligned per `gofmt -d`, but so is
   `cmd/server/store.go` at upstream master, and so are `db.go`, `main.go`,
   `cmd/server/decoder.go` and `cmd/ingestor/decoder.go`. The whole repo is gofmt-dirty
   today, which is what #1881 exists to fix. Not a defect of this PR. Do not raise it.
2. **The ReadMemStats cache is process-global, not per-store.** `estimatedMemMu`,
   `estimatedMemVal` and `estimatedMemAt` are package-level vars behind a `*PacketStore`
   method, so every PacketStore in the process shares one 5-second cache. At least twelve test
   helpers build `&PacketStore{...}` directly (`area_filter_test.go:79`,
   `cache_invalidation_test.go:11,296,461`, `channel_analytics_test.go:13`,
   `coverage_test.go:4249,4283,4308,4367,4417` and others), so a value cached by one test is
   returned to the next for five seconds. That is order-dependent test state. It should be a
   field on PacketStore.
3. **`filterTxSlice` pre-allocates `n/2` with no evidence.** The comment argues that 2x
   over-allocation beats repeated growth, but for a selective filter this allocates n/2
   pointers that are never used. Staging runs under a 3GB cap, so a PR titled "perf" that
   raises peak allocation in the common case needs a benchmark before it lands.

**Checked and fine:** the ctx buffer reuse in `resolvePathForObs` is safe. Its only consumer
is `pm.resolveWithContext`, which reads the slice and never stores it (verified across
store.go:6770-6935). Worth saying in the review that it is safe *and* fragile: store.go:6845
already aliases the caller's backing array, so a later change that retains `lowerCtx` would
silently corrupt hop resolution.

Joel-Claw:
Addressed both review points in `4db18f8b`:

1. **ReadMemStats cache moved to per-store fields.** Removed the package-level `estimatedMemMu`/`estimatedMemVal`/`estimatedMemAt` vars and added `estMemMu`, `estMemVal`, `estMemAt` as fields on `PacketStore`. Each `&PacketStore{...}` constructed by test helpers now gets independent cache state, eliminating the order-dependent test failures.

2. **Removed n/2 pre-allocation in `filterTxSlice`.** Reverted to baseline `var result []*StoreTx` (nil-slice append). No benchmark was available to justify the heuristic, and for selective filters the n/2 allocation wastes pointer slots that may never be used.

3. **ctx buffer reuse in `resolvePathForObs`** — no change, as noted in the review.

All tests pass (`go test ./...`, 32.7s).

efiten:
Both points check out in `4db18f8b`, verified in the diff rather than taken on trust.

**The ReadMemStats cache is now per store.** The package-level `estimatedMemMu`/`estimatedMemVal`/`estimatedMemAt` are gone and `estMemMu`, `estMemVal`, `estMemAt` are fields on `PacketStore`. The twelve-odd test helpers that build `&PacketStore{...}` directly each get their own cache state, so the order-dependent result is gone.

**The `n/2` pre-allocation is not adjusted but removed.** `filterTxSlice` is no longer in the diff at all; it touches two files now.

One thing I re-checked because your change makes it load-bearing in a way it was not before. Reusing the ctx buffer means every hop now shares one backing array where each previously got a fresh one, so the "callee must not retain the slice" property went from incidental to required. In `resolveWithContext` the line `lowerCtx := contextPubkeys` does alias the caller's slice, but `lowerCtx` is only ranged over inside the call and is never stored in a map or struct and never returned, so no reference outlives the call. Safe.

One thing to fix at merge rather than in the branch: the title still says "pre-allocate filter slices", which is the half that came out. A squash merge would land a commit in master describing a change it does not contain. Suggested subject:

    perf: reuse ctx buffer in resolvePathForObs, cache ReadMemStats per store (#1873)

Nothing else blocking from my side. Thanks for turning both around this quickly.

PR #1878, opened 2026-07-28
perf: use prepared statements for frequently-called server DB queries

### LINKED ISSUE #1875: perf: use prepared statements in server DB layer (matching ingestor pattern)
## Background

The server (`cmd/server/db.go`) makes 85+ raw `Query`/`QueryRow`/`Exec` calls, none using prepared statements. The ingestor (`cmd/ingestor/db.go`) already uses 10 prepared statements (`stmtGetTxByHash`, `stmtInsertTransmission`, etc.) and sees significant performance benefits.

## Impact

Every raw `db.conn.Query(...)` or `db.conn.QueryRow(...)` call in the server requires SQLite to parse and compile the SQL query from scratch. For queries called frequently (stats endpoints, packet lookups, observer listings), this is wasted CPU per request.

Key hot spots in `db.go`:
- `GetStats()` — 8+ `QueryRow` calls per stats request (line 325+)
- `GetDBStats()` — 10+ `QueryRow` calls (line 371+)
- `QueryPackets()` — complex query builder, called on every packet list request
- Node/observer lookups — called per-request for name resolution

## Proposal

1. Add `*sql.Stmt` fields to the `DB` struct for the most frequently called queries
2. Prepare them during `initDB()` (same pattern as ingestor)
3. Replace raw `db.conn.Query/QueryRow` calls with prepared statement calls
4. Focus on stats and lookup queries first (simple, parameterized, called frequently)

## Benefits

- Eliminates SQL parsing/compilation overhead per request
- SQLite can cache query plans
- Reduces CPU usage on the server's main request path

Joel-Claw:
## Problem

The server makes 72+ raw `db.conn.QueryRow` / `db.conn.Query` calls. Every call requires SQLite to parse and compile the SQL from scratch. The ingestor already uses 10 prepared statements (`stmtGetTxByHash`, `stmtInsertTransmission`, etc.) but the server layer never adopted the pattern.

## Fix

Add 13 prepared statements to the `DB` struct, prepared once at `OpenDB` time and reused across all requests:

| Statement | Used by |
|---|---|
| `COUNT(*) FROM transmissions` | GetStats + 2 query fast paths |
| `COUNT(*) FROM observations` | GetStats |
| `COUNT(*) FROM nodes WHERE last_seen > ?` | GetStats |
| `COUNT(*) FROM nodes` | GetStats |
| `COUNT(*) FROM observers WHERE inactive IS NULL OR inactive = 0` | GetStats |
| `COUNT(*) FROM observations WHERE timestamp > ?` | GetStats (last hour + last day) |
| `SELECT public_key FROM nodes WHERE public_key = ? OR name = ?` | resolveNodePubkey |
| `SELECT id FROM transmissions WHERE hash = ?` | GetObservationsForHash |
| `COUNT(*) FROM nodes WHERE role = ? AND last_seen > ?` | GetRoleCounts |
| `COUNT(*) FROM nodes WHERE role = ?` | GetAllRoleCounts |
| `COALESCE(MAX(id), 0) FROM transmissions` | GetMaxTransmissionID |
| `COALESCE(MAX(id), 0) FROM observations` | GetMaxObservationID |

## Impact

- `GetStats` is called on every `/api/stats` request — 8 queries, now all prepared
- `GetRoleCounts` + `GetAllRoleCounts` — 8 queries (4 roles x 2 functions), now all prepared
- `GetMaxTransmissionID` + `GetMaxObservationID` — polled by WebSocket clients for new-data detection
- `resolveNodePubkey` — called per-request for node name resolution
- `GetObservationsForHash` — called on packet detail lookups

These are the hottest query paths. The remaining ~50 raw calls are dynamic query builders (optional WHERE clauses, GROUP BY, etc.) that are harder to prepare and less frequently called.

## Testing
- `go build` passes
- `go vet` passes
- No behavior change — same queries, same results, just compiled once

Closes #1875

### DIFF
--- a/cmd/server/db.go
+++ b/cmd/server/db.go
@@ -87,6 +87,16 @@ func OpenDB(path string) (*DB, error) {
 	return d, nil
 }
 
+// stmtQueryRow returns a QueryRow-like helper that uses the prepared statement
+// when non-nil (production), or falls back to a direct db.conn.QueryRow query
+// when the statement is nil (test DBs that skip prepareStatements).
+func (db *DB) stmtQueryRow(stmt *sql.Stmt, fallbackSQL string, args ...interface{}) *sql.Row {
+	if stmt != nil {
+		return stmt.QueryRow(args...)
+	}
+	return db.conn.QueryRow(fallbackSQL, args...)
+}
+
 // prepareStatements creates prepared statements for frequently-called queries.
 // SQLite compiles and caches the query plan once, avoiding re-parsing per request.
 func (db *DB) prepareStatements() error {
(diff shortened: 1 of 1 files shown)


efiten:

**Verdict: approve with comments.** No correctness defect found.

1. **The `stmtQueryRow` fallback is genuinely needed**, which I did not expect. Twelve test
   helpers construct `&DB{conn: ...}` directly and never call `prepareStatements`
   (`db_test.go:141,1273,2237`, `coverage_test.go:69`, `multibyte_capability_test.go:39,468`,
   `node_reach_bench_test.go:66,167`, `node_reach_endpoint_test.go:96`, `node_reach_test.go:36`,
   `resolved_index_test.go:1168`, `scopes_test.go:50`). The cost is that every SQL string now
   exists twice, prepared and as a fallback literal, with nothing keeping them in sync. Worth
   asking whether the test helpers should get a constructor instead.
2. `stmtCountObsLastHour` and `stmtCountObsLastDay` are byte-identical SQL
   (`SELECT COUNT(*) FROM observations WHERE timestamp > ?`). Two statements for one query.
3. `OpenDB` now returns an error if any Prepare fails, where a missing column previously
   degraded a single query. `detectSchema` shows the codebase does expect schema variance
   (`isV3`, `hasResolvedPath`, `hasObsRawHex`, `hasScopeName`, `hasLastSeen`,
   `hasMultibyteSupCols`). None of the 13 prepared queries touch a conditional column, so the
   risk is contained today, but the failure mode changed from degrade to refuse-to-start.
   Flag it, do not block on it.

efiten:
This went CONFLICTING because #1909 merged first, which I flagged in my review above as the deliberate order: #1909 is the correctness fix (`detectSchema` failing loud instead of caching a wrong schema mode) and this is the optimisation, so it went first.

The conflict is in the two places I named: both PRs rewrite `OpenDB` right after `d := &DB{conn: conn, path: path}`, and both rewrite `func (db *DB) Close()`. #1909 replaced `d.detectSchema()` with the pinned-connection version and deleted the WAL checkpoint from `Close`; this PR inserts `prepareStatements()` and prepends statement closing.

**Want me to do the rebase?** I have done the same for #1916, #1771 and #1863 today, keeping the original authorship on the commits and opening it as a continuation PR rather than force-pushing to your branch. Say the word and I will, or take it yourself if you would rather.

My review points are unchanged and none of them block: the duplicated SQL between `stmtQueryRow`'s fallback and `prepareStatements` (genuinely needed, because twelve test helpers build `&DB{conn: ...}` directly), the two byte-identical statements for the observations count, and `OpenDB` now refusing to start rather than degrading when a Prepare fails.

Joel-Claw:
Yes, please go ahead with the rebase. Happy to let you handle it given you have the context from the other rebases today. Thanks.

Joel-Claw:
Closing in favour of #1934 (rebase by @efiten). Thanks for handling the rebase.

efiten:
Merged via #1934, which is your two commits rebased onto master with your authorship intact. Closing this as the vehicle rather than the work.

The conflict resolution, for the record: `prepareStatements()` now runs **after** `detectSchema`, so statements can never be compiled against a schema mode #1909 exists to stop trusting. And `Close()` keeps your statement closing but does not restore the WAL checkpoint, since #1909 removed it deliberately on the grounds that a `mode=ro` handle can only ever fail it.

The three review points from my earlier comment are unchanged and none of them blocked: the SQL duplicated between `stmtQueryRow`'s fallback and `prepareStatements`, the two byte-identical observations-count statements, and `OpenDB` now refusing to start rather than degrading when a Prepare fails. Worth a follow-up if you feel like it, not worth holding the PR for.

Thanks for the quick reply on the rebase offer, and for turning around the `ParsedDecoded` copy fix on #1871 this morning.

Issue #1287, opened 2026-05-19
followup #1283: 13 remaining cachedRW write sites in cmd/server/ — schema migrations, backfills, neighbor persistence


Kpa-clawbot:
## Summary
Follow-up to #1283 / PR #1286. After moving VACUUM + PruneOldPackets + PruneOldMetrics + RemoveStaleObservers to ingestor, **13 remaining cachedRW writer call-sites in cmd/server/** still violate the read-only invariant.

## Full list (post-#1286 baseline)
| # | File:line | Function | Operation type |
|---|---|---|---|
| 1 | `ensure_indexes.go:17` | `ensureServerIndexes` | CREATE INDEX on startup |
| 2 | `from_pubkey_migration.go:32` | `ensureFromPubkeyColumn` | ALTER TABLE on startup |
| 3 | `from_pubkey_migration.go:139` | (backfill) | UPDATE backfill rows |
| 4 | `main.go:266` | neighbor-graph build at boot | INSERT neighbor edges |
| 5-13 | `neighbor_persist.go:23,131,250,289,326,364,400,443,609,838` | various `ensure*` + persist | CREATE TABLE / INSERT / UPDATE / DELETE on neighbor_edges, resolved_path, observer.inactive column |

## Why the read-only invariant matters
Per the architecture, server is the read path. Mixing writes here:
- Causes SQLITE_BUSY contention with ingestor (#1283 root cause)
- Makes the read/write boundary fuzzy and hard to reason about
- Forces server to ship docker image with sqlite write permissions
- Means a security/privilege escalation in any server handler can corrupt the DB

## Fix strategy
Each call-site needs to move OUT of cmd/server. Three categories:

### Category A: Schema migrations (idempotent CREATE/ALTER)
- `ensureServerIndexes`, `ensureFromPubkeyColumn`, `ensureResolvedPathColumn`, `ensureObserverInactiveColumn`, `ensureNeighborEdgesTable` (and any other ensure* functions)
- These run at startup, ONCE. Should run from ingestor (which owns the DB) before subscribing to MQTT. Server, on startup, can ASSERT the expected columns/indexes exist and refuse to start if not — but should NOT create them.

### Category B: Backfill jobs (one-shot UPDATEs)
- from_pubkey backfill at `from_pubkey_migration.go:139`
- Move to ingestor maintenance loop OR a separate maintenance binary (cron). Server has no business backfilling.

[...]

PR #1882, opened 2026-08-09
fix: use index from_pubkey in nodes region filter

Jonher937:
The region subquery in GetNodes was pulling the advert pubkey out of decoded_json with JSON_EXTRACT for every row the join touched, instead of reading the from_pubkey column that #1143 already added and indexed

It looks like buildPacketWhere, GetRecentTransmissionsForNode, QueryMultiNodePackets etc. moved to from_pubkey already, but not this.

### DIFF
--- a/cmd/server/db.go
+++ b/cmd/server/db.go
@@ -903,8 +903,11 @@ func (db *DB) GetNodes(limit, offset int, role, search, before, lastHeard, sortB
 			if !db.isV3 {
 				joinCond = "obs.id = o.observer_id"
 			}
+			// #1143: from_pubkey is a dedicated, indexed column populated at
+			// ingest (and backfilled) for ADVERT rows specifically so pubkey
+			// lookups don't need to JSON_EXTRACT + parse decoded_json per row.
 			subq := fmt.Sprintf(`public_key IN (
-				SELECT DISTINCT JSON_EXTRACT(t.decoded_json, '$.pubKey')
+				SELECT DISTINCT t.from_pubkey
 				FROM transmissions t
 				JOIN observations o ON o.transmission_id = t.id
 				JOIN observers obs ON %s


efiten:

**Verdict: one blocking question. Do not merge until it is answered.**

Swapping `JSON_EXTRACT(t.decoded_json,'$.pubKey')` for `t.from_pubkey` is the right direction,
and `cmd/ingestor/db.go:481-488` already does it for the advert_count migration. But that
migration justifies itself with "the column may not yet be backfilled on legacy DBs: the
migration is already marked done on those DBs and won't re-run". A one-shot migration can lean
on that. **A runtime query cannot.** `from_pubkey` is backfilled by a background shim
(`cmd/server/from_pubkey_migration.go`, issue #1287), not by the boot-blocking `from_pubkey_v1`
ALTER.

On an instance where that backfill has not finished, this query returns fewer nodes for the
region filter than the JSON_EXTRACT version did. Silent under-reporting, no error.

Ask the author: is the region filter gated on backfill completion, or should it fall back to
JSON_EXTRACT while `fromPubkeyBackfillSnapshot().done` is false?

efiten:

efiten:
**Withdrawing my blocking question. I measured it instead of leaving you to chase it.**

I asked whether swapping `JSON_EXTRACT(t.decoded_json,'$.pubKey')` for `t.from_pubkey` could silently under-report on an instance where the background backfill (`cmd/server/from_pubkey_migration.go`, #1287) has not finished. That was a fair question for a runtime query, since the one-shot migration in `cmd/ingestor/db.go:481` justifies itself with "the column may not yet be backfilled on legacy DBs" and a runtime query cannot lean on that.

Measured on a live instance with **874,846 transmissions**:

| check | result |
|---|---|
| rows where `JSON_EXTRACT` finds a pubKey but `from_pubkey` is empty | **0** |
| rows where the two return **different values** | **0** |
| distinct pubkeys via `JSON_EXTRACT` | 5,260 |
| distinct pubkeys via `from_pubkey` | **5,260** |

Supporting numbers: 134,286 ADVERT rows, of which **64** lack `from_pubkey` (0.05%), and those 64 have no `pubKey` in `decoded_json` either, which is why the first row of that table is zero. The 740,624 transmissions with an empty `from_pubkey` are non-ADVERT rows that never carried a pubKey for either expression to find. The `from_pubkey_v1` migration row is present.

So the two are equivalent on real data, and your change is a straight index win with no behavioural difference. **Verdict revised to approve.**

One thing worth keeping in the back of your mind rather than acting on: this holds because the backfill has completed here. On a brand-new instance the window between first boot and backfill completion is real, just short. If the region filter ever moves somewhere latency-sensitive at startup, that window is where it would show.

PR #1883, opened 2026-08-09
fix: use replaceState for traces/roles redirects to preserve back-button history

Jonher937:
fix: use replaceState for traces/roles redirects to preserve back-button history

The #/traces/<hash> and #/roles backward-compat redirects used location.hash = ..., which pushes a new history entry instead of replacing the current one. This trapped users navigating back from a trace view: the intermediate #/traces/<hash> entry would immediately re-redirect forward again on hashchange, so back button never reached the packets view they came from.

### DIFF
--- a/public/app.js
+++ b/public/app.js
@@ -1039,16 +1039,20 @@ function navigate() {
   closeNav();
 
   // Backward-compat redirect: #/traces/<hash> → #/tools/trace/<hash> (issue #944).
+  // Uses replaceState (not location.hash =) so this redirect doesn't add its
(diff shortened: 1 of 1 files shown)


efiten:

**Verdict: approve.** The recursion terminates: `#/tools/trace/` does not match `#/traces/`,
and `#/analytics?tab=roles` does not match `#/roles`. `replaceState` correctly suppresses the
hashchange, and the direct `navigate()` call replaces what that hashchange used to trigger.
Ships without a test, which is worth asking for given it is history behaviour.

efiten:

Issue #1172, opened 2026-05-08
bug(slide-over): focus-restore to originating row is CI-flaky on X-click and resize-cleanup paths


Kpa-clawbot:
## Symptom

Two Playwright assertions in `test-slideover-1056-e2e.js` flake intermittently in CI Chromium (headless), despite passing reliably locally:

1. `focus-restore@800: X-button click returns focus to originating row` — after clicking the slide-over X button, `document.activeElement` is occasionally not the originating `<tr data-value="...">`.
2. `resize@800→1440 nodes: cleanup releases panel, backdrop, scroll-lock, focus` — the `focusRestored` portion (the panel/backdrop/scroll-lock cleanup parts of the same test pass reliably).

These are softened to non-fatal warnings on PR #1168 and tracked here.

## What was tried

Four production-side fix attempts on PR #1168 — all green locally, all flaky in CI Chromium:

- `7891b70` — focus-restore survives caller table re-renders (look up row by stored data-value at restore time, not by stale element ref).
- `366af4f` (test-only) — switch tests to stable `data-value` selector to remove false negatives from selector instability.
- `36ebecc` — defer focus-restore via `requestAnimationFrame` so it runs past the keydown unwind.
- `df5397f` — restore focus BEFORE hiding the panel (avoids blur-during-hide reordering).
- `d681505` — `mousedown.preventDefault()` on the X button so the click never moves focus to the X in the first place.

Each survived locally and on Linux container repros, none stuck reliably in CI Chromium-headless.

## Working behaviors (slide-over itself is fine)

Open from row, close via Escape, close via backdrop click, close via X (panel/backdrop go away), body scroll-lock + release, ARIA attrs, focus-trap inside panel, **Escape** focus-restore to row, second-row open race (single backdrop, single onClose), wide-viewport guard (no slide-over ≥1280), resize cleanup of panel + backdrop + scroll-lock.

## Hypothesis

Chromium's focus/blur queue across `panel.hidden = true` + delegated table re-render is non-deterministic in headless CI. Possible directions:

- Replace the `requestAnimationFrame` defer with a small `setTimeout(..., 0)` or `setTimeout(..., 16)` to clear the microtask queue more aggressively.
- Stop relying on `panel.hidden = true` for hide; remove the panel from the document instead so there's no "hidden but focused" intermediate state.
[...]

PR #1884, opened 2026-08-09
feat(packets): add filter to All Observer dropdown

Jonher937:
Lets users type a prefix to filter the observer checkbox list, instead of scrolling a long list.


efiten:

**Verdict: request changes. Two findings, one user-visible.**

1. **`startsWith` should be `includes`.** `item.dataset.obsName.startsWith(term)` means
   searching "brussels" finds nothing when the observer is named "ON4XYZ Brussels". A search
   box that only matches from the first character reads as broken.
2. `background: var(--card-bg, #fff)` puts a hardcoded hex in a component style, and none of
   the neighbouring rules in the same block use fallbacks (`--border`, `--input-bg`, `--text`
   and `--accent` are all bare). Either `--card-bg` exists and the fallback is dead code, or
   it does not and `#fff` is the real value in dark mode, which would be wrong.

No cache-buster comment needed: this repo bumps automatically, see repo facts above.

efiten:

efiten:
Recycling for a fresh run. **The failure looks unrelated to your change, but it is not the flake that was hitting everything else, so it is worth watching.**

Your pipeline failed on:

```
✗ ArrowRight cycles focus across swatches: ArrowRight should move focus to next swatch (was #ef4444, now #ef4444)
```

That is a keyboard-focus test on the customizer colour swatches, which belongs to the focus-race family this repo already has history with (#1172, #1616). It is not #1923 (the packets time-window bug) and not #1925 (the neighbour-graph race). Your PR adds a search input to the observer dropdown in `packets.js` plus CSS, which should not touch swatch focus cycling.

If it fails the same way twice, it is worth a proper look rather than another re-run. My review above still stands on the `startsWith` versus `includes` point.

efiten:
Follow-up on my note above: I said that if it failed the same way twice it would be worth a proper look. **It passed on the rerun**, so the `ArrowRight cycles focus across swatches` failure was a flake from the focus-race family (#1172, #1616) and not caused by your change. Nothing for you to do about CI.

The review point above is still open and is the only thing holding this: `item.dataset.obsName.startsWith(term)` means searching "brussels" finds nothing when the observer is named "ON4XYZ Brussels". A search box that only matches from the first character will read as broken to whoever uses it. `includes` is the one-word fix.

The second point is smaller: `background: var(--card-bg, #fff)` is the only rule in that block with a hardcoded fallback, while the neighbouring `--border`, `--input-bg`, `--text` and `--accent` are all bare. Either `--card-bg` exists and the fallback is dead, or it does not and `#fff` is what dark mode actually gets.

efiten:
Reviewed against master. The escaping is right on both paths (the `data-obs-name` attribute and the rendered label), focusing the input on open is a nice touch, and stopping propagation on the input keeps the menu from closing while typing.

Two things before this goes in.

**The match is prefix-only.** `applyObserverSearchFilter` uses `startsWith`, so typing `ar` does not find `ON8AR`, and typing `hss` does not find `BE-HSS-JessaZH.VIR`. Observer names on this network are mostly prefixed by country and location codes, which is exactly the part a user is least likely to type first. For a control labelled `Search observers…` the expectation is substring matching, and it is a one-word change to `includes`.

**No test.** The repo runs `test-frontend-helpers.js` for exactly this kind of pure rendering and filtering logic. A case asserting that a term matching the middle of a name keeps the item visible would pin the behaviour above, and one asserting that `All Observers` is never hidden by a search term would pin the deliberate exception that is currently implicit in the missing `data-obs-name`.

efiten (changes requested):
Reviewed. The wiring is clean: the search box is sticky inside the menu, its click is stopped so the menu does not close under it, the "All Observers" row has no `data-obs-name` so it survives every filter, and the trigger focuses the input on open. `escapeHtml` covers the new attribute, quotes included. Nothing to fix there.

Two things before this goes in.

**1. Prefix matching makes the box close to useless on a real network.** `startsWith` means the term has to match from the first character. Measured against the 45 observers on a live instance:

- 28 of 45 names start with `BE`, so the first characters carry no information
- `rry` appears in 12 names and starts 0 of them
- `oostende` appears in 1 name (`BE Oostende Observer (RRY88)`) and starts none

So on that instance, typing what an operator actually remembers about a station returns an empty list. The fix is one word:

```js
item.dataset.obsName.includes(term)
```

That also matches the placeholder, which says "Search observers…", not "starts with".

**2. No test.** `AGENTS.md` is explicit: "Every change that touches logic MUST have tests. If you add new logic, add tests. No exceptions." `applyObserverSearchFilter` is logic, and a prefix-vs-substring bug is exactly what a test pins. The repo has jsdom-based frontend tests to copy from (`test-issue-1461-mobile-page-actions.js`, `test-mqtt-status-panel.js`); a case that seeds two observers, types a term matching the middle of one name, and asserts which rows stay visible would cover it.

Happy to merge once those two land.

efiten (approved):
Both points addressed, and the first one better than I asked for. Substring is the default now, with `^` as an opt-in for prefix-only narrowing — that keeps the callsign-prefix case someone will want without making it the behaviour everyone else trips over.

Checked out `a05445d6` and ran the new test here: 9 cases, all pass. It also does the thing that makes a test worth having — `extractApplyObserverSearchFilter` pulls the real function body out of `packets.js` and runs it against fake list items, so it fails if the production function changes rather than agreeing with a copy of it. The `^`-anchored, trimming, empty-term and case-insensitivity cases are all covered.

Approving. Merging once Go Build & Test finishes.

One nit, take it or leave it, not worth a push on its own: `background: var(--card-bg, #fff)` in `.multi-select-search-wrap`. `--card-bg` is defined at `public/style.css:343` and overridden in both dark blocks, so the `#fff` fallback is unreachable — and if it ever were reachable it would be white in dark mode. Every neighbouring rule in that block uses the bare form.

Apologies for the review noise on this one: you got several near-identical rounds of the same two points from me today, plus a workflow rerun that tested a stale base. That was my end, not yours. This is the last word on it.

efiten (changes requested):
Retracting my approval from an hour ago. A deeper pass over the same code found a bug I missed, and it is the kind a user hits on the first try. My earlier round only looked at the matching logic; this one looked at what the new input does to the handlers already on that menu.

**1. The search input's `change` event bubbles into the observer-selection handler, and eats the first click on an observer.**

`#observerSearchInput` is a child of `#observerMenu`, and `obsMenu.addEventListener('change', ...)` at `public/packets.js:1769` reads `e.target.dataset.obsId` with no guard. For the text input that is `undefined`, so the handler runs the whole else branch, writes localStorage, and calls `buildObserverMenu()` — which replaces `obsList.innerHTML` — plus `updatePacketsUrl()` and `renderTableRows()`.

The damage is the timing. Clicking an observer label moves focus off the input on mousedown, so `change` fires and the list is rebuilt before mouseup: the checkbox the user pressed no longer exists.

Reproduced in a browser, minimal page with the same DOM shape and the same two listeners, typing "brussels" and then clicking the Brussels checkbox once:

```
event log:        ["rebuild","change handler id=undefined","rebuild","renderTableRows"]
after one click:  ["__all__=true","a=false","b=false"]
```

`a=false` is the observer the click landed on. The selection is silently dropped and the user has to click again. A milder version of the same fires on every blur or Enter in the search box: a full table re-render for no state change.

One line fixes it:

```js
obsMenu.addEventListener('change', (e) => {
  const id = e.target.dataset.obsId;
  if (!id) return;
```

or bind the listener to `obsList` instead of `obsMenu`.

**2. Clear Filters does not clear the search box.** The handler at `public/packets.js:1936` resets `fHash`, `fNode`, `packetFilterInput`, `fChannel`, `fTimeWindow`, both menus' checkboxes and the region filter, but not `#observerSearchInput`, and it never re-applies the filter. So: type "brussels", close the dropdown, click Clear, reopen — every non-matching observer is still `display:none` while the UI says no filters are active. Clearing the input and calling `applyObserverSearchFilter()` in the same handler covers it.

**3. Smaller: `obsSearchInput.focus()` on open raises the on-screen keyboard on touch devices**, over the list the user is about to tap. Worth gating on a pointer/width check, the way the nav drawer does at `>768px`.

None of the three is covered by `test-observer-search-filter.js`, which is not a criticism of the test — it tests the matcher, and the matcher is correct. These are interaction bugs, so they need the jsdom style the repo uses elsewhere, or an E2E case.

The substring/`^` work from the last round stands and is good. Sorry for the approve-then-retract; that was me reviewing the diff instead of the behaviour.

efiten (approved):
All three points from my last round are fixed, and I checked the behaviour rather than the diff. I built the server from `2e8e3217`, ran it against a migrated copy of the e2e fixture (31 observers) and drove it in headless Chromium. Then I ran the same script against a copy of `packets.js` with only your three fixes reverted.

1. **Change event from the search box.** With `repea` typed (5 of 31 visible), one click on an observer checkbox now selects it: the checkbox stays checked, the URL gets `observer=` and the trigger updates. Enter and blur in the search box cause 0 list rebuilds. With the guard at `public/packets.js:1778` removed, the same click leaves the checkbox unchecked, and Enter plus blur cause 1 rebuild.
2. **Clear Filters.** With a search that matches nothing, Clear and reopen gives an empty box and all 31 observers. Without `packets.js:1948-1949`, the term and the empty list stay.
3. **Touch focus.** In a mobile context where `(pointer: coarse)` matches, opening the menu no longer focuses the input. Without the check, it does. Desktop still focuses it.

`test-observer-menu-interactions.js` passes 5 of 5 here and fails 3 of 5 against the reverted copy, one per point, so it tests what it says. `test-observer-search-filter.js` passes 9 of 9 and `check-css-vars.js` is clean.

**One thing about CI.** Both new test files are only added to `test-all.sh`, which no workflow runs. The unit step in `.github/workflows/deploy.yml` keeps its own list and neither file is in it, so as it stands neither test runs in CI. #1965 makes `test-all.sh` the single runner. I will merge #1965 first, so nothing more is needed here.

**Separate from this PR, already on master:** Clear Filters unchecks the observer checkboxes but does not empty `selectedObservers`, so select A, Clear, select B ends with both A and B in the URL. Not something for this PR to fix.

Not checked: a physical touch device (only Chromium emulation), Firefox or Safari, and the full Playwright E2E suite.

PR #1965, opened 2026-09-05
test: unify frontend test runs and prevent inventory drift

### LINKED ISSUE #1858: test: `npm test` is red on master — `test-all.sh` and the CI test list have diverged
## Summary

`npm test` (i.e. `sh test-all.sh`) fails on a clean checkout of `master`. 9 of the 53 tests it runs are red — and **all 9 are tests that CI does not run**. `test-all.sh` is never invoked by any workflow, so the failures are invisible.

This is not a one-off breakage. It is the predictable result of the repo maintaining three independent, hand-curated lists of which tests exist:

| List | Count |
|---|---|
| `test-*.js` files in the repo | 267 |
| `node test-*.js` lines in `.github/workflows/deploy.yml` | 159 |
| `node test-*.js` lines in `test-all.sh` (= `npm test`) | 53 |

`grep -n "test-all\|npm test" .github/workflows/deploy.yml` → no match. The CI job re-lists 159 tests by hand instead. Anything that lives only in `test-all.sh` rots silently; anything in neither list (267 − union) is dead weight nobody runs at all.

## Reproduction

```
git clone https://github.com/Kpa-clawbot/CoreScope.git && cd CoreScope
npm ci
sh test-all.sh
```

Verified against `a06ac8ac` (master, 2026-07-18).

## The 9 failing tests

None of these appear in `deploy.yml`:

| Test | Failure |
|---|---|
| `test-frontend-helpers.js` | 2 of 627 assertions — `favStar` expects `★`/`☆` |
| `test-channel-psk-ux.js` | expects `🔓`/`🔒` row icons |
| `test-analytics-channels-integration.js` | sidebar link text assertion |
| `test-issue-1648-m6-final-sweep.js` | **1 real emoji-as-icon violation** (see below) |
| `test-issue-1648-m6-lint-self.js` | cascades from the above (`repo MUST be clean before anti-tautology probe`) |
| `test-issue-1438-customizer-mcrole.js` | `--mc-role-companion` empty |
| `test-issue-1446-cb-preset-cascade.js` | `--mc-role-repeater` empty |
| `test-issue-1470-node-tile-helper.js` | 3 of 6 — tile provider resolution |
| `test-issue-1485-live-anim-z.js` | `.addTo(animLayer)` call-site count |

## Triage: 8 stale, 1 real defect

I traced each one rather than assume:

- **`favStar` / `test-channel-psk-ux` / `test-analytics-channels-integration`** — stale since `30627454` (#1648 M2, 2026-06-11). The emoji → Phosphor migration replaced `★`/`☆` with `<svg><use href="#ph-star-fill">`; the assertions were never updated. Red for five weeks.
- **`#1438` / `#1446`** — the colour-blind presets were deliberately moved into `style.css` (`body[data-cb-preset="deut"] { --mc-role-repeater: #FE6100; ... }`). Both tests still read inline custom properties off `root` in a `vm` sandbox that never loads CSS. The behaviour is correct; the tests describe the old design.
- **`#1470`** — asserts on tile provider id `voyager-inverted`, which was removed in `d7cd9203` (#1533) and is now `carto-voyager-dark`. `grep -rn "voyager-inverted" public/*.js` → no hits. Dead reference.
- **`#1485`** — asserts that the *source text* of `live.js` contains ≥3 `.addTo(animLayer)` call sites (2 today). Counting call sites is not a behavioural assertion; a refactor that consolidates two shapes into one loop breaks it without changing what renders.

That leaves one genuine product defect, filed separately: an emoji-as-icon leak in `public/rx-coverage.js:29`, which `test-issue-1648-m6-final-sweep.js` correctly catches. The gate built for #1648 works exactly as designed — it just isn't wired into CI, so the leak shipped.

## Why this matters beyond the 9

The `#1648` emoji gate is the clearest illustration: a deliberate, working regression gate that catches a real defect on `master` today, sitting in `test-all.sh`, not running anywhere. Every test outside the CI list is in that position.

It also costs contributor time. `test-all.sh` reads as the project's "run all tests" entry point and is what `npm test` maps to, so a new contributor runs it, sees 9 failures on a clean checkout, and has to work out which are theirs.

## Suggested fix

Make one list authoritative:

1. Have `deploy.yml` invoke `sh test-all.sh` (or a `test-all-ci.sh` split for browser-dependent tests) instead of 159 hand-maintained `node` lines.
2. Add a guard test that fails if any `test-*.js` in the repo root is absent from that script — otherwise the lists drift again within a release or two.
3. Fix or delete the 8 stale tests. The three #1648 ones want updating to the Phosphor markup; #1438/#1446 want rewriting against CSS rather than inline styles; #1470 wants the provider id corrected; #1485's call-site count is arguably worth dropping entirely in favour of the pane-binding assertions in the same file, which do test behaviour and do pass.

Happy to open the PR for any part of this if the direction is agreeable — I'd suggest starting with (1), since (3) is only durable once the list is enforced.

---

Investigated on a clean clone; Go tests for `internal/*` (lora, channel, packetpath, sigvalidate, prunequeue, dbconfig) all pass. I could not build `cmd/server` / `cmd/ingestor` in my environment (no module proxy access), so this covers the JS suite only.

n30nex:
Make `test-all.sh` the authoritative standalone frontend runner for npm and CI. Restore stale assertions and reject missing, duplicate, removed or undocumented inventory entries.

Fixes #1858.

Rebased onto `a2f039d4`. Retains release-routing, map scope-state and Scope Audit stylesheet tests, adds `test-packets-local-channels.js` to the sorted runner, and classifies `test-neighbor-map-btn-clip-e2e.js` under browser. Its separate CI browser step is preserved. Inventory: 280 root suites, 167 standalone, 113 requiring separate setup.

The icon repair fixes two suites red on master: `test-issue-1648-m2-emoji-scan.js` and `test-issue-1648-m6-final-sweep.js`. Node/live configured-scope confirmations now use the existing accessible Phosphor check sprite. Values, visibility conditions and scanner assertions are preserved.

- Red evidence: `89e45a9` inventory assertions; `4e255df` accessible confirmation assertion. The latest rebase also reproduced both unclassified-file failures before adding their entries. This follow-up only changes runner/classification configuration and counts; no test files modified.
- Local validation: all 167 standalone suites; 27 M2 browser checks and 16 neighbor geometry checks in Chromium. Syntax, whitespace, PII, CSS-variable and XSS checks passed.
- Browser coverage includes populated/empty/null configured scopes in node and live views.
- No new dependencies, requests, application settings or Go changes. Workflow outside the unit step matches master.
- Windows validation uses process-local UTF-8 settings. Encoding and node-reach confirmation follow-ups remain separate, as requested.

## Preflight override

External `run-all.sh` is unavailable; applicable repository checks were run directly.


efiten (changes requested):
Reviewed, and I checked the repairs rather than taking them on trust. Checked out the branch and ran a sample: `test-map-clustering.js` passes here and fails on master with the stale `>3<` pill assertion, which is the exact staleness this PR is about. `test-channel-colors.js`, `test-live-anims.js` and `test-test-inventory.js` pass too; the inventory reports 275 root suites classified, 163 run.

The direction is right and several of the "restored" assertions are stronger than what they replace, not weaker: `test-channel-colors.js` moves from two `includes()` checks to an exact style string, and `test-map-clustering.js` from a bare `>3<` substring to `class="mc-pill role-repeater"[^>]*>R3<`. `test-live-anims.js` trades source greps for behaviour (`doesNotThrow` on the real function), which is the better trade. The guard self-tests its own checker before running it on the tree, so a broken guard cannot pass silently.

**One blocker: it needs a rebase, and the guard makes that non-optional.**

`checkInventory` is exhaustive over every root `test-*.js`, so any test file added to master since this branch was cut fails it. Merging current master into the branch locally:

- conflict in `test-all.sh` (master moved it in #1964)
- `test-issue-1956-release-routing.js` exists on master, is absent from this branch, and is in neither `test-all.sh` nor `scripts/non-unit-tests.json`, so the guard fails on the merged tree

Green CI here is green against this branch's base, not against master. That is the design working as intended, but it has to be resolved before merge.

**Merge-order note.** The guard asserts `doesNotMatch(unitStep, /\bnode test-/)`, so the CI unit step may not carry a second list. #2005 and #2006 each add a `node test-…` line to that step, because they were written against today's master. If this lands first I will rebase both to drop those lines and rely on `test-all.sh` alone; if they land first, this branch picks up two more files to classify. Either way it is mechanical, but it cannot be both at once.

**Nit, not a blocker.** The `test-preflight-xss-gate.js` change says relative paths are there so Windows drive-letter colons are not mistaken for separators. That part is right, but the suite still fails on Windows for a different reason: the preflight script's Python reader writes non-ASCII to a cp1252 console and dies with `UnicodeEncodeError: 'charmap' codec can't encode characters in position 0-1`. Measured on this branch and on master, same failure. Worth a follow-up if Windows is meant to be a supported dev environment; it does not affect CI.

Rebase and I will re-check the guard against master before merging.

efiten (commented):
Re-checked `f31fe7bb` against current master (`a2f039d4`). The branch conflicts again, in 1 block in `test-all.sh` and 1 in `.github/workflows/deploy.yml`, both from #2009, which added `test-packets-local-channels.js` to the old CI unit list and to `test-all.sh`.

I merged master into the branch locally and kept your side of both blocks, with `test-packets-local-channels.js` added to the sorted runner list. The guard then failed, correctly:

```
AssertionError [ERR_ASSERTION]: unclassified root tests: test-neighbor-map-btn-clip-e2e.js
```

That file came in with #2010 and belongs under `browser` in `scripts/non-unit-tests.json`. With that 1 entry added, the guard reports `280 root suites classified; 167 run by local and CI unit checks`. I ran all 167 suites one by one: 166 pass, and the 1 failure is `test-preflight-xss-gate.js` with the Windows cp1252 `UnicodeEncodeError` from last round, which also fails on master.

**The icon change is right, and it is needed rather than scope creep.** The bare `✓` in `public/nodes.js:738` and `public/live.js:2632` came in with #1971. With master's copies of those 2 files, `test-issue-1648-m2-emoji-scan.js` and `test-issue-1648-m6-final-sweep.js` both fail. Master's `test-all.sh` already runs both, but master's CI does not run `test-all.sh`, so it went unnoticed. Once this PR makes CI run the runner, the PR cannot be green without this fix. `ph-check` exists in the sprite (`public/icons/phosphor-sprite.svg:22`), `--status-green-text` is defined for both themes (`public/style.css:258`, `:388`), and `role="img"` gives the `aria-label` an element it actually applies to. Cache busting is automatic, so no bump is needed. Could you add a line to the description saying the PR repairs 2 suites that are red on master today?

Follow-up, not for this PR: `public/node-reach.js:84` still renders the same confirmation as a bare `✓`. No scan catches it because the whole file is allowlisted (`tests/emoji-allowlist.txt:67`) for its legend glyphs.

CI on this head is green (run 34651301756), and the E2E log shows all 6 configured-scope cases passing. That is green against `51a2a7dd`, not against current master.

Not checked: the E2E job against the merged tree, and a run of `test-all.sh` on Linux. Both happen on the next CI run after the rebase.

Rebase, classify `test-neighbor-map-btn-clip-e2e.js`, and I will approve.

efiten (approved):
Re-checked `a2fe0d8f`: it is based on current master `a2f039d4`, `node test-test-inventory.js` reports 280 root suites classified and 167 run, `test-neighbor-map-btn-clip-e2e.js` is classified under browser, `test-packets-local-channels.js` is in the runner, and `test-issue-1648-m2-emoji-scan.js` passes. CI is green on this head. Approving; merging now. The open PRs that still add `node test-` lines to the deploy.yml unit step will be rebased by me to rely on `test-all.sh` alone, as promised in my earlier review.

Issue #1901, opened 2026-08-17
bug(db): detectSchema() silently swallows PRAGMA failure — server runs v2 SQL against a v3 DB until restart (empty Packets, /channels/*/messages 500)


MarekWo:
## Summary

If `detectSchema()`'s `PRAGMA table_info(observations)` fails for any transient reason at startup, the server silently caches `isV3 = false` and runs the **entire session** in v2-schema mode against a v3 database. Every read path that branches on `isV3` then emits SQL against columns that do not exist:

```
[store] ingest query error: SQL logic error: no such column: o.observer_id
[store] ingest observations query error: SQL logic error: no such column: o.observer_id
```

User-visible effect: **Packets page shows 0 packets** and **`/api/channels/<name>/messages` returns 500**, while the database is perfectly healthy and the ingestor keeps writing normally. Nothing recovers it — the flag is never re-checked — so it persists until someone restarts the server process.

Observed in production on v3.9.1 (2026-08-17): 2.5 hours of a fully unusable read API, ended by a plain container restart with no change to the database. `detectSchema()` is unchanged on current master (`v3.9.2-13-g1476b857`).

## Root cause

`cmd/server/db.go:72-76`:

```go
func (db *DB) detectSchema() {
	rows, err := db.conn.Query("PRAGMA table_info(observations)")
	if err != nil {
		return          // <-- isV3 stays false. No log, no retry, no error propagated.
	}
```

`isV3` is set **only** here. `cmd/server/store.go` (and ~14 sites in `db.go`) branch on it:

```go
if s.db.isV3 {
    // ... LEFT JOIN observers obs ON obs.rowid = o.observer_idx      (correct for v3)
} else {
    // ... LEFT JOIN observers obs ON obs.id = o.observer_id          (v2 — column absent on v3)
}
```

So a single transient failure of one PRAGMA silently flips the whole server onto the wrong query set for as long as the process lives.

Two properties make this exceptionally hard to diagnose:

1. **The failure is completely silent.** Startup logs contain no line stating which schema mode was selected, so a clean-looking startup log is not evidence that detection succeeded. The only signal is the `no such column` spam that begins afterwards.
2. **The database looks fine, because it is fine.** `PRAGMA quick_check` returns `ok`, the `packets_v` view is intact, and `SELECT observer_id FROM packets_v LIMIT 1` works. This actively misleads toward "restore a backup", which is exactly the wrong move.

[...]

PR #1992, opened 2026-09-09
perf(sqlite): swap modernc.org/sqlite for mattn/go-sqlite3, cross-built with zig

sylr:
Swaps the SQLite driver from `modernc.org/sqlite` (pure Go, SQLite 3.46.0) to `github.com/mattn/go-sqlite3` (cgo, bundled SQLite 3.53.4), and pays the resulting cross-compilation cost with `zig cc`.

Draft because the riskiest part of this deletes rows — see [Please review this part first](#please-review-this-part-first) — and because three things remain unverified at the bottom.

`modernc.org/sqlite` is a transpilation of the C amalgamation. This repo is read-heavy: `cmd/server` chunk-loads a graph at startup and fans out neighbour/topology/analytics queries per request, and it pays for that transpilation on exactly those paths. Head-to-head on the same 120k-transmission / 240k-observation database, running our own hot-path SQL under both drivers (Apple M4, `-count=5`, medians):

| workload | modernc | mattn | |
|---|---:|---:|---|
| chunk load (`chunked_load.go` v3 join, 20k tx) | 449ms | 196ms | **2.3×** |
| aggregate scan (240k-row join + `GROUP BY`) | 276ms | 137ms | **2.0×** |
| 1500 prepared-statement lookups | 512ms | 403ms | **1.3×** |

Allocations fall with it: 1.12M vs 1.64M allocs and 21MB vs 30MB on the chunk load.

**Superseded by a production run.** @efiten measured both drivers on a real instance — 11,077,038 observations, 9.7GB database, 4-core arm64 — as server-only containers against the same live volume, one at a time, with round 2 reversing the order so the page cache favours the old driver:

| | audit 7d | audit 24h | background fill (13 chunks) | start → /api/health |
|---|---:|---:|---:|---:|
| modernc, round 1 | 16.67s | 2.27s | 130.2s | 16.6s |
| mattn, round 1 | 7.87s | 1.34s | 93.8s | 13.5s |
| mattn, round 2 | 8.15s | 1.35s | 96.4s | 13.0s |
| modernc, round 2 | 13.46s | 2.29s | 137.8s | 15.5s |

Warm, the old driver improves to 13.46s on the 7d audit and still loses by ~1.8×. Chunk load is ~1.4×. `/api/nodes?limit=500` is 0.039s against 0.037s — nothing.

**So the real gain is ~1.4–1.8× on the paths that matter, not 2–2.3×.** The shape the harness predicted holds — scans and joins gain, small lookups do not — which is more reassuring than the magnitude would have been. Quote these numbers.

**The counterweight**, cold and native on that machine: a build goes from **52s to 163s**. An instance that builds its own image pays that per deploy.

## The build is cgo now, and one thing about that is a trap

**`CGO_ENABLED=0` still builds.** mattn links a stub, and the binary dies on its first query with `go-sqlite3 requires cgo to work. This is a stub`. A green build is not evidence of anything here, which is why `AGENTS.md` now says so explicitly. `GOOS=linux go build` genuinely cannot cross-compile any more.

A new root `Makefile` is the entry point. `make crossbuild` uses `zig cc -target {x86_64,aarch64}-linux-musl` and links static, so each artifact stays a single self-contained file and the `alpine:3.20` runtime no longer depends on the base image's libc at all.

`-Wl,-s` is load-bearing: Go's own `-s -w` does not reach the musl objects zig links in, and without it the server binary is 19.8MB instead of 12.1MB.

The Dockerfile keeps its single `$BUILDPLATFORM` builder — still no QEMU for compilation — and gains a checksum-pinned zig plus BuildKit cache mounts. The mounts are not a nicety: without them an image build recompiles the amalgamation from cold and takes over half an hour.

## Please review this part first

`internal/dbschema/dedup_index.go` **deletes observation rows**. It is the one part of this change that can lose data, and it exists because the migration exposed a real bug rather than causing one.

`stmtInsertObservation` resolves its `ON CONFLICT` against `idx_observations_dedup`, which `cmd/ingestor/db.go` only ever created inside the branch that creates the `observations` table for the first time. Any database whose table predates that branch never got one, so the UPSERT had no conflict target. modernc failed on the first insert; mattn fails at `OpenStore`. Same bug, found earlier.

Creating the index unconditionally repairs it — but the index is what was supposed to prevent duplicates, so a database that never had it can already hold rows violating it. **`test-fixtures/e2e-fixture.db` in this repo holds one.** So duplicates are collapsed first. Refusing is not the safer option: without the index the ingestor cannot prepare its UPSERT, so it cannot start at all.

Replaying that UPSERT faithfully is subtler than it looks, and a first cut of this got it wrong twice:

- `COALESCE(excluded.x, x)` means the **incoming** value wins, so down a group in id order the survivor keeps the **last** non-NULL value. Taking the first silently discarded newer readings.
- The UPSERT names exactly five columns (`snr`, `rssi`, `score`, `raw_hex`, `resolved_path`). Every other column must keep the surviving row's own value; merging those too invents history the ingestor would never have written.

Merge, delete and `CREATE UNIQUE INDEX` now share one transaction. Split apart, a writer inserting a duplicate in the gap fails the index creation while leaving the deletions committed — rows destroyed and no index to show for it.

Cost, measured on 2.4M synthetic rows holding 5 duplicates: **4.1s**, holding the write lock throughout, once, at ingestor startup before MQTT subscribe. Materialising the duplicate-group scan once rather than per column took that from 9.7s; the pathological case (400k of 600k rows duplicated) is 5.7s, slightly worse than the 4.2s it was before that change.

## Four more behavioural differences

Full detail in `docs/sqlite-driver-migration.md`. Briefly:

**Statement preparation is eager.** modernc's `newStmt` stored the SQL and compiled lazily; mattn calls `sqlite3_prepare_v2` inside `Prepare`, so SQL naming a missing table fails at *open*. 59 server tests failed on this alone, all fixtures with partial schemas. `OpenDB` keeps failing loudly (#1901; `main.go` gates on `dbschema.AssertReady` anyway) and the fixtures now declare what they are prepared against via `ensurePreparable`. This also exposed nine `nodes(pubkey …)` declarations across seven files, where production has only ever had `public_key` — lazy compilation had hidden the mismatch for as long as it existed.

**`synchronous` silently dropped FULL → NORMAL.** mattn defaults it to NORMAL and executes the pragma unconditionally, where SQLite's own default (what modernc left alone) is FULL. In WAL mode that weakens durability under power loss. Pinned in `dbschema.WriterDSN`, which both writers now share — `cmd/migrate` kept a bare path at first and so quietly wrote at NORMAL, which is what a second copy of a DSN buys you.

**The DSN dialects are mutually invisible.** modernc understood only `_pragma=name(value)`, mattn only `_`-prefixed parameters, and neither errors on the other's form — a driver-only rename would have dropped every pragma in silence. `_journal_mode=WAL` is also gone from the server's read handle: modernc ignored it, mattn honours it, and setting `journal_mode` on a read-only connection is a write. Dropping `_busy_timeout` with it costs nothing, since mattn already defaults to 5000ms — which means the read handle finally *gets* the busy timeout it had silently lacked.

**`mode=ro` survives for a non-obvious reason.** mattn always passes `READWRITE|CREATE` and its amalgamation has `SQLITE_USE_URI=0`; what makes the URI work is its C wrapper ORing `SQLITE_OPEN_URI` in. So the #1283/#1289 invariant holds with no build flags — but it depends on the `file:` prefix. `cmd/decrypt` had been building its DSN without one, so its `mode=ro` had never applied and a missing path was created read-write. Fixed in passing; never a migration regression.

## What did not change

No modernc-specific API was in use: no `RegisterFunction`, no `*sqlite.Conn`, no `sqlite/lib` error constants, no `sql.Register`. No `time.Time` is ever bound as a query argument, so driver time handling is not in play. Both drivers convert declared `DATE`/`DATETIME`/`TIMESTAMP` columns to `time.Time`, so `/api/dropped-packets` keeps emitting `dropped_at` as RFC3339 — an earlier draft "fixed" that with a `CAST` and would have been the regression.

## Tests and CI

New regression tests, each written because something got through without it:

- `TestEnsureObservationsDedupIndexKeepsLatestValues` — the merge ordering. The original test used complementary NULLs, which passes whichever direction you pick, which is why the bug survived it.
- `TestCollapseDuplicatesAndIndexIsAtomic` — a failed index creation must roll the deletions back.
- `TestOpenStorePragmas` / `TestWriterDSNPragmas` — every writer pragma, read back through the store's own connection. A separate `sqlite3` session or the startup log line would prove nothing.
- `TestOpenDBRefusesMissingDatabase` — the read-only invariant, which now rests on a detail of the driver's C wrapper.
- `TestEnsurePreparableMatchesPrepareStatements` — fails when a new prepared statement outgrows the fixture helper.

CI gains test execution for `cmd/migrate` and `internal/dbschema`, which had none and both open the database. A PR-time two-arch build plus an arm64 QEMU smoke gate is new: the GHCR push is push/tag-only, so without it nothing on a PR would exercise zig, static musl linking or arm64, and the first signal would arrive on master. `cache-dependency-path` widens from 2 of the 5 tracked `go.sum` files to all of them.

`make test` passes across all 14 modules, `cmd/server` also under `-race -count=2` with no failures and no races. `gofmt` and `go vet` clean. Release-routing and Dockerfile COPY-invariant gates pass.

## Verified by running

- All 8 cross-builds static and correct-architecture; both arches of the container image built, exported and run under QEMU, serving `/api/health` and `/api/nodes` against a 2.9M-observation production snapshot.
- The `migrate` binary repairing that snapshot's duplicate on bare Alpine.
- `CGO_ENABLED=0` producing a binary that builds and then fails on first query.

## Not verified

- ~~The 2–2.3× figures come from a standalone harness, not this load under the old driver.~~ **Closed** by @efiten's production run above, which also corrected the multiplier.
- SQLite 3.46.0 → 3.53.4 query-planner differences on queries with no total `ORDER BY`.
- Sustained live ingest through the new writer DSN, and the duplicate collapse against a database an ingestor is actively writing to. Verified against a static snapshot only, and the collapse is measured at 4.1s on 2.4M synthetic rows with 5 duplicates — well short of an 11M-row instance. @efiten has offered a staging instance taking real MQTT traffic; **this is the item to close before the PR leaves draft.**

An earlier revision of this branch shipped the dedup merge in the wrong direction with a green test suite, and review then found three more things in the same file: the repair gated on an error string, a non-atomic TEMP table drop aimed at the wrong connection, and a deletion whose only record was a row count. All fixed in ac7e8d38. Passing tests did not establish safety here, which is why the deletion path wanted a second pair of eyes rather than a rubber stamp.


efiten (commented):
Read the deletion path first, as you asked. It is a draft so I am not merging it; this is the second pair of eyes on `internal/dbschema/dedup_index.go`.

**The merge direction is right, and for the right reason.** `excluded` is the incoming row, so a later non-NULL value replaces an earlier one, and replaying that down a group means the last non-NULL wins. `ORDER BY o2.id DESC LIMIT 1` with `IS NOT NULL` in the predicate does exactly that, and returns NULL when the whole group is NULL, which is what the UPSERT leaves behind too. Restricting the merge to the five columns in the `DO UPDATE SET` list is the part I would have got wrong: merging `direction`, `timestamp` or `path_json` as well would invent rows the ingestor never wrote.

Putting `CREATE UNIQUE INDEX` inside the same transaction as the `DELETE` is the correct call, and the failure mode you describe (rows destroyed, no index) is real.

Three things.

**1. The fallback hangs on an error string, in the one PR that changes what produces it.** `ensureObservationsDedupIndex` tries the index, then decides whether to repair by `strings.Contains(err.Error(), "UNIQUE constraint failed")`. Everything downstream, including the deletion, is gated on that match. This PR swaps the driver underneath it. If mattn ever words that message differently, or wraps it, the repair silently does not run and `OpenStore` fails instead — the failure is loud, but the diagnosis is not.

`mattn/go-sqlite3` exposes `sqlite3.Error` with `Code` / `ExtendedCode`, so the check can be on `ErrConstraint` rather than on prose. Failing that, a test that creates the index over known duplicates under the new driver and asserts the repair path is taken would pin it. Right now `TestCollapseDuplicatesAndIndexIsAtomic` exercises the repair directly, so nothing covers the branch that decides to call it.

**2. The deferred cleanup does not do what its comment says.** The comment on `DROP TABLE IF EXISTS temp.dedup_groups` says the connection returns to the pool clean, but the deferred drop runs `rw.Exec`, which takes whatever connection the pool hands out, while the temp table belongs to the one `tx` held. After commit that connection can go back to the pool still carrying `dedup_groups`.

The code is correct anyway, because the in-transaction `DROP ... IF EXISTS` before `CREATE` handles a dirty connection. So this is the comment claiming a guarantee that the pre-drop is actually providing. Worth fixing the comment rather than the code.

**3. The repair deletes rows at startup with a count as the only record.** I agree with your reasoning that refusing is not safer: without the index the UPSERT cannot prepare, so the ingestor cannot run at all. But `collapsed N duplicate observation row(s)` is everything an operator gets, and if the merge direction is ever wrong again, that line is all there is to reconstruct from. The group keys are already sitting in `dedup_groups` at that point; logging them (or the first N) costs nothing and makes the operation auditable after the fact.

On the three unverified items at the bottom of your description: sustained live ingest through the new writer DSN is the one I would want closed before this leaves draft, and it is the one I can help with — I can run this branch on a staging instance taking real MQTT traffic, with the duplicate collapse against a database an ingestor is actively writing to, which is the case your static snapshot cannot cover. Say the word.

The `synchronous` FULL → NORMAL catch is the kind of thing that would have gone unnoticed for a year. Good find.

Issue #1340, opened 2026-05-24
feat(observability): SQLite write-lock wait/hold instrumentation per component — make starvation detectable


Kpa-clawbot:
## Summary
The #1339 root cause (neighbor-builder write-lock starvation) was invisible to operators for ~3 days. We need first-class instrumentation for SQLite write-lock acquisition latency + duration so the next regression of this class is detectable from operator metrics.

## Why this matters now
- #1289 added a periodic builder that silently held the write lock for 2+ minutes
- No metric exposed lock-wait time, lock-hold time, or contention count
- The symptom (low ingest rate) looked like "broker not sending" until DB-level investigation
- paho's silent drops + 5s busy_timeout swallowed the actual SQLite contention

## What to add
Wrap every `Exec`/`tx.Begin` on the writer connection in instrumentation:

1. **wait_ms histogram**: time spent waiting for the write lock before the query actually starts
2. **hold_ms histogram**: time the writer holds the lock during the query
3. **contention_total counter**: count of writes that had to wait > 100ms
4. **slow_writer_log**: log writer name + duration whenever hold_ms > 500ms (configurable)

Expose via the existing `/api/perf` endpoint structure (already has per-component metrics from #1123).

## Per-writer attribution
The infra to distinguish writers exists in #1123's per-component IO metrics. Use the same component tags:
- `neighbor_builder`
- `mqtt_handler` (transmission/observation inserts)
- `prune_packets`, `prune_observers`, `prune_metrics`
- `mbcap_persist`
- `vacuum`

## Acceptance
- [ ] `/api/perf` includes `db.writer.wait_ms_p50/p95/p99` and `db.writer.hold_ms_p50/p95/p99` per component tag
- [ ] Log on hold_ms > 500ms: `[db-slow-writer] component=X duration=Y query=<truncated>`
- [ ] Test: synthetic 60s-blocking query → wait_ms > 50000 for subsequent writes; log line emitted
- [ ] Mutation: revert the instrumentation → metrics gone, test fails
- [ ] CI gate: add a smoke test that asserts neighbor-builder hold_ms < 1000ms on a fixture with 100k observations

## Out of scope
- Replacing SQLite with a multi-writer DB
- Per-statement timing (too granular)
- Cross-process tracing

## References
- #1339 (the bug that motivated this)
- #1283/#1287/#1289 server-readonly refactor — introduced the new ingestor write paths
- #1123 per-component disk IO metrics
[...]

PR #2000, opened 2026-09-10
fix(ingestor): delete aged packets in bounded batches so prune stops stalling ingest

liquidraver:
## Problem

`PruneOldPackets` deletes an entire retention day inside a **single** `WriterTx`. `writerMu` serialises every wrapped writer call, so for the full duration of that delete the MQTT ingest path is blocked.

On a production instance running v3.10.1 (`retention.packetDays: 10`, ~16k transmissions and ~260k observations/day, 3 MQTT sources), `/api/perf/write-sources` reported:

```
prune_packets    hold_ms_p50 33905   hold_ms_max 38798
mqtt_handler     wait_ms_max 35797
prune_metrics    wait_ms_p50 34015
prune_observers  wait_ms_p50 29727
```

A ~35 second ingest stall once every 24 hours, with the two other daily maintenance writers queued behind it. It is a stall rather than data loss — MQTT buffers behind it — but the Live view freezes for the duration, and the `ingest-buffer` is finite.

The instrumentation from #1340 is what made this visible; the ownership move in #1283 is why the prune lives on the ingestor's writer in the first place.

## Fix

Delete in bounded batches of 250 transmissions, one transaction each, releasing the lock between batches.

Go's `sync.Mutex` enters starvation mode once a waiter has been blocked for 1 ms and hands off FIFO, so an ingest goroutine queued behind the prune is served at the **next batch boundary** rather than after the whole retention day. No sleep between batches is needed.

Both statements of a batch are built from one shared subquery, `pruneAgedTransmissionIDs`, so they resolve the same bounded set: nothing modifies `transmissions` between them inside the transaction.

That subquery orders by **`first_seen, id`**. The ordering is satisfied straight off `idx_transmissions_first_seen` — the index carries the rowid as its tiebreaker — so the `LIMIT` is deterministic even on tied timestamps, and the planner never falls back to a table scan. An earlier revision ordered by `id` alone, which silently dropped the index; see *Review follow-up* below.

The loop ends on a **short batch**: fewer rows than the limit already proves nothing is left below the cutoff. Only an exact multiple of the batch size pays for one extra, empty batch.

## This is a latency fix, not a throughput fix

**A real prune** — 2000 transmissions / 32,000 observations, all aged out (median of 7 runs):

| | txns | total | **hold_ms_max** |
|---|---:|---:|---:|
| master (single transaction) | 1 | 168.2 ms | **168.2** |
| this PR | 9 | 182.6 ms | **26.3** |

About **9% more total wall time** for a **6.4x shorter worst-case hold** (168.2 ms → 26.3 ms). The 9 transactions are 8 full batches plus one empty one, because 2000 is an exact multiple of the batch size. I want to be explicit that the win is entirely in how long the writer lock is held at once, not in total work.

At production scale the hold shrinks in proportion to the batch: the measured 33.9 s day becomes roughly **0.53 s per batch across 64 batches**.

**The no-op prune** — 1M transmissions, nothing aged out, which is the state most instances are in most days (median of 9 runs, warm cache):

| | txns | total | **hold_ms_max** |
|---|---:|---:|---:|
| master (single transaction) | 1 | 30 µs | **0.033** |
| first revision (`ORDER BY id`) | 1 | 152.1 ms | **152.1** |
| this PR | 1 | 80 µs | **0.076** |

The first revision's 152.1 ms is two full-table scans of roughly 76 ms each — one per statement. The fixed no-op costs 80 µs against master's 30 µs because it issues two statements instead of one, which is negligible under `writerMu`.

## Behaviour change

On error the count from already-committed batches is now returned alongside it, rather than `0` — those rows really are gone, so reporting 0 would be wrong. Both callers in `main.go` (startup prune and the 24 h ticker) discard `n` when `err != nil`, so nothing observable changes.

## Tests

- **`TestPruneOldPacketsDeletesInBoundedBatches`** — the perf assertion required by AGENTS.md rule 0. It counts `prune_packets` transactions through the #1340 writer instrumentation (2 full batches + 1 partial = 3) and fails against the pre-patch single-transaction implementation with `got 1`.
- **`TestPruneAgedTransmissionIDsUsesFirstSeenIndex`** — pins the query plan of the subquery and of both DELETE statements: uses `idx_transmissions_first_seen`, no `SCAN transmissions`, no temp b-tree sort. Transaction counts cannot see an index being dropped, so the plan is the assertion. Verified by flipping the constant back to `ORDER BY id`, which fails on all three statements while every count-based test still passes:

  ```
  batch subquery: plan scans transmissions, so the empty terminating batch walks the whole table under writerMu: SCAN transmissions
  observations delete: plan does not use idx_transmissions_first_seen: ... | LIST SUBQUERY 1 | SCAN transmissions
  transmissions delete: plan does not use idx_transmissions_first_seen: ... | LIST SUBQUERY 1 | SCAN transmissions | ...
  --- FAIL: TestPruneAgedTransmissionIDsUsesFirstSeenIndex
  ```

- **`TestPruneOldPacketsExactMultipleTerminates`** — the one case the short-batch exit cannot catch on its own.
- A cutoff spanning several batches keeps fresh rows and leaves no orphaned observations; the `days <= 0` short-circuit takes no writer lock at all; the steady state with nothing aged out runs exactly one empty batch.

The pre-existing `TestIngestorPruneOldPackets` (#1283) passes unchanged.

Validated on current `origin/master` (including #1993, #1994 and #2002) plus this change:

```
go build ./...        OK
go vet ./...          OK
gofmt -l              clean
go test -race ./...   ok  github.com/corescope/ingestor  492.737s   (0 data races)
```

Go 1.27.1, matching the Dockerfile and CI.

## Note on the batch size

`pruneBatchTransmissions` is a package constant at 250. The bound is on transmissions, but hold time scales with the rows actually deleted — at ~16 observations per transmission that is ~4k deletes per batch — so an instance with a denser observation ratio gets a proportionally longer hold from the same batch size. The doc comment says so. Per AGENTS.md rule 8 this is a candidate for `config.json` later, alongside `db.incrementalVacuumPages`; I have deliberately not added config surface in this PR.

## Review follow-up

Addressed in `ecf0b371`, from @efiten's review, which measured the problem on a 1M-transmission / 11M-observation production database:

1. `ORDER BY id` → `ORDER BY first_seen, id`, so the batch keeps the covering index. Reproduced independently at 1M rows: the empty terminating batch went from **73.22 ms** (`SCAN transmissions`) to **10 µs** (`SEARCH ... USING COVERING INDEX`).
2. The loop exits on a short batch instead of an empty one.
3. The doc comment now states that hold time scales with the observation ratio.

## Not addressed

Deleting a retention day also strands pages on the SQLite freelist faster than `db.incrementalVacuumPages` (default 1024) reclaims them. On the same instance the freelist sat at 160,509 pages / 627 MiB against a 2234 MB database. That is a separate concern and a config default question, not part of this change.

### DIFF
--- a/cmd/ingestor/maintenance.go
+++ b/cmd/ingestor/maintenance.go
@@ -10,42 +10,83 @@ import (
 	"github.com/meshcore-analyzer/dbschema"
 )
 
+// pruneBatchTransmissions bounds how many transmissions (and their child
+// observations) one prune transaction deletes before it commits and
+// releases the writer lock.
+//
+// Releasing between batches is the entire point. writerMu serialises every
+// wrapped writer call, so while the prune holds it the MQTT ingest path is
+// blocked. Go's sync.Mutex switches to FIFO handoff once a waiter has been
+// blocked for 1ms (starvation mode), so an ingest goroutine queued behind
+// the prune is served at the next batch boundary instead of after the whole
+// retention day.
+//
+// 250 keeps a batch near ~4k observation deletes at a typical ~16
+// observations per transmission — a few hundred milliseconds — while
+// keeping the number of commits low (a 16k-transmission day is 64
+// transactions, not 16k).
+const pruneBatchTransmissions = 250
+
 // PruneOldPackets deletes transmissions (and their child observations)
 // older than `days`. Returns count of transmissions deleted.
 //
 // Owned by the ingestor per #1283: the writer process is the only one
 // allowed to hold the DB write lock; previously this lived in
 // cmd/server/db.go and raced ingestor INSERTs (SQLITE_BUSY).
+//
+// Deletion is chunked into bounded transactions so ingest is never blocked
+// for longer than a single batch. Deleting a whole retention day in one
+// transaction held the writer lock for ~35s on a mesh doing ~260k
+// observations/day, stalling MQTT ingest for the same duration.
+//
+// On error the transmissions deleted by already-committed batches are
+// returned alongside it — those rows are gone, so reporting 0 would be
+// wrong.
 func (s *Store) PruneOldPackets(days int) (int64, error) {
 	if days <= 0 {
 		return 0, nil
 	}
 	cutoff := time.Now().UTC().AddDate(0, 0, -days).Format(time.RFC3339)
 
-	// Tagged for writer-perf visibility (#1340).
-	var n int64
-	err := s.WriterTx("prune_packets", func(tx *sql.Tx) error {
-		// Delete child observations first (no CASCADE in SQLite).
-		if _, err := tx.Exec(`DELETE FROM observations WHERE transmission_id IN (
-			SELECT id FROM transmissions WHERE first_seen < ?
-		)`, cutoff); err != nil {
-			return fmt.Errorf("prune observations: %w", err)
-		}
+	var total int64
+	for {
+		var batch int64
+		// Tagged for writer-perf visibility (#1340).
+		err := s.WriterTx("prune_packets", func(tx *sql.Tx) error {
+			// Both statements resolve the same bounded set: nothing modifies
+			// `transmissions` between them inside this transaction, and
+			// ORDER BY id makes the LIMIT deterministic.
+			//
+			// Delete child observations first (no CASCADE in SQLite).
+			if _, err := tx.Exec(`DELETE FROM observations WHERE transmission_id IN (
+				SELECT id FROM transmissions WHERE first_seen < ? ORDER BY id LIMIT ?
+			)`, cutoff, pruneBatchTransmissions); err != nil {
+				return fmt.Errorf("prune observations: %w", err)
+			}
 
-		res, err := tx.Exec(`DELETE FROM transmissions WHERE first_seen < ?`, cutoff)
+			res, err := tx.Exec(`DELETE FROM transmissions WHERE id IN (
+				SELECT id FROM transmissions WHERE first_seen < ? ORDER BY id LIMIT ?
+			)`, cutoff, pruneBatchTransmissions)
+			if err != nil {
+				return fmt.Errorf("prune transmissions: %w", err)
+			}
+			batch, _ = res.RowsAffected()
+			return nil
+		})
 		if err != nil {
-			return fmt.Errorf("prune transmissions: %w", err)
+			return total, err
 		}
-		n, _ = res.RowsAffected()
-		return nil
-	})
-	if err != nil {
-		return 0, err
+		// A batch that deleted nothing means no rows are left below the
+		// cutoff; every batch before it deleted exactly what it selected.
+		if batch == 0 {
+			break
+		}
+		total += batch
 	}
-	if n > 0 {
-		log.Printf("[prune] deleted %d transmissions older than %d days", n, days)
+	if total > 0 {
+		log.Printf("[prune] deleted %d transmissions older than %d days", total, days)
 	}
-	return n, nil
+	return total, nil
 }
 
 // PruneOldClientReceptions deletes mobile client-RX coverage rows older than
--- a/cmd/ingestor/maintenance.go
+++ b/cmd/ingestor/maintenance.go
@@ -21,11 +21,35 @@ import (
 // the prune is served at the next batch boundary instead of after the whole
 // retention day.
 //
-// 250 keeps a batch near ~4k observation deletes at a typical ~16
-// observations per transmission — a few hundred milliseconds — while
-// keeping the number of commits low (a 16k-transmission day is 64
-// transactions, not 16k).
+// The bound is on transmissions, but hold time scales with the rows actually
+// deleted, and each transmission carries an unbounded number of observations.
+// At ~16 observations per transmission a batch is ~4k row deletes and a few
+// hundred milliseconds; an instance with a denser observation ratio gets a
+// proportionally longer hold from the same batch size. 250 also keeps the
+// commit count low (a 16k-transmission day is 64 transactions, not 16k).
 const pruneBatchTransmissions = 250
+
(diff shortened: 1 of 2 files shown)


efiten (commented):
Checked this against a production instance (1,018,582 transmissions, 11M observations). The batching itself reads correctly to me: both statements resolve the same bounded set inside one transaction, the partial count on error is the right call, and the starvation-mode reasoning about the handoff at batch boundaries holds.

One measured problem, and it is in the `ORDER BY id`.

`transmissions` has `idx_transmissions_first_seen`. Adding `ORDER BY id` makes SQLite abandon it, because satisfying the ordering from the rowid walk is cheaper than sorting the index result. `EXPLAIN QUERY PLAN` on that database:

```
-- this PR
SELECT id FROM transmissions WHERE first_seen < ? ORDER BY id LIMIT 250;
`--SCAN transmissions

-- master
SELECT id FROM transmissions WHERE first_seen < ? LIMIT 250;
`--SEARCH transmissions USING COVERING INDEX idx_transmissions_first_seen (first_seen<?)
```

While rows are being deleted this costs nothing: the oldest rows have the lowest rowids, so the walk matches immediately (0.001s measured). The cost lands on the **terminating batch**, the one that finds nothing and breaks the loop. That one walks the whole table:

| query | empty batch | batch with rows |
|---|---:|---:|
| `ORDER BY id` | **0.269s** | 0.001s |
| `ORDER BY first_seen` | 0.000s | 0.000s |
| no ORDER BY (master's shape) | 0.000s | n/a |

Warm cache, 1M rows. It happens under `writerMu`, once per prune run, and it is the state most instances are in most of the time. This one is at `packetDays: 90` with nothing aged out yet, and `/api/perf/write-sources` reports `prune_packets hold_ms_max 0.468` today. Under this patch that same no-op prune would hold the writer for roughly half a second: two of those scans, one per statement.

Half a second against the 35s you are removing is a good trade, so this is not an objection to the approach. But it is avoidable: **`ORDER BY first_seen` keeps the covering index and keeps the LIMIT deterministic**, which is what the ordering is there for. Verified above on the same data, both cases.

Two smaller notes:

- `if batch < pruneBatchTransmissions { break }` ends the loop one iteration earlier: a short batch already proves nothing is left below the cutoff. With `ORDER BY first_seen` the extra iteration is cheap anyway, so this is a nicety, not a fix.
- The batch is 250 transmissions but an unbounded number of observations. At ~16 observations per transmission that is ~4k row deletes per batch, and the hold scales with that ratio rather than with the transmission count. Worth a sentence in the doc comment, since an instance with a denser observation ratio gets a proportionally longer hold from the same batch size.

Nothing here blocks the direction, and the test that fails against the pre-patch implementation is the right kind of proof.

efiten (approved):
Re-reviewed `ecf0b371`, and checked the claims rather than reading them.

Verified here, on a scratch checkout of the head:

- the seven prune tests and the pre-existing `TestIngestorPruneOldPackets` pass; `go vet` and `gofmt` clean
- the query plans, dumped on a 600-transmission / 2400-observation database. The subquery is `SEARCH transmissions USING COVERING INDEX idx_transmissions_first_seen`, as claimed. Both deletes are index-driven too, which the PR's own plan test does not assert: observations via `idx_observations_transmission_id`, transmissions via `INTEGER PRIMARY KEY`, each with a `LIST SUBQUERY`
- termination. `transmissions.id` is a rowid alias, so `ORDER BY first_seen, id` is a total order and both statements of a batch resolve the same set. New rows always take `first_seen = time.Now()`, and `stmtUpdateTxFirstSeen` only ever moves it to another `time.Now()`, so concurrent ingest cannot feed rows back below a fixed cutoff. No livelock path
- `foreign_keys` is ON (`cmd/ingestor/db.go:133`) and `observations.transmission_id` is the only FK into `transmissions`, so child-first ordering inside a batch is required, and correct
- both callers discard `n` when `err != nil`, so the changed error return is observably inert, as you say
- the writer-stats count assertions are not racy: nothing in `cmd/ingestor` uses `t.Parallel()` and `OpenStore` starts no instrumented background writer

Two low findings. Neither blocks, and I am merging on the strength of the above, but both are worth a follow-up.

**1. `maintenance.go:87` — the dropped `RowsAffected` error now gates the loop.** `batch, _ = res.RowsAffected()` used to feed a log line; it now decides whether another batch runs. If a driver ever returns an error there, `batch` is 0, the loop breaks on `batch < pruneBatchTransmissions`, and `PruneOldPackets` returns `(committed_so_far, nil)`. Both callers read that as success, so retention is silently under-enforced until the next daily tick. `modernc.org/sqlite` returns nil there today, which is why this is low — but #1992 proposes swapping that driver for `mattn/go-sqlite3`, so "the current driver never errors here" is a shorter-lived guarantee than it looks. One line to return it.

**2. `maintenance.go:30` — the batch size trips the alarm this fix exists to quiet.** Extrapolating your own numbers, ~0.53s per batch at production scale, against `defaultSlowWriterMs = 500.0` (`cmd/ingestor/db.go:2315`): a prune day would emit up to ~64 `[db-slow-writer] component=prune_packets` lines where it previously emitted one. Operators can raise `CORESCOPE_DB_SLOW_WRITER_MS`, so it is noise rather than breakage, but a batch size that lands under 500ms on the reference instance would avoid the conversation entirely.

Deliberately not raised: the prune is no longer atomic, so a reader can observe a partially deleted retention day. I could not construct a query that returns a wrong answer from it, only a transient one. And an error mid-loop now leaves a partially pruned day and skips that run's `RunIncrementalVacuum` — the next run resumes, which is better than the old all-or-nothing rollback.

Not checked by me: the full `-race` run for `cmd/ingestor`, and the production timing numbers.

Merging once the race-detector and build jobs finish. Thanks for the measurements on the index plan, that was the part that needed a second pair of eyes.

PR #2007, opened 2026-09-11
channels: 23 more names in the rainbow guess list

nullrouten0:
## Change

Adds 23 hashtag channel names to `channel-rainbow.json` (297 → 320).

Names were collected from operators' client-side channel lists and checked
against stored ciphertext on two independent archives before landing: every
name here decrypted real traffic. Region-specific names from that harvest were
left out; these are generic topic channels that can plausibly turn up on any
mesh. Entries are name-only; keys are derived at load time as today.

## Notes for operators

This file is baked into the image, so it takes effect on the next build. The
live path is `hashChannels` in `config.json`. Only packets ingested afterwards
decrypt; rows already filed under `enc_XX` stay there.

## Perf

`decodeGrpTxt` tries every key in the map with a full HMAC per packet, so a
longer list is a proportionally more expensive loop for undecryptable packets.
Indexing keys by the one-byte channel hash the packet carries would reject 255
of every 256 candidates after two hashes and remove the per-name cost entirely.
Happy to open that as a follow-up if there is interest.


efiten (approved):
Checked the data rather than the prose: every value in the file is `SHA-256(name)[:16]`, which is `channel.DeriveKey` (`internal/channel/channel.go:16`). 319 of 320 entries match exactly, and the single exception is `Public`, the firmware default whose key is fixed rather than derived — so all 23 new entries are verifiable from their own names, no duplicate hashes anywhere in the file, and it parses.

That also means a wrong name here costs nothing but a dead entry: the key is a function of the name, so it either decrypts that channel or it decrypts nothing.

Approving and merging. Thanks for leaving the region-specific names out; a generic topic list is the part that travels.

PR #2008, opened 2026-09-11
feat(ingestor): log a throttled warning when the IATA whitelist drops a region

nullrouten0:
## Problem

`observerIATAWhitelist` drops non-whitelisted regions silently. An allow-list
fails in the dangerous direction: a legitimate but unlisted region vanishes with
nothing to show for it. A denylist at least lets the junk that gets through be
visible.

## Change

Emit one log line per dropped region, re-logged at most every
`iataWarnIntervalSec` (new optional config key, default 21600 = 6h) while that
region keeps arriving.

```
MQTT [gomesh] [region-filter] dropping unknown region 'XYZ' (not in observerIATAWhitelist) -- further messages from XYZ suppressed for 6h
```

Per-message logging is not an option: a foreign feed runs to thousands of
messages a day. The periodic re-log is deliberate. A strict log-once emits a
single edge event that any scrape window eventually rolls past, leaving an
actively-dropping region indistinguishable from a healthy one. The line format
is fixed and grep-friendly so a log scraper can key on it.

Behaviour with no whitelist configured is unchanged. No change to what is
accepted or dropped, only to what is logged.

## Perf

One mutex-guarded map lookup per dropped message (drops only, not the accept
path). Map is bounded by the number of distinct dropped region codes.

## Tests

`cmd/ingestor/iata_drop_warn_test.go`: throttle interval, default, edge cases
(first drop logs, repeat within interval does not, repeat after interval does,
regions independent). Existing IATA whitelist gate tests unchanged and passing.
`gofmt` and `go vet` clean.

### DIFF
--- a/cmd/ingestor/config.go
+++ b/cmd/ingestor/config.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/meshcore-analyzer/dbconfig"
 	"github.com/meshcore-analyzer/geofilter"
@@ -76,6 +77,14 @@ type Config struct {
 	// obsIATAWhitelistCached is the lazily-built uppercase set for O(1) lookups.
 	obsIATAWhitelistCached map[string]bool
 	obsIATAWhitelistOnce   sync.Once
+
+	// IATAWarnIntervalSec throttles the one-line-per-region warning emitted when
+	// ObserverIATAWhitelist rejects a region. 0 => defaultIATAWarnIntervalSec.
+	IATAWarnIntervalSec int `json:"iataWarnIntervalSec,omitempty"`
+
+	// iataWarnLast tracks when each dropped region was last logged.
+	iataWarnMu   sync.Mutex
+	iataWarnLast map[string]time.Time
 
 	// ObserverBlacklist is a list of observer public keys to drop at ingest.
 	// Messages from blacklisted observers are silently discarded â€” no DB writes,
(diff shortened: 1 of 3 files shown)


efiten (changes requested):
Right problem, and the reasoning for a periodic re-log over log-once is the part I would have argued for too: a single edge event is indistinguishable from a healthy instance once the scrape window rolls past it. The gate itself is untouched, the lookup is on the drop path only, and the tests cover the throttle, the default and the per-region independence.

One thing before merge.

**`iataWarnLast` is keyed on an unvalidated, publisher-controlled string and never evicts.** `parts[1]` is the second segment of the MQTT topic (`main.go:691`, `strings.Split(topic, "/")`), and the default subscription is `meshcore/#` (`main.go:158`). So anyone who can publish to the broker chooses those keys, and every distinct one is retained for the life of the process.

Measured on this branch, feeding distinct codes through `ShouldWarnIATADrop`:

```
distinct codes fed: 200000, map entries retained: 200000, heap delta: 15.1 MB
```

"Bounded by the number of distinct dropped region codes" is true, but that number is set by whoever publishes, not by the instance. The whitelist drops those messages either way, so the only thing that accumulates is the bookkeeping for messages we already decided to throw away.

Cheapest fix that keeps the feature intact: refuse to *record* what cannot be a region code. A real IATA code is three letters; anything longer than, say, 8 characters or non-alphanumeric can still be logged once and then skipped for recording, or simply rejected at the top of `ShouldWarnIATADrop` the way the empty string already is. A size cap on the map (stop recording past N, keep logging) works too, and has the advantage of bounding the pathological case regardless of shape.

Minor, take it or leave it: the log line's `%.0fh` reads "0h" for any interval under 30 minutes, so an operator setting `iataWarnIntervalSec: 300` to debug a feed gets `further messages from XYZ suppressed for 0h`. `time.Duration.String()` on `cfg.IATAWarnInterval()` prints `5m0s` / `6h0m0s` and needs no format verb.

Happy to merge once the map cannot be grown from outside.

efiten:
Ping on this one. It has been sitting since my review on 2026-09-12 and the single blocker is still open: `iataWarnLast` is keyed on a publisher-controlled topic segment and never evicts (200,000 distinct codes retained 200,000 entries and 15.1 MB of heap in the measurement on your branch).

Either of the two shapes from that review is fine by me: reject what cannot be a region code before recording it, or cap the map and keep logging past the cap.

Master has moved on since then, so the branch will need a rebase too.

Happy to merge once the map cannot be grown from outside. If you would rather not carry it further, say so and I can finish it with you credited as the author.

efiten:
Second ping, and one piece of news worth having before you push: master was red from 2026-09-16T10:00Z until about an hour ago, for a reason that had nothing to do with any open PR. Three ingestor tests carried a fixture date that aged past the 30 day timestamp cutoff in `resolveRxTimeCore`. Fixed in #2034, merged as `5430bc79`. So if you rebased in that window and saw a red Go job, that was not your branch.

The blocker from my review of 2026-09-12 is still the only thing in the way, and either of the two shapes there works. v3.11.0 shipped meanwhile, so the rebase is a bit larger than it was.

If I hear nothing by 2026-09-24 I will finish it with you credited as the author, so the work lands either way. Say the word earlier and it is yours.

efiten:
Finishing this as offered on 2026-09-16 and 2026-09-17, one day past the date I named. It is #2067.

**Your commit is the first of the two there, unchanged and under your authorship.** I cherry-picked it onto current master rather than reimplementing, so the feature, its reasoning and its tests are yours. The second commit is only the blocker fix.

On that blocker, I took the cap rather than the shape-validation of the two shapes I offered, and the reason is worth recording: nothing in this codebase constrains an IATA code's shape. It is uppercased and trimmed in `config.go` and `db.go` and never validated. Rejecting by shape would have invented a rule operators never agreed to, and would have silently dropped the warning for anyone using a code that did not fit — the same failure mode your PR exists to fix, one level down. So the map is capped at 512 and drops past the cap are still logged, throttled on a shared timestamp.

Closing this in favour of #2067 so the rebase and the fix travel together. If you would rather carry it yourself after all, say so and I will close mine instead — the work is yours either way, and there is no hurry now that it is unblocked.

Thanks for the feature. The periodic re-log rather than a strict log-once was the right call and I would not have got there on my own.

efiten:
Landed as #2067, merged just now. Your commit is in master under your authorship — `git log --author=nullrouten` finds it.

Closing this as superseded, not rejected. Thanks again for the feature, and for the periodic re-log design in particular; the throttle shape was the part I would not have got right on my own.

Issue #812, opened 2026-04-21
QA #809 §4.7: channel filter on Packets page is non-functional (UI + API)


Kpa-clawbot:
## Problem
QA #809 §4.7 — Channel filter on Packets page is non-functional both in UI and server-side.

## Evidence
QA run on `9e90548`.

UI: `/#/packets` has no channel filter control. Deep-linking `/#/packets?channel=%23test` does not apply (count stays at full window).

API: `GET /api/packets?channel=<x>` returns identical `total=53109` for `public`, `#test`, `#fire`, and `nonexistentchannel`. Latency is fine (~135 ms), but the `channel` query parameter is being silently ignored.

## Acceptance
- Server: `GET /api/packets?channel=#test` filters to that channel; `total` and `packets[]` reflect the filter; `?channel=nonexistentchannel` returns `total=0`
- UI: a channel filter control on `/#/packets`, URL-syncable via `?channel=...`, that round-trips on reload
- Perf criterion already met (≤ 500 ms) — preserve it (target ~150 ms)

Source PR: #762/#763

PR #2009, opened 2026-09-11
fix(packets): show browser-added channels in the channel filter

nullrouten0:
## Problem

Channels an operator adds in their own browser (Channels page → Add channel) never
appear in the Packets page channel dropdown.

Those keys live only in `localStorage` and never reach the server
(`channel-decrypt.js` keeps them client-side by design). The ingestor therefore
can't decrypt that traffic and files it under the synthetic channel hash
`enc_<HH>`, and `/api/channels` excludes `enc_` rows. The picker is populated
purely from that endpoint, so a channel you just added is invisible there even
though its packets are in the DB.

## Fix (frontend only)

Derive the same `enc_<HH>` value client-side from each stored key
(`SHA-256(key)[0]`, the derivation the decoder uses) and offer those channels
as an extra **"My Channels (this browser)"** option group.

No backend change: `/api/packets?channel=enc_<HH>` already matches on that value
in both the SQLite path and the in-memory store (`packetMatchesChannel`).

- A local channel is dropped when the server already exposes it, either by the
  same `enc_` hash or by the same channel name, so nothing is listed twice.
- Options are built with `textContent`, never `innerHTML` (channel names are
  untrusted input, #812).

## Perf

One SHA-256 per stored key (a handful), once per page init, issued in parallel
with the existing `/api/channels` fetch. No new request, no work in a render
path.

## Tests

`test-packets-local-channels.js`: 14 unit tests covering merge/dedupe/sort and
the `enc_<HH>` derivation, checked against real node-crypto SHA-256 including
the zero-padded uppercase format. Wired into `test-all.sh` and the CI unit-test
step.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -229,6 +229,7 @@ jobs:
           node test-confidence-indicator.js
           node test-1659-analytics-warmup.js
           node test-channels-merge-1498-unit.js
+          node test-packets-local-channels.js
           node test-issue-1518-home-url.js
           node test-channel-decrypt-insecure-context.js
           node test-live-region-filter.js
--- a/public/packets.js
+++ b/public/packets.js
@@ -1042,6 +1042,92 @@
     else _docColMenuCloseHandler = handler;
   }
 
+  // --- Locally-added channels in the channel filter ---------------------
+  // Channels the operator adds in their own browser (Channels page → "Add
+  // channel") live only in localStorage — the keys never leave the client
+  // (channel-decrypt.js). The server therefore cannot decrypt their traffic
+  // and files it under the synthetic channel hash "enc_<HH>", which
+  // /api/channels omits. Result: a channel you just added is invisible in
+  // the packets channel picker.
+  //
+  // Fix: derive the same "enc_<HH>" value client-side from the stored key
+  // (SHA-256(key)[0], exactly what the ingestor writes) and offer those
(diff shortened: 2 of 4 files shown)


efiten (approved):
Checked the claim the whole change rests on — that `enc_<HH>` is exactly what the server stores and filters on — against production rather than the source alone.

- The ingestor writes `"enc_" + decoded.Payload.ChannelHashHex` (`cmd/ingestor/db.go:2282`), and `packetMatchesChannel` has an explicit `enc_` branch (`cmd/server/store.go:3536`).
- Real values on a live database are uppercase two-digit hex with real volume behind them: `enc_28` 2708 packets, `enc_9A` 1693, `enc_D0` 1645, `enc_11` 1511. Your `toString(16).padStart(2,'0').toUpperCase()` produces exactly that shape.
- `GET /api/channels` there returns 91 channels, none of them `enc_`, which is the invisibility you describe.
- `GET /api/packets?channel=enc_28&limit=3` returns three packets, so the option the picker adds resolves to a working filter without any backend change.

The test suite passes (14) on the branch, and it fails when I disable the dedupe-by-hash guard, so the dedupe is genuinely pinned rather than incidentally true.

Two things I like beyond the fix itself: `buildChannelOptions` is pure and tested separately from the DOM, and the options are built with `textContent`, which keeps an untrusted channel name out of `innerHTML` (#812). The name-based dedupe on top of the hash dedupe is the right extra step — the same channel can reach the picker by either route.

Merging.

PR #2010, opened 2026-09-11
fix(tables): stop the trailing action column cropping its own button

nullrouten0:
## Problem

On the node detail page the Neighbours table's "Map" button renders as a
cropped "◎ M" on every row. Not browser-related: it reproduces across Chrome,
Safari, zoom levels and incognito. The table fits its container exactly with
no overflow anywhere; the button is cropped by its own `<td>`.

Two rules combine to starve that column. Its `<th>` is empty, so it contributes
nothing to the table's width distribution, and the global

```css
.data-table td { max-width: 0; overflow: hidden; }
```

zeroes its max-content contribution as well. The auto table algorithm hands it
only the leftover space and `overflow: hidden` crops the rest, with no
scrollbar to reveal it. Measured on a 17-neighbour node:

| viewport | column | button | rows cropped |
|----------|--------|--------|--------------|
| 1024px | 28px | 53px | 17/17 |
| 1440px | 40px | 53px | 17/17 |
| 1920px | 54px | 53px | 17/17 |
| 2560px | 73px | 53px | 0/17 |

The 420px split panel is worse: 12px of column (padding only), cropping 47 of
53px at every width. Below ~640px the table finally overflows and scrolls, so
the bug vanishes again. It bites precisely in the band every desktop user is in.

## Fix

A `.col-action` class that opts the column out of the zeroing
(`max-width: none`), sizes it to its content (`width: 1%`, the
shrink-to-content idiom for auto-layout tables) and stops it cropping
(`overflow: visible`, so a future miscalculation overhangs visibly rather
than silently). The column settles at 65px; nothing is cropped at
390/768/1024/1280/1440/1920 in either host. No new horizontal overflow: where
the table already overflowed, `.panel-right` and `.node-full-body` already
scroll.

`analytics.js` had the same shape in "Top 20 Longest Hops" and "Top 10 Longest
Multi-Hop Paths" (empty trailing `<th>`, icon button in the last `<td>`) and
was cropping 20/20 rows. Both are tagged too.

## Tests

`test-neighbor-map-btn-clip-e2e.js` needs no server: the bug is entirely
`style.css` plus the markup contract, so it renders the real stylesheet
against the real emitted row markup via `setContent()`, in both the
full-screen and split-panel hosts. Mutation-checked: reverting the rule to
`overflow: hidden` fails 14 of its 16 assertions. Wired into the CI e2e step.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -628,6 +628,7 @@ jobs:
           CHROMIUM_REQUIRE=1 BASE_URL=http://localhost:13581 node test-issue-1630-reach-mobile-e2e.js 2>&1 | tee -a e2e-output.txt
           CHROMIUM_REQUIRE=1 BASE_URL=http://localhost:13581 node test-node-reach-coverage-e2e.js 2>&1 | tee -a e2e-output.txt
           CHROMIUM_REQUIRE=1 BASE_URL=http://localhost:13581 node test-issue-1640-compare-discovery-e2e.js 2>&1 | tee -a e2e-output.txt
+          CHROMIUM_REQUIRE=1 node test-neighbor-map-btn-clip-e2e.js 2>&1 | tee -a e2e-output.txt
 
       # #1616: slide-over focus-restore flake-gate. Runs the slide-over
       # E2E 20 consecutive times against the SAME backend instance so
--- a/public/analytics.js
+++ b/public/analytics.js
@@ -2917,7 +2917,7 @@
       }
 
       // Top hops leaderboard
-      html += `<div class="analytics-section"><h3><svg class="ph-icon" aria-hidden="true"><use href="/icons/phosphor-sprite.svg#ph-trophy"/></svg> Top 20 Longest Hops</h3><table class="data-table"><thead><tr><th scope="col">#</th><th scope="col">From</th><th scope="col">To</th><th scope="col">Distance (${distUnitLabel})</th><th scope="col">Type</th><th scope="col">Obs</th><th scope="col">Best SNR</th><th scope="col">Median SNR</th><th scope="col">Packet</th><th scope="col"></th></tr></thead><tbody>`;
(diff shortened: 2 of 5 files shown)


efiten (approved):
Reproduced the bug and the fix on a live instance before reading the patch, and your numbers hold up. Node detail page, Neighbours table, measuring the last cell against the button inside it:

| viewport | column | button | `max-width` | `overflow` | clipped |
|---|---:|---:|---|---|---|
| 1440px | 41px | 53px | `0px` | `hidden` | yes |
| 1024px | 29px | 53px | `0px` | `hidden` | yes |

Your table says 40px and 28px at those widths; the pixel of difference is my scrollbar. Injecting the `.col-action` rule and tagging the cells took the column to 65px at both widths with nothing cropped, so the diagnosis (empty `<th>` contributes nothing, `max-width: 0` zeroes the max-content contribution, auto layout hands over the remainder, `overflow: hidden` eats the rest) is the whole story rather than a plausible one.

The test is real too. On your branch it reports 16 passed. I removed only the `.col-action` block from `style.css` and re-ran: 3 passed, 13 failed, with the geometry checks reporting `6 !== 0` cropped rows. So it fails for the reason it claims, and it fails on the CSS rather than on the markup.

Two details I checked because they are where this kind of change usually leaks:

- The CI line omits `BASE_URL` where its neighbours in that step all set it. That is correct here: the test builds its own shell with `setContent` and loads the real `style.css`, so it needs no server. Worth knowing it is deliberate rather than a copy-paste slip.
- Other tables with an empty trailing `<th>`: `map.js`'s path inspector and `audio-lab.js`'s note table both carry one, and neither is a `.data-table`, so the `max-width: 0` rule never reaches them. No other `.data-table` puts a button in the last cell. The fix covers what is affected.

`width: 1%` with `max-width: none` is the right idiom, and `overflow: visible` is the part I would have asked for: a future miscalculation shows as an overhang instead of silently eating the control again.

Merging.

PR #2027, opened 2026-09-13
feat(home): link My Mesh node cards to the node page

sylr:
## What

The node cards in the **My Mesh** grid on the home page could open the full
health panel or the node's packets, but there was no way to reach the node
detail page from them — you had to go search for the node again.

Each card now leads with a **Node page →** button that navigates to
`#/nodes/<pubkey>`, the same route the channels, live and analytics pages
already link to.

## Details

- Wired through the existing `.mnc-btn` click delegation, so it inherits the
  `stopPropagation()` that keeps the card's own click-to-health handler from
  firing as well.
- The error-state card (health fetch failed, including the 404 *"waiting for
  first advert"* case) gets the button too, on its own actions row. *Full
  health* and *View packets* stay off that card — the health fetch is exactly
  what failed, but the node page still resolves for a node that has so far only
  been seen in channel messages.
- `.mnc-actions` now wraps, so three buttons don't overflow a narrow card.

## Testing

`node --check public/home.js` passes. Lint and the Playwright suite were not
run: the worktree this was written in has no `node_modules`. The existing home
e2e test targets `.mnc-btn[data-action="health"]` specifically, so the new
button does not disturb it.


efiten (approved):
Reviewed and checked in a browser. Approving.

**Code.** The new button goes through the existing `.mnc-btn` delegation, so it keeps the `stopPropagation()` that stops the card's own health handler. `encodeURIComponent` on the pubkey matches how `analytics.js` builds `#/nodes/` links. `data-key="${mn.pubkey}"` is unescaped, but that is the same pattern as the two existing buttons and the value is a hex pubkey from the My Mesh list; `scripts/check-xss-sinks.sh --diff upstream/master` reports nothing. The CSS change only adds `flex-wrap`, no colours.

**Browser.** Local server with the e2e fixture, headless Chromium, My Mesh seeded with one known node and one unknown pubkey, at 1280 px and 390 px wide:
- normal card: `Node page`, `Full health`, `View packets`; error card: only `Node page`
- no overflow in `.mnc-actions` at either width
- clicking `Node page` goes to `#/nodes/<pubkey>` without opening the health panel; for the unknown pubkey the node page shows "Node not found", as expected

**Not covered:** there is no test for the new action; the existing home e2e only targets the health button. Fine for a navigation-only change of this size, but a one-line assertion next to the health case would pin it if someone touches the delegation later.

CI is green on this head. Merging.

PR #2033, opened 2026-09-17
fix(rx-coverage): open at the configured map default, with its own saved viewport

### LINKED ISSUE #2032: feat: make rx-coverage menu use the center that config.json sets for Map
rx-coverage centers around Antwerp now on every reload, it should use the already set mapDefaults from config.json (and the stored "last used" zoom/coordinates from localstorage if possible)

liquidraver:
RX coverage always opened near Antwerp, ignoring the centre the deployment configures for Map.

This initializes the viewport in this order: explicit URL coordinates, this page's own saved position (`rx-coverage-view`), `/api/config/map`, then the main map's offline fallback. Pan/zoom saves the viewport and URL while retaining the time window and observer filter. Observer-only links still fit the observer's coverage; links with explicit coordinates retain their view. Initialization and delayed redraws are guarded against navigation away and re-entry.

Per review, the two pages remember their positions independently: coverage reads and writes `rx-coverage-view` and never touches `map-view`, so panning the hex map does not move where Map opens next. What is shared is the configured default, which was the actual complaint in #2032. Both suites assert `map-view` stays untouched, so re-introducing a shared write fails rather than passing quietly.

Fixes #2032.

## Validation

- Full standalone frontend suite (`sh test-all.sh`): passes, except `test-preflight-xss-gate.js`, which fails identically on a clean `origin/master` checkout here — its Python helper crashes printing a non-ASCII character to a cp1252 Windows console. Environmental, not from this branch.
- Inventory guard (`node tests/unit/test-test-inventory.js`): 292 suites classified, 178 in the unit gate.
- `node tests/unit/test-rx-coverage-viewport.js`: passes, including from a working directory other than the repo root.
- Mutation-checked: pointing the write back at `map-view` fails the suite, and writing both keys fails the new independence assertion specifically.
- Go: unchanged by this branch. The earlier red job was the expired RF-sample fixtures, fixed in #2034 and now in the base.

## Performance and configuration

Viewport selection is constant-time, with at most one config request per initialization and none when the URL or saved state supplies the view. Existing debouncing coalesces redraw and persistence. No new configuration values or customizer settings.

## Not done

The E2E suite has not been run locally — it needs a served instance plus a browser. It was registered in `scripts/non-unit-tests.json` but invoked by nothing, so this branch adds the missing `deploy.yml` line; CI is the only place it has executed.


efiten:
The red Go job here is not caused by this PR. The three failures are in `cmd/ingestor/client_rf_sample_test.go`, and this branch touches only `public/rx-coverage.js` plus test files.

Those tests seed samples with the fixed timestamp `2026-08-17T10:00:00.000Z` and then query that August window back, while `resolveRxTimeCore` (`cmd/ingestor/main.go:1527`) replaces any timestamp older than 30 days with the ingest time. That cutoff passed on 2026-09-16T10:00Z, so the seeds now land at `time.Now()` and no delta falls inside the queried window. Master is red for the same reason: its last green run, 35077917769 on `b8c8d98e`, finished 50 minutes before the cutoff.

Fix is in #2034. Once that is merged, rebase or push any commit here and this job should go green. No change needed on your side.

I will review the viewport change itself separately.

efiten (changes requested):
Reviewed the viewport path. One thing needs a change, one is a question, the rest is verification.

## Must fix: the new E2E never runs

`deploy.yml` invokes E2E suites by explicit line, 93 of them starting at `deploy.yml:487`. `scripts/non-unit-tests.json` only exempts a file from the unit runner, it does not schedule anything.

Measured on this branch: 12 of the 104 `test-*-e2e.js` files sit in exactly that position, present in `non-unit-tests.json` and in no workflow or script at all. `test-rx-coverage-mobile-nav-e2e.js` is one of them, so the gap predates this PR, but `test-rx-coverage-viewport-e2e.js` would become the 13th. One line next to the other invocations fixes it:

```
BASE_URL=http://localhost:13581 node test-rx-coverage-viewport-e2e.js 2>&1 | tee -a e2e-output.txt
```

The unit test is wired correctly through `test-all.sh`.

## Question: the page now writes the main map's saved viewport

`map-view` is the key `public/map.js` persists its own position in (written at `map.js:707`, read at `map.js:599`), and the `moveend` handler here writes the same key with the same `{lat, lng, zoom}` shape. So panning the coverage page changes where the main map opens next. The unit test asserts that write, so it looks deliberate, while the title says "use".

Two ways to land it:

- **(a)** keep the shared write, which is what makes "share the main map's saved position" symmetrical in both directions
- **(b)** read the shared key but persist the coverage page's own position under a separate key, so the two pages remember independently

I would take (a) and note it in the merge as intended, but it should be a decision rather than a side effect. Say which you meant and I will not hold the PR on it.

## Verified

- `node test-rx-coverage-viewport.js` passes on this head.
- The test bites. Mutation checks: replacing the saved-viewport read with `null` fails it (exit 1), and dropping the `current !== generation` guard in front of `L.map` fails it too (exit 1). So both the feature and the destroy/re-entry race are actually covered, not just exercised.
- Precedence matches the main map exactly: URL hash, then `localStorage`, then `/api/config/map`, then `[37.6, -122.1]` at zoom 9 (`map.js:581-602`). Skipping the config fetch when the first two answer is a fair improvement, the outcome is identical.
- Routing the saved value through `parseViewportHash` gives this page validation that `map.js:600-602` does not have: it JSON-parses and uses `lat`/`lng`/`zoom` raw. The 7 malformed-saved-value cases in the unit test are worth having.
- `parseViewportHash` is a real global (`app.js:314`, exposed at `app.js:347`), so the call resolves.
- `test-test-inventory.js` passes: 292 root suites classified.

## Nits, no action needed

- `parseViewportHash` is called unguarded here, while `live.js:1411` and `map.js:593` both guard with `typeof parseViewportHash === 'function'`.
- eslint flags `setImmediate` as `no-undef` at `test-rx-coverage-viewport.js:9`. CI lints only `public/*.js` (`deploy.yml:268`) so it will not fail the build, and Node defines it. The `selectedName` unused-var warning in `rx-coverage.js` predates this PR, it is on `upstream/master` too.

## Not done

I did not run the new E2E: it needs a served instance plus a browser, and as wired it would not run in CI either. I did not review the coverage drawing itself, only the viewport path.

The red Go job was unrelated, the RF-sample fixtures expired 30 days after they were written. Fixed in #2034, merged as `5430bc79`. Rebase and the Go job should pass, and the E2E job will run on this branch for the first time.

efiten:
Heads-up, two things landed on master that affect this branch, and both are cheap to absorb in one rebase.

**1. Tests moved out of the repo root** (#2036, merged as `89377333`). Your two new files need to go to:

- `tests/unit/test-rx-coverage-viewport.js`
- `tests/e2e/test-rx-coverage-viewport-e2e.js`

and the two list entries become `node tests/unit/test-rx-coverage-viewport.js` in `test-all.sh` and `tests/e2e/test-rx-coverage-viewport-e2e.js` in `scripts/non-unit-tests.json`. The inventory guard now fails the build if a `test-*.js` sits in the root, so this is not optional after the rebase.

**2. The must-fix from my review still stands**, and it fits in the same push: the E2E suite needs a line in `deploy.yml` next to the other invocations, otherwise it is registered but never runs.

```
BASE_URL=http://localhost:13581 node tests/e2e/test-rx-coverage-viewport-e2e.js 2>&1 | tee -a e2e-output.txt
```

Also on master since your run: the red Go job you saw was the expired RF-sample fixtures, fixed in #2034. After the rebase that job should pass.

The only open question from my review is the one about writing `map-view`, the key the main map persists its own position in. Tell me which of the two you meant and I will not hold the PR on it.

liquidraver:
Option (b), and slightly stricter than you framed it: the two pages do not read or write each other's key at all.

Your (b) kept the shared read, seeding coverage from the main map's saved position on a first visit. I went fully independent instead — coverage reads and writes `rx-coverage-view` and never touches `map-view`. The difference only shows on a first visit with no saved coverage position: your version would open wherever the main Map was left, this one opens at the deployment's configured centre, which is what #2032 asked for. Sharing the configured *default* was the bug; sharing the *session position* was never part of it.

Both suites now assert `map-view` is untouched after a pan, so re-introducing the shared write fails instead of passing quietly. Mutation-checked: making the handler write both keys exits 1 on that assertion.

One thing worth flagging, since it was invisible while the suite wasn't running: the E2E itself asserted `localStorage.getItem('map-view')`. Wiring it into `deploy.yml` without touching that line would have turned the job red on its first ever run.

Rebased onto 89377333, which absorbs both of your heads-ups:

- Suites moved to `tests/unit/test-rx-coverage-viewport.js` and `tests/e2e/test-rx-coverage-viewport-e2e.js`, with the `REPO_ROOT` pattern you applied to the siblings in #2036. Verified the unit suite passes from an arbitrary cwd, not just the repo root.
- `test-all.sh` runs `node tests/unit/test-rx-coverage-viewport.js`; `scripts/non-unit-tests.json` lists `tests/e2e/test-rx-coverage-viewport-e2e.js`. The inventory guard passes: 292 suites classified, 178 in the unit gate.
- The must-fix is in, next to the other invocations:

      BASE_URL=http://localhost:13581 node tests/e2e/test-rx-coverage-viewport-e2e.js 2>&1 | tee -a e2e-output.txt

  Confirmed by parsing the workflow that it lands inside `Run Playwright E2E tests (fail-fast)`.
- Also placed the `test-all.sh` entry in alphabetical order; the previous head had it between `config-race` and `escape`.

Both jobs are green on this head: Go passes now that #2034 is in the base, and Playwright passes with the new suite included — so the E2E has now had its first real execution, and the `map-view`-untouched assertion holds against a live Leaflet map rather than only the harness.

Left your two nits alone as you said: the unguarded `parseViewportHash` call and the `setImmediate` no-undef, which is outside the `public/*.js` lint scope in any case.

Not done: I have not run the E2E locally — it needs a served instance plus a browser, so CI is the only place it has executed.

efiten (approved):
Second pass. Everything I asked for is in, the option question is answered better than I framed it, and I verified the claims rather than taking them.

## Verified

**The two keys are fully separate.** `public/rx-coverage.js` touches `rx-coverage-view` only, at lines 253 and 279, and `map-view` appears nowhere in the file. Both suites pin it: `tests/unit/test-rx-coverage-viewport.js:62` and `tests/e2e/test-rx-coverage-viewport-e2e.js:54`.

**The tests bite, on all three behaviours.** I mutated the source and re-ran the unit suite:

| Mutation | Result |
|---|---|
| also write `map-view` on `moveend` (the shared write you rejected) | exit 1 |
| ignore the saved `rx-coverage-view` (the persistence half) | exit 1 |
| drop the `/api/config/map` fetch (the original #2032 bug) | exit 1 |

So the suite covers the feature, the fix and the thing we discussed, not just the happy path.

**The wiring is right.** `test-all.sh` runs the unit suite, `scripts/non-unit-tests.json` lists the E2E one, and the `deploy.yml` line sits inside `Run Playwright E2E tests (fail-fast)`, which I checked by parsing the workflow rather than by eyeballing the diff. Inventory guard: `292 suites classified; 178 run by local and CI unit checks`.

**It has now actually run.** CI run 35261689694: the Go job prints `RX coverage viewport regressions OK` and the Playwright job prints `RX coverage viewport browser regressions OK`, the latter being the suite's first real execution anywhere. All jobs green.

## On the option question

Going fully independent is the better call and your reasoning is the part I want to keep on the record: sharing the configured *default* was the bug, sharing the *session position* never was. My (b) would have left a first visit opening wherever someone last dragged the main Map, which is not what #2032 asked for.

The catch you flagged is the useful half: the E2E asserted `localStorage.getItem('map-view')`, so wiring it into `deploy.yml` without touching that line would have turned the job red on its first ever run. That is exactly the failure mode #2037 is about, and this PR is the first case where the gap was closed before it bit.

## Nits, still fine to leave

The unguarded `parseViewportHash` call and `setImmediate` as `no-undef` in the unit suite. Neither is reachable in CI (`eslint` only covers `public/*.js`, `deploy.yml:268`).

Approving. I did not run the E2E myself either, for the same reason you give: it needs a served instance and a browser.

PR #2035, opened 2026-09-17
fix(store): account path, decode-cache and dedup-key bytes so maxMemoryMB eviction triggers

A13xB0:
## Summary

`packetStore.maxMemoryMB` never triggers on a real instance because `trackedBytes` undercounts the packet store by about 2.5×. On our instance (2 cores, 43.9k transmissions, 258k observations, 168 h retention):

- `trackedMB` was 104–151 MB.
- The store-owned heap was about 370 MB (heap profile after a forced GC).
- The Go heap sat against `GOMEMLIMIT=650MiB`, so the GC ran every ~1.4 s: 214,824 cycles in 84 h. `corescope-server` used 160–190% CPU the whole time.

## Causes

1. **Path costs are never charged.** A `StoreTx` is charged in `Load`, `loadChunk`, `LoadChunked` and `IngestNewFromDB` when it is created. That is before `pickBestObservation` sets `PathJSON`, so `estimateStoreTxBytes` sees zero hops, and the `byPathHop` and `spTxIndex` costs (O(path²)) are never added.
2. **Eviction over-subtracts.** Eviction recomputes the estimate with the path known, so it subtracts more than was added. `trackedBytes` then drifts downwards: 129 → 104 MB over two days with a steady packet count.
3. **Some allocations aren't estimated at all:**
   - the `ParsedDecoded` cache: about 4× the JSON length (~40 MB here), and analytics fill it for every transmission;
   - the `obsKeys` dedup key, which is a separate concatenated string per observation (~40 MB here);
   - each observation's `ObserverName`, `ObserverIATA`, `Direction`, `RawHex` and `Timestamp`.

## Fix

- `StoreTx.accountedBytes` records what the transaction was charged.
- `rechargeTx(tx)` returns the difference against a fresh estimate. It is called at creation, as before, and again after every `pickBestObservation`.
- Eviction, and the candidate selection in `evictionCandidateTxIDs`, subtract `tx.accountedBytes` instead of recomputing, so the total stays consistent however often the path changed.
- `estimateStoreTxBytes` and `estimateStoreObsBytes` now count the missing allocations above (`decodedCacheFactor`, `obsKeyEntryBytes`).

## Proof

The same production DB copy was loaded with v3.11.0 and with this branch, and a heap profile was taken with `?gc=1` after the load finished:

| | trackedMB | heap in use after GC |
|---|---|---|
| v3.11.0 | 151 | 402 MB |
| this branch | 351 | 396 MB |

The remaining gap is the neighbor graph, analytics caches and SQLite row buffers, which are not part of the store.

## In production

This branch has been running on our instance for 12 h (`GOMEMLIMIT=1GiB`, `maxMemoryMB=400`, same data), with no restarts:

| | v3.11.0 | this branch, 12 h |
|---|---|---|
| GC cycles | ~0.71/s | 485 in 43,242 s (~0.011/s) |
| total GC pause | – | 469 ms |
| `corescope-server` CPU | 160–190% sustained | ~0.3–3% at rest, with short spikes during analytics and page loads |
| `trackedMB` | 104–151, drifting down | ~350, flat |
| heap allocated | against the limit | 513 MB |

Eviction keeps `trackedMB` steady: 683 eviction passes, 3,251 packets evicted in total.

## Perf

- `rechargeTx` is O(path) per transaction: it re-parses the path already cached by `txGetParsedPath`. It runs once per `pickBestObservation`, which is already O(observations).
- Eviction does less work than before, because it no longer re-estimates each transaction.

## Tests

- New test: `TestTrackedBytes_PathKnownAfterCreate` charges a transaction before its path exists, then recharges it and evicts it, and requires `trackedBytes` to end at 0.
- Updated for the new accounting: `makeTestStore` and `TestTrackedBytes_MatchesExpectedAfterMixedInsertEvict`.
- Updated for the new estimate terms: `TestEstimateStoreTxBytes` and `TestEstimateStoreObsBytes`.
- `go test -race ./cmd/server` passes.

## Operator note

Only instances with `packetStore.maxMemoryMB` set are affected. It is unset by default (`config.go:310`), and where it is unset this change is inert: nothing is evicted and no load is clamped.

Where it is set, two things change, both because the estimate is now larger rather than because any limit moved:

- **Eviction:** `maxMemoryMB` now caps the real store size, so an instance whose true store is larger than its configured value starts evicting after the upgrade.
- **Cold load:** the startup walk is clamped by `estimateStoreTxBytesTypical(10)` (`chunked_load.go:358-366`), which goes from 5478 to 8078 bytes, so about a third fewer packets are loaded at boot: 124420 to 84374 at `maxMemoryMB=650`, 76566 to 51922 at 400. Visible history is correspondingly shorter until the store refills.

`maxMemoryMB` bounds the store **and its warm caches**: `decodedCacheFactor` is charged for every transaction, whether or not `ParsedDecoded()` has been called on it (charging inside `ParsedDecoded()` would mean writing `trackedBytes` from read paths that do not hold `s.mu`). On this instance that is roughly 40 MB of 350 MB.

`docs/go-migration.md` claimed the Go server ignored this setting entirely, which was never true; it is corrected here, along with the two summary tables and the deployment guide.


efiten (approved):
Audited the accounting rather than the arithmetic, since a counter that drifts is worse than one that is wrong by a constant. It holds up. Two things to add before merge, both operator-facing, neither a code change.

## Verified

- **Every creation site recharges, and every subtraction uses what was charged.** All six production `pickBestObservation` calls are now followed by a recharge (`chunked_load.go:470`, `main.go:341`, `store.go:1011`, `store.go:1331`, `store.go:2909`, `store.go:3301`), no site still charges the raw estimate, and the three subtraction sites (`store.go:4701`, `4770`, `4815`) all read `tx.accountedBytes`. The single `trackedBytes -=` is `store.go:4974`.
- **The recharge sites all hold `s.mu`.** `Load` holds it from `store.go:857`; `IngestNewFromDB` and `IngestNewObservations` take it at `2778` and `3142` with a deferred unlock, so the whole body is covered; `loadChunk`'s locals are not published yet. Worth stating explicitly because CI's race job runs the ingestor only, so `go test -race ./cmd/server` is on your word and this lock audit is the independent check.
- **Duplicate observations are not charged.** `store.go:1001` and `store.go:3234` sit after `totalObs++` inside the accepted branch, so an observation rejected by the `obsKeys` gate never reaches the estimate.
- **No counter leak in the chunk loader.** I went looking for a tx that gets charged and then discarded during the merge, which would strand its bytes in the total forever. There is none: `store.go:1449` publishes every local tx into `s.packets`, so each one's `accountedBytes` comes back out at eviction.

## Please add: the cold-load budget shrinks by a third

`chunked_load.go:358-366` derives the startup cursor clamp from `estimateStoreTxBytesTypical(10)`, and that value moves with this change:

```
estimateStoreTxBytesTypical(10): 5478 -> 8078 bytes (1.47x)

maxMemoryMB=400    maxPackets 76566 -> 51922   (32% fewer)
maxMemoryMB=650    maxPackets 124420 -> 84374  (32% fewer)
maxMemoryMB=1000   maxPackets 191415 -> 129806 (32% fewer)
```

So an instance with the cap set loads about a third fewer packets at boot, on top of the eviction your Operator note already describes. That is the more honest number, no argument, but it is a second reason the visible history gets shorter after the upgrade and the body does not mention it. One line in the Operator note is enough.

## Please add: say that the estimate assumes a warm decode cache

`decodedCacheFactor` is charged for every transaction whether or not `ParsedDecoded()` was ever called, and `store.go:125` fills that cache lazily while nothing ever clears it. On your own figures it is roughly 40 MB of 350 MB, so on a store whose transactions analytics never touch it is a systematic overcount that makes eviction fire earlier than the real footprint.

I looked at the alternative before raising it: charging inside `ParsedDecoded()` would mean writing `trackedBytes` from read paths that do not hold `s.mu`, which buys a correct number for an atomic and a wider contract. Charging up front is the better trade. So this is only a documentation ask: state in the config reference that `maxMemoryMB` now bounds the store plus its warm caches, not just the rows.

## One data point from another instance

Your diagnosis is consistent with what a much looser instance shows, which is useful for framing the severity. On ours, 336 h retention over a 9.7 GB database with `runtime.maxMemoryMB` at 8000 and `packetStore.maxMemoryMB` unset, the pathology does not appear at all: CPU 2.3 to 4.2 percent, RSS 7.1 GiB of 23.4 available. Nothing to evict, and the GC has headroom.

That puts the boundary where the fix matters: a `GOMEMLIMIT` close to the true store size, which is exactly your 650 MiB against about 370 MB. Where `maxMemoryMB` is unset, which is the default (`config.go:310`), the accounting change is inert. Saying so in the body would help operators decide whether to care.

## Not done

I did not reproduce your heap profile, and I did not run the Go tests: since #1992 the server package needs cgo for the sqlite driver and this machine has no C toolchain, so CI is the check. I did not benchmark `rechargeTx`; your O(path) argument matches the code, which reuses the parsed path from `txGetParsedPath`.

CI was sitting at `action_required` and had never run. I approved the runs, they are going now.

efiten:
Checklist of what is still open here, so it is unambiguous. Both are documentation, the code is approved as it stands.

- [ ] **1. Operator note: the cold-load budget shrinks too, not just eviction.** `chunked_load.go:358-366` derives the startup cursor clamp from `estimateStoreTxBytesTypical(10)`, which goes from 5478 to 8078 bytes (1.47x). So on any instance with the cap set, the boot walk loads about a third fewer packets: at `maxMemoryMB=650` that is 124420 down to 84374. One line is enough.

- [ ] **2. Say that `maxMemoryMB` now bounds the store plus its warm caches**, since `decodedCacheFactor` is charged whether or not `ParsedDecoded()` was ever called.

While you are in the docs for point 2, one thing that makes this upgrade sharper than the PR body suggests. `docs/go-migration.md:81-83` currently says:

> ### `packetStore.maxMemoryMB` (Go ignores this — it's Node-only)
> The Go server has its own in-memory store that loads all packets from SQLite on startup — it does not read this config value. This is safe to leave in your config; Go simply ignores it.

That is not true any more, and it was not true before this PR either: `store.go:702` assigns it to the store, `chunked_load.go:358` clamps the cold load with it, the eviction watermarks use it, and `main.go:124-152` derives `GOMEMLIMIT` from it when `runtime.maxMemoryMB` is unset. An operator who read that page believes the setting is inert, leaves an old Node-era value in their config, and gets eviction after upgrading to your branch. Worth correcting in the same pass, or in a separate PR if you would rather keep this one narrow.

`maintainerCanModify` is on for this branch, so I can push both of these myself if you prefer. Say which and I will either wait or do it.

A13xB0:
Thanks — both added, and I took the `docs/go-migration.md` correction too rather than leaving it for a separate PR, since an operator hitting this upgrade is exactly who reads that page. Pushed as fe4d57a7; no code change.

**Operator note** now says:

- the change is inert where `packetStore.maxMemoryMB` is unset, which is the default;
- eviction, as before;
- the cold load shrinks with it. I recomputed `estimateStoreTxBytesTypical(10)` from the constants on both sides and get your figures: 5478 → 8078 bytes, 124420 → 84374 packets at `maxMemoryMB=650`, 76566 → 51922 at 400;
- `maxMemoryMB` bounds the store plus its warm caches, with the reason charging inside `ParsedDecoded()` was rejected (writing `trackedBytes` from read paths that do not hold `s.mu`).

**Docs** (`fe4d57a7`):

- `docs/go-migration.md`: the section now says both engines read the setting, and lists what it does in Go (stops the cold load, evicts oldest-first above it, sets `GOMEMLIMIT` when that and `runtime.maxMemoryMB` are both unset), with the warning that a leftover Node-era value is not inert. The two summary table rows are updated too.
- `docs/deployment.md`: one paragraph by the `maxMemoryMB` example, saying it covers the store's caches as well as the rows, and pointing at `/api/perf` (`trackedMB` against `maxMB`).

Thanks for the second data point as well. It matches the boundary in the description: with `packetStore.maxMemoryMB` unset and `GOMEMLIMIT` far above the store, there is nothing to evict and nothing to thrash against.

On the lock audit: it is the check I would have wanted, since `go test -race ./cmd/server` passing locally is only my word. Your reading matches mine at all six recharge sites.

Issue #1975, opened 2026-09-05
feat(scopes): network-wide Scope Audit page — which repeaters declare a region but never forward it


efiten:
Proposing a page that has been running on a production instance for a while and that I would like to contribute, rather than keep as a local fork feature. Filing the issue first because there is one real design decision in it, and I would rather agree that here than in a diff.

### What it does

One row per repeater that has answered a region query, ranked so the interesting rows sit at the top. For each repeater it answers a single question: **you declare these regions, but are you actually forwarding them?**

That question is currently unanswerable. `default_scope` tells you what a node's adverts were observed under, and `transported_scopes` (#1751) tells you what it carried, but nothing lines a repeater's *declared* region list up against what it was *seen* forwarding in a window. A repeater configured for eight regions that only ever forwards one looks healthy from every existing view.

Measured on a 1179-repeater instance, 197 repeaters with a declared-region answer, 24h window:

| | rows |
|---|---|
| every declared region observed forwarding | 7 |
| **none observed** | **78** |
| mixed, some observed and some not | 44 |
| no declared regions on record | 68 |

78 repeaters declaring regions they were not seen forwarding at all in 24h is the signal. Some of that is a quiet region rather than a fault, which the page says in its own header rather than leaving the reader to infer, and the window is selectable (1h / 24h / 7d) precisely so an operator can tell a quiet hour from a dead config.

### The design decision, and why this needs #1971 first

On my fork the declared-region list comes from a custom MQTT topic that a mobile companion app publishes to, `meshcore/client/{PUBLIC_KEY}/regions`. **That should not come upstream.** It works only for operators who also run that app, so it would ship a page that renders an empty table for everyone else.

#1971 solves this properly. It brings `configured_scope` on `nodes`, read back off the node itself via the observer `/neighbors` report, from the ESP32-native observer firmware that is already public and already validated on two independent meshes (@dborup on stg.meshview.dk, @SaarMesh-Bot on a Heltec V4 on SaarLorLux). That is exactly the same claim the audit needs, from a source anyone can deploy.

So: **this page should be fed by `configured_scope`, and this issue depends on #1971 landing.**

### How portable is it, measured rather than estimated

I checked what actually depends on the fork-only data source before proposing this:

- `node_declared_regions` is referenced **3 times** in the whole server file, all inside the declared-region lookup.
[...]

PR #2044, opened 2026-09-18
feat(ingestor): store CoreDrive RX region answers so the Scope Audit sees them

liquidraver:
## Problem

The Scope Audit page (#1975) says a repeater's declared region list can come from the [CoreDrive RX](https://github.com/efiten/coredrive-rx) app, but nothing the app sends ever reaches it.

- CoreDrive RX publishes region answers on `meshcore/client/{PUBLIC_KEY}/regions`.
- The ingestor's client-topic switch only handles `packets` and `rf`, so `/regions` is dropped without a log line.
- `node_declared_regions` is the table the server's `AllCurrentDeclaredRegions` and the ingestor's `declaredRegionSources` both read, but nothing in this repo creates or fills it. A stock install only ever gets the ESP32 `/neighbors` source.

I confirmed this on a live v3.11.0 instance. `mosquitto_sub` on `meshcore/client/+/regions` showed answers arriving at the broker while the table did not exist.

## Change

**Ingestor**
- New `case "regions"` in the client-topic dispatch, behind the existing `clientRxCoverage.enabled` switch.
  - Validates a full 64-hex `target` and requires a `regions` array.
  - An empty array is stored, because "floods nothing" is a real answer. A missing or non-array field is dropped rather than turned into one.
  - Names containing a comma, names over 64 characters, and anything past the 64th entry are dropped, and the answer is flagged `truncated`.
- `node_declared_regions` is added to the base schema, in the exact shape the existing readers and test fixtures use, plus an `(target, observed_at)` index.
  - `observed_at` is the answer's own timestamp, read with `resolveRxTimeCore` and stored as canonical RFC3339 so it compares directly with `configured_scope_at`.
  - Only the newest answer per target is kept, so a late buffered upload can't displace a fresher one and the table stays bounded.
- The companion's `origin` name is upserted into `client_observers`, the same as `/packets` does.

**Server**
- `declaredRegionsTablePresent()` re-probes `sqlite_master` until the table is seen, then latches.
  - Supervisord starts both processes together, so on the first run of this build the server's startup probe can lose the race and ignore every answer until the next restart.
  - Both callers sit behind 30-second caches, and after the latch there is no further cost.

**Docs**
- `docs/client-rx-coverage.md` documents the `/regions` topic, its payload, and the ACL and topic-list requirements.

## Testing

- New ingestor tests: storing an answer, empty vs missing `regions`, bad target or type, newest-wins with a late older answer, unstorable names and truncation, gate off with no observer-path fall-through, and the answer feeding `declaredRegionSources`.
- New server test: a table created after startup is picked up without a re-probe.
- `gofmt` is clean and `go vet` passes. The full server suite passes, and the full ingestor suite passes under `-race`, both on origin/master plus this branch.
- Hot-swapped into a live v3.11.0 container. A real CoreDrive RX answer was logged as `regions: ba5e00011bc8 declares [*,hu]`, stored in `node_declared_regions`, and served by `/api/scope-audit` as `declaredRegions: ["hu"], declaredWildcard: true`.


efiten:
The diagnosis is right, and I verified it separately before saying so: `upstream/master` reads `node_declared_regions` in `cmd/ingestor/region_keys.go` and `config.go`, and nothing in the tree ever creates it. The client-topic switch handles `packets` and `rf` only, so `/regions` is dropped silently. A stock install can only ever get the ESP32 `/neighbors` source, exactly as you describe.

Before this goes further you should know that an implementation of the same feature already exists and has been running in production: the ON8AR fork has carried it since 2026-09-06, and it is the deployment CoreDrive RX actually publishes to. It is not a competing design, it is the same one further along, and the difference matters mostly where it has since been corrected.

## What the two have in common

Same table name, same core columns, same `UNIQUE(target, rx_pubkey, observed_at)`, same `idx_ndr_target`, same newest-by-`observed_at` rule, and the same file names (`cmd/ingestor/client_regions.go` and its test).

## Where the fork's version goes further

| | fork | this PR |
|---|---|---|
| `lat`, `lon`, `pos_acc_m` | stored | not stored |
| `repeater_clock` | stored | not stored |
| retention | `pruneOldClientDeclaredRegionsAt` + `idx_ndr_prune`, `clientRegionsDays` | none |
| config gate | its own `clientRegions` block | rides on `clientRxCoverage` |
| older answers | kept per collector | deleted |

The position columns are what let the Scope Audit say *where* an answer was collected, which is the difference between "this repeater declares `#hu`" and "someone 40 km away, driving, got that answer once". `repeater_clock` is what exposes a repeater whose clock is wrong, which otherwise shows up as an answer that appears newer than it is. Both were added after the first version, because the first version did not have them either.

## One thing worth a second look regardless of whose code lands

```sql
DELETE FROM node_declared_regions
WHERE target = ?
  AND observed_at < (SELECT MAX(observed_at) FROM node_declared_regions WHERE target = ?)
```

That is not scoped by `rx_pubkey`, and this table is shared with the ESP32 `/neighbors` source. So a CoreDrive RX answer deletes the ESP32's answers for the same repeater, and what is lost is exactly the corroboration the Scope Audit is for: two independent collectors agreeing, or disagreeing, about one repeater's configured regions. The readers only need the newest row today, so nothing breaks; the signal just quietly stops existing. The fork keeps one row per collector and prunes by age instead.

## Where the fork's version is behind yours

Your `declaredRegionsTablePresent()` re-probe is a real bug fix the fork does not have: supervisord starts both processes together, so the server's startup probe can lose the race and ignore every answer until the next restart. The fork has been getting away with it because its table predates its server builds. Your `docs/client-rx-coverage.md` section is also better than the fork's documentation of the same topic.

## Proposal

The owner has asked me to put the fork's version upstream rather than have two implementations of one feature diverge further, so I will open that PR with **you credited as co-author**: the diagnosis, the re-probe fix and the docs are yours, and the PR will say so. If you would rather carry it yourself instead, say so and I will close mine and review yours against this list instead. Either way is fine by me; what I want to avoid is upstream and the fork growing two different schemas for one table name, where `CREATE TABLE IF NOT EXISTS` makes the difference invisible until a column is missing.

efiten:
Numbers for the two claims above, from the instance CoreDrive RX publishes to, since it is easy to argue this in the abstract and I would rather not.

`node_declared_regions` there, today:

```
rows   targets  collectors  oldest                    newest
1840   275      51          2026-08-18T18:34:57.190Z  2026-09-19T05:28:25.894Z
```

**On the extra columns:** `repeater_clock` is filled on all 1840 rows, position on 1263 of them (69%, the rest are answers collected without a GPS fix). So they are not aspirational fields, they carry data on every answer that arrives.

**On the cross-collector delete:** 138 of the 275 repeaters have answers from more than one collector, and one has 13. The delete keeps one row per target, so on this dataset it would remove 1565 of 1840 rows, 85%.

**On what that costs:** comparing each collector's newest answer per repeater, **40 repeaters have collectors that disagree** about the region list. That disagreement is the thing the Scope Audit exists to surface, and it is only visible while more than one collector's answer survives. After the delete, those 40 become 40 repeaters with one confident-looking answer.

None of this makes your PR wrong about the gap it identifies, which was real and which nothing else had caught. It is the reason I would rather land the version that has been carrying this data for a month than start that clock again.

efiten:
Landed as #2047, merged as `e0156573`, with you credited as co-author on the commit.

What went in from here: `declaredRegionsTablePresent()` and its test, which is a bug the fork version did not have and would have hit on any fresh install, and the `docs/client-rx-coverage.md` section on the `/regions` topic.

What did not: the storage side, replaced by the fork implementation that has been running since 2026-09-06 with position, `repeater_clock` and age-based retention. The reasoning and the numbers are in the two comments above and in #2047.

Thanks for finding the gap. Nothing else had caught that the page advertised a source the tree could not fill, and that diagnosis is what made the rest of this possible.

Issue #2030, opened 2026-09-16
map: leaving the map page right after load logs "Cannot read properties of null (reading 'invalidateSize')"


efiten:
Navigating away from the map within about 200 ms of it finishing its first render logs an uncaught error and the map view stops initialising.

## Reproduction

Headless Chromium, both against a staging instance on `a2ea18f7` and against a production instance on v3.10.1, so this is not new:

1. Open `#/map` and wait for `.leaflet-container`.
2. After 200 ms, set `location.hash = '#/live'`.
3. Console shows:

```
Map load error: TypeError: Cannot read properties of null (reading 'invalidateSize')
    at _renderMarkersInner (map.js:...)
```

With 600 ms or 1500 ms between the two steps there is no error. Ten runs each at 200 / 600 / 1500 ms: 1 error, 0, 0 on both instances.

## Why

`_repositionMarkers` calls `map.invalidateSize({ animate: false })` unguarded (`public/map.js:2024` on `a2ea18f7`). Its two timer-based callers guard the same variable (`public/map.js:2324`, `:2341`, both `if (map)`), which is what the rest of the file does after the view is destroyed and `map` is set to null. `_renderMarkersInner` runs from a deferred path, so it can land after the map page has been torn down.

## Impact

Cosmetic in the common case: the error is caught by the map page's own handler, which prints "Map load error", and the user is already on another page. It is worth fixing because it hides real map load failures in the same message, and because a reader cannot tell this apart from a genuine error.

## Suggested fix

Guard the call the way its callers do, or return early from `_repositionMarkers` when `map` is null (it already returns early when `_currentMarkerData` is empty). A regression test can drive the pure function with `map` unset.

## Not verified

- Only Chromium.
- Whether any user-visible state is left behind after the aborted render; the next visit to `#/map` initialises normally in the same session.

efiten:
Fixed by #2048, merged as `7f4b357d`.

The 100ms `invalidateSize` timer is now tied to the Leaflet instance that scheduled it (`public/map.js`), and `destroy()` clears the reference (`map.remove()` then `map = null`, lines 2428-2433), so leaving the page before the timer fires makes the callback return instead of touching a removed map. The same guard covers the geo-filter and node-loading continuations, which could previously let an old response render into a replacement map.

Covered by `tests/e2e/test-map-nodes-pagination-e2e.js`, which is invoked by `deploy.yml`.

[...]

Issue #2037, opened 2026-09-17
18 suites in tests/e2e are invoked by no runner, and one cannot run at all


efiten:
`tests/e2e/` now holds 113 suites, and 18 of them are invoked by nothing: no line in `.github/workflows/deploy.yml`, no script, no workflow, no npm script.

```
test-1110-live-filter.js                     test-node-reach-e2e.js
test-analytics-fluid-charts.js               test-nodes-export-e2e.js
test-channel-modal-e2e.js                    test-packets-scope-column.js
test-e2e-1267-mobile-vcr.js                  test-path-inspector-e2e.js
test-issue-1274-legend-coverage-e2e.js       test-pr-1490-live-map-gpu-animations-e2e.js
test-issue-1522-trace-url-sync-e2e.js        test-rx-coverage-mobile-nav-e2e.js
test-issue-1648-m5-icons-e2e.js              test-show-neighbors.js
test-live-dedup.js                           test-table-sort.js
test-marker-outline-weight.js                test-touch-targets.js
```

This predates #2036: E2E suites are invoked by 96 explicit lines in `deploy.yml`, and `scripts/non-unit-tests.json` only exempts a file from the unit runner rather than scheduling it. So a suite can be written, classified, and never run. #2033 was about to add the 19th before review caught it.

## Why it matters now

Before the move these were loose files in a root of 348 entries. Now they sit in a folder named `tests/e2e`, which reads as "the E2E suite". For 18 of the 113 that is not true, and nothing in the repo says which is which.

## One of them cannot run at all

`test-table-sort.js` requires `jsdom`, which is in neither `dependencies` nor `devDependencies`. It fails with `Cannot find module 'jsdom'` on master and on #2036 alike, so it has been inert for a while rather than merely unscheduled.

## Options

1. **Wire them in.** Add a `deploy.yml` line per suite, fix whatever has rotted while unwatched, and the folder means what it says. Expect some of the 18 to fail on first contact: three of them (`test-analytics-fluid-charts.js`, `test-table-sort.js`, `test-touch-targets.js`) had path code rewritten in #2036 with only `node --check` covering it.
2. **Make the guard enforce it.** `tests/unit/test-test-inventory.js` already fails when a suite is unclassified; extend it so a file in `tests/e2e` must also appear in `deploy.yml`. That turns this class of drift into a build failure instead of a discovery two years later, and it forces option 1 or option 3 for each file.
[...]

PR #2048, opened 2026-09-19
fix(map): keep inspector controls clickable and ignore stale loads

### LINKED ISSUE #1998: bug(map): Map Controls intercepts the Path Inspector toggle on desktop
Found while activating the previously unselected Path Inspector browser suite for #1995, against `b5c7612`.

## Reproduction

1. Open the Map page at a 1400×900 desktop viewport with the Path Inspector collapsed.
2. Click its `#mapPaneToggle` normally.

The click is intercepted by the Map Controls button. Measured bounds:

- `#mapPaneToggle`: x=1369, y=52, width=31, height=35.
- `#mapControlsToggle`: x=1354, y=62, width=36, height=36; z-index=1001.

Hit-testing the inspector toggle's center returns `#mapControlsToggle`. The controls panel also covers the inspector gutter. Both controls are positioned relative to the entire map wrapper, while the inspector occupies its right edge.

## Expected / acceptance

- Normal clicks can expand and collapse Path Inspector without being intercepted.
- Map Controls and its toggle remain inside the Leaflet portion when the inspector is collapsed or expanded.
- Keep the existing mobile behavior, where the inspector is hidden at widths ≤640px.
- Add pointer hit-test and actual-click regressions; do not use forced clicks or bypass event delivery.

Proposed fix: account for the existing 32px/320px inspector widths in the desktop controls offsets. This can stay within CSS, preserving DOM and event ownership.

n30nex:
Fixes #1998 and #2030.

Map Controls overlaps the Path Inspector toggle on desktop, intercepting normal clicks. Offset the controls by the existing collapsed/expanded inspector widths, preserving the mobile layout.

Leaving Map while its node data is loading can log `invalidateSize` errors or let an old response overwrite a replacement map. Tie node-loading continuations and associated initialization callbacks to their originating Leaflet instance, and ignore obsolete results before changing state or rendering.

- Red → green: `e675100` → `7d736d1` reproduces three desktop pointer failures; `fd56ea2` → `823668b` reproduces six stale-response failures.
- Validation: all 178 standalone suites, 10 controls browser checks and 9 pagination/lifecycle checks passed in independent local runs. Syntax, whitespace, CSS-variable and XSS checks passed. Three independent reviews found no required changes.
- Main browser suite: 131 passed, 3 fixture-dependent skips, no failures.
- Browser verified: local Chromium against the current Go server and fixture database; screenshots `coverage/map-desktop.png` and `coverage/map-mobile.png`.
- E2E assertion added: `tests/e2e/test-issue-1329-map-controls-accordion-e2e.js:182` and `tests/e2e/test-map-nodes-pagination-e2e.js:215`; both suites already run in CI.
- Constant-time identity checks at existing async boundaries; no additional requests or collections. No dependencies, backend changes or new settings.

## Preflight overrides

- External `run-all.sh` unavailable; applicable repository checks were run directly.


efiten (approved):
Checked both halves against the code rather than the description. It holds, and the test half is the part I would point other PRs at.

## The stale-load guard is sound

The identity check works because teardown actually clears the reference: `destroy()` calls `map.remove()` and then `map = null` (`public/map.js:2428-2433`). So after leaving the page every `map === initializedMap` and `map === loadingMap` is false, and after a replacement the new instance is a different object. That is what closes #2030: the 100ms `invalidateSize` timer now returns instead of touching a removed map.

Each `await` in `loadNodes` is followed by a check, and so are the two deferred callbacks, the error branch and the `finally`. The reordering of `nodes = data.nodes || []` to after the observers await reads like a detail but is the right one: shared state is only mutated once the load is known to still own the map, rather than half-updated by a load that has been superseded.

## The tests are the strong part

`elementFromPoint` on the toggle's centre, across five viewport widths and both pane states, is the correct way to test an overlap: a `click()` would have passed against an overlapping control by forcing the event. Doing it at 641 and 640 pins both sides of the breakpoint rather than the middle of each range.

I also checked the thing that has been biting this repo lately: both suites you extended are actually invoked by `deploy.yml`, one line each. A test that does not run is the subject of #2037, and yours run.

The commit chain matches the claim: `e675100` test then `7d736d1` fix, `fd56ea2` test then `823668b` fix.

## One nit, not a blocker

The offsets hardcode the pane widths:

```css
#mapSidePane ~ .map-controls { right: calc(32px + 12px); }
#mapSidePane.expanded ~ .map-controls { right: calc(320px + 12px); }
```

Those two numbers are the flex-basis values at `style.css:4532-4533`, now repeated four times about eight lines away. Change the pane width and the controls silently stop clearing it. Your own test does catch that, but as a `waitForFunction` timeout on the 320/32 width wait, which reads as a hang rather than "the controls moved".

A pair of custom properties on a shared ancestor (`--map-pane-w`, `--map-pane-w-expanded`) used by both rules would make the coupling explicit and let one edit move both. Fine to leave as is if you would rather not widen the diff.

## Note for anyone carrying a fork

`public/map.js` is currently byte-identical between upstream and the ON8AR fork (the `safeEsc` work has been upstreamed), so this merges into a fork of that shape with no conflict.

Approving.

Issue #1833, opened 2026-07-08
Legend toggle button (palette icon) is hidden behind the bottom bar on Live view


anieto:
Running the latest version (v3.9.2). On the Live map page, there's no way to hide the "Packet Types" legend box in the bottom right — I looked everywhere in the settings/customize panel and couldn't find a way to turn it off either.

I poked through the CSS a bit and think I found why. The legend panel itself and the "show feed" button both have their bottom position set based on how tall the playback bar at the bottom of the screen is (`var(--vcr-bar-height...)`), so they sit above it correctly. But the actual toggle button for the legend (`.legend-toggle-btn` in `live.css`) is just hardcoded to `bottom: 1rem`, without that same adjustment. So on a normal desktop window, it looks like it's sitting right underneath the playback bar — which would explain why it's invisible.

This looks like basically the same bug as #685 from a few months back (legend getting covered by the VCR bar), just on a different element that must've gotten missed when that got fixed.

For now I worked around it by running this in the console:

```js
localStorage.setItem('live-legend-hidden', 'true'); location.reload();
```

That hides the legend, so the toggle does exist and work, it's just not clickable/visible where it's supposed to be.

Kpa-clawbot:
**Severity:** P2 — legend palette toggle unreachable behind VCR bar on Live view; localStorage workaround exists.

**Claim:** confirmed.

**Code today:** `public/live.css:1311-1314` — `.legend-toggle-btn { position: fixed; bottom: 1rem; right: 1rem; ... }`. Peer overlays correctly use `bottom: calc(var(--vcr-bar-height, 58px) + 10px)` (see `.live-feed` L1204, `.live-overlay[data-position="br"]` L1372, `.feed-show-btn` L903). Toggle button was missed.

**Findings:**
- (tufte): The palette toggle IS the affordance for user control of the legend. Hiding it behind a fixed chrome bar is a graphical-integrity failure — the control exists but the viewer cannot perceive it. Same class of defect as #685/#1206/#1107. Fix is one line; leaving it broken teaches users the app is untrustworthy.
[...]

Issue #1107, opened 2026-05-05
Fix layout on Live view to make the main live window with options not be as big, and fix better button placement on the bottom.


Bjorkan:
This is very big, bigger than needed, which covers the map:
<img width="853" height="164" alt="Image" src="https://github.com/user-attachments/assets/f8658385-26b9-40de-9409-76340d2254db" />

Buttons look a bit cramped, maybe options to activate/hide both should be bottom right and locked there so it feels more clean?
<img width="1911" height="87" alt="Image" src="https://github.com/user-attachments/assets/86876a84-39b6-455b-b398-e5bb2055e2cb" />

Kpa-clawbot:
Decomposed into two sub-issues:

- #1178 — compact `.live-header` (covers-the-map issue)
- #1179 — pin `.live-controls` bottom-right + collapse to toggle on narrow viewports

Locked design decision (per user): at narrow viewports, both surfaces collapse into a single toggle (header → expandable stats; controls → expandable sheet/popover). Implementation will batch into one PR; tracked separately for clarity.

Kpa-clawbot:
**Triage — Live view layout: panel too big + cramped buttons**

**Severity:** P3 (UX polish on Live view)

**Reporter claims:** Confirmed via screenshots — Live options panel takes more vertical space than necessary; bottom button row is cramped and inconsistent.

**Recommended path:** Two-part polish —
  1. Audit Live options panel: collapse rarely-used controls into a popover or reduce padding/font; goal is ≤ 50% current height
  2. Bottom button row: lock to bottom-right, consistent spacing, group the activate/hide pair

Needs a `tufte` pass on the IA before coding — these are layout/IA decisions, not pure CSS tweaks. Verify the changes against the existing breakpoints (#1396 / #1402 territory for nav-related regressions).

**Surfaces involved:** `public/live.html` and associated CSS, `public/live.js`.

**Duplicate of:** none found.

cc @Bjorkan: thanks — would you be willing to sketch a target layout (even a hand-marked screenshot)? Would speed up the design pass considerably.

Kpa-clawbot:
**Severity:** P3 — UMBRELLA, both children shipped.

**Claim:** confirmed. Decomposition landed.

[...]

PR #2049, opened 2026-09-20
fix(live): anchor the legend toggle to .live-page so the VCR bar stops eating its clicks

liquidraver:
## Problem

The PACKET TYPES legend cannot be dismissed on the Live view — clicking the palette toggle does nothing. The reporter found that running this in the console *does* work:

```js
document.querySelector('#legendToggleBtn').click()
```

That asymmetry is the whole diagnosis: a synthesised `.click()` skips hit-testing, so the handler is fine and something is painted over the button.

## Root cause

This is the residual half of #1833, reaching the user by a different route.

#1833 fixed the button's hard-coded `bottom: 1rem` by switching it to `calc(var(--vcr-bar-height, 58px) + 10px)`. The **offset** became correct; the **anchor** stayed wrong:

- `.legend-toggle-btn` and `.feed-show-btn` are `position: fixed` → their `bottom` resolves against the **viewport**.
- `.vcr-bar`, the thing that offset exists to clear, is `position: absolute` inside `.live-page`.
- `.live-page` is `height: calc(100dvh - 52px - var(--bottom-nav-reserve, 0px))`, and `bottom-nav.css` sets that token to `56px + safe-area` at `≤768`.

So at `≤768` the two anchors disagree by exactly the reserve, and **both** buttons drop by it. `.vcr-bar` is `z-index: 1000` against their `500` and takes pointer events.

Measured in Chromium against the real stylesheet at `reserve = 56px` (bar 50px tall), as offsets from `.live-page`'s bottom edge:

| | reserve 0 | reserve 56 | result |
|---|---|---|---|
| `.legend-toggle-btn` | `68` | `12` | **inside the bar** — clicks swallowed |
| `.feed-show-btn` | `122` | `66` | ~16px clear of it — still clickable |

`elementFromPoint` at the legend toggle's centre returned `DIV#vcrBar` with `0` clicks registered; after the fix it returns the button and the click lands.

The window is **641–768px**: at `≤640` the feed, the legend and both buttons are `display:none`, above `768` the reserve is `0` and nothing moves. That is why neither the desktop nor the phone layout tests caught it.

## Fix

Anchor both buttons with `position: absolute` so they resolve their offsets against the same containing block as the bar they clear — the offset then cannot drift from the bar again, whatever the reserve formula does later. `.live-page` is `position: relative` and spans the whole map area, so the docked bottom-right placement #1107 asked for is unchanged.

`.feed-show-btn` carries the identical defect on the identical anchor and is displaced by the full reserve in the band. It was **not** a second user-visible break: its `--legend-toggle-stack` offset parks it a row above the legend toggle, and that row absorbed the 56px drop, leaving it ~16px short of the bar. Same bug, 16px from the same symptom.

Desktop geometry is unchanged (verified at `--bottom-nav-reserve: 0`).

## Tests

- **`tests/e2e/test-issue-1833-legend-toggle-clickable-e2e.js`** (new) — at `720x900`, for **both** buttons: (a) the reserve band is genuinely active, so the test cannot silently pass on a desktop layout, (b) visible, (c) clears the bar, (d) `elementFromPoint` resolves to it, (e) a **real** hit-tested `page.click()` works. (f) is the anchor invariant — each button's offset from `.live-page`'s bottom edge, net of the bar height, must be the same at 720 (reserve on) and 1440 (reserve off); under `fixed` those differ by the reserve, under `absolute` they match. That is what covers `.feed-show-btn`, whose hit test passes on the buggy CSS too. (g) re-checks desktop.
- Wired into `.github/workflows/deploy.yml` next to the other live-view suites, and listed in `scripts/non-unit-tests.json`.
- **`test-issue-1833-legend-toggle-vcr-offset.js`** — pins the anchor next to the existing offset assertion. Fails on the old CSS (`got: fixed`), passes on the new.
- **`test-issue-1107-live-layout.js`** — its `position: fixed` invariant encoded the bug, so it is updated to `absolute` with the reasoning inline. #1107's actual intent (one docked bottom-right cluster) is preserved.

Unit suite: the affected suites pass. `test-all.sh` exits 1 on this machine at the xss-gate self-test with a `cp1252` `UnicodeEncodeError` — clean `master` in a worktree gives a byte-identical set of failures, so it is a Windows-environment issue, not this branch.

The Playwright E2E was **not** executed locally (no `node_modules` and no local fixture server on this machine) — the geometry and click assertions it encodes were verified by hand in Chromium against the real stylesheet, as in the table above, but CI is the first run of the test itself.

Reported by AmPeR, who also found the console workaround.


efiten (changes requested):
The diagnosis is right and I checked each step of it rather than taking the write-up's word:

- `.live-page` is `position: relative` with `height: calc(100dvh - 52px - var(--bottom-nav-reserve, 0px))` (`live.css:8-11`), so it is a containing block and it does end short of the viewport by the reserve.
- `.vcr-bar` is `position: absolute; bottom: 0; z-index: 1000` (`live.css:923-927`) against the button's `500`, so it wins the hit test.
- Both buttons are `display: none !important` at `≤640` (`live.css:611,613`), which is why the band is 641 to 768 and why neither the desktop nor the phone layout test saw it.

Anchoring to the same containing block as the bar is the right fix rather than another offset correction: it cannot drift again whatever the reserve formula becomes.

## Must fix: the new suite never runs

`tests/e2e/test-issue-1833-legend-toggle-clickable-e2e.js` is 204 lines and is listed in `scripts/non-unit-tests.json`, but there is no line for it in `.github/workflows/deploy.yml`. That file only exempts a suite from the unit runner; it schedules nothing. Checked on this branch: `deploy.yml` mentions it 0 times.

That is the gap tracked in #2037, where 18 suites sat in exactly this position, and #2045 wired in the eight that still passed. One line next to the others fixes it:

```
BASE_URL=http://localhost:13581 node tests/e2e/test-issue-1833-legend-toggle-clickable-e2e.js 2>&1 | tee -a e2e-output.txt
```

Without it, the `elementFromPoint` check at 720px that makes this PR convincing never executes again after today.

## The `.feed-show-btn` reasoning is wrong, and the change is better than claimed

The PR says it is "`display:none` at `≤768`, so that change is behaviour-neutral today". It is hidden at `≤640`, not `≤768`: the only rules are `live.css:611` inside `@media (max-width: 640px)` and the unrelated `body.live-fullscreen` one at `:528`.

So `.feed-show-btn` is **visible in the affected band**, and it carries the same defect:

```css
.feed-show-btn {
  position: fixed;
  bottom: max(1rem, calc(var(--vcr-bar-height, 0px) + 10px + var(--legend-toggle-stack)));
  z-index: 500;
}
```

Same `fixed` anchor, same VCR-derived offset, same losing z-index. It is not a tidy-up for symmetry, it is a second instance of the bug, fixed here.

That matters for coverage rather than correctness. The unit suite now pins it as `position: absolute` in the stylesheet text, which stops a regression to `fixed`, but nothing exercises it: the E2E asserts `elementFromPoint` on `#legendToggleBtn` only. The viewport block at 720px is already set up, so asserting the same for `#feedShowBtn` is a few lines in a test that has done the hard part. Worth doing while you know why it is there.

## Verified

The unit suite you extend (`tests/unit/test-issue-1833-legend-toggle-vcr-offset.js`) is listed in `test-all.sh`, so those 42 lines do run. All checks green on this branch.

Requesting changes for the one line in `deploy.yml`; the rest is a suggestion I would not hold the PR for, though correcting the `≤768` claim in the body would be good since it is the reason the second button is touched.

liquidraver:
Review feedback addressed (commit `e0fc9ab3`).

**1. The suite never ran — fixed.** You are right, and I checked it the same way: `grep -c 1833 .github/workflows/deploy.yml` returns 0 on the previous commit. I had taken `scripts/non-unit-tests.json` for a schedule when it is only an exemption from the unit runner. Added next to the other live-view suites, after `test-issue-1206-vcr-overlap-e2e.js`:

```
CHROMIUM_REQUIRE=1 BASE_URL=http://localhost:13581 node tests/e2e/test-issue-1833-legend-toggle-clickable-e2e.js 2>&1 | tee -a e2e-output.txt
```

`CHROMIUM_REQUIRE=1` to match its neighbours, so a broken Chromium fails rather than skips.

**2. The `≤768` claim was wrong.** Confirmed: the only rules are `live.css:611` inside `@media (max-width: 640px)` and the fullscreen one at `:528`. I misread the enclosing media query. Corrected in the PR body and in the CSS comment, which repeated it.

**3. On it being a second instance — measured, and it is more interesting than either of us said.** You are right that it is visible in the band and carries the same defect, so I measured both buttons on master's CSS rather than reasoning about it. Offsets from `.live-page`'s bottom edge, bar 50px tall:

| | reserve 0 | reserve 56 | result |
|---|---|---|---|
| `.legend-toggle-btn` | `68` | `12` | inside the bar, clicks swallowed |
| `.feed-show-btn` | `122` | `66` | ~16px clear of it, still clickable |

Both drop by the full reserve — same anchor, same defect, and `.feed-show-btn` is not a tidy-up. But it was not a second *break*: its `--legend-toggle-stack` offset parks it a row above the legend toggle, and that row is what absorbed the 56px. It missed the bar by 16px. So the fix there prevents a break rather than repairs one, which I would rather state precisely than let the PR claim either "behaviour-neutral" or "second instance of the bug".

**4. Coverage for it.** The E2E now probes both buttons at 720 (`#feedShowBtn` revealed by clicking `#feedHideBtn` first). Since its hit test passes on the buggy CSS too, a hit test alone would not have caught it, so I added the invariant that actually separates fix from bug — **(f)**: each button's offset from `.live-page`'s bottom edge, net of the bar height, must be equal at 720 (reserve on) and 1440 (reserve off). Under `fixed` those differ by the reserve; under `absolute` they match. It fails on master's CSS for both buttons.

Still worth flagging: I have no `node_modules` or local fixture server here, so this suite has never executed — the numbers above come from driving the real stylesheet in Chromium with a DOM harness. Its first real run will be on your CI, and the `#feedHideBtn` reveal step is the part most likely to need adjusting against the real page.

efiten:
Both points landed, and wiring it in did exactly what it was supposed to do: the suite failed on its first real execution, on itself rather than on the product.

## The failure is assertion (a), and it is a test bug

```
FAIL: (a) expected a non-zero --bottom-nav-reserve at 720px wide
         (got "calc(56px + 0px)", gap 58px) — test is not exercising the regressed band
PASS: (b) ... (g)   [13 of 14]
```

`getComputedStyle(el).getPropertyValue('--bottom-nav-reserve')` on an **unregistered** custom property returns the token sequence after `var()` substitution, not an evaluated length. `calc()` is left unevaluated and `env(safe-area-inset-bottom, 0px)` resolved to its fallback, so the string is literally `calc(56px + 0px)`. There is no number in it to be non-zero. `bottom-nav.css:45` declares it exactly that way, so this is the expected return value, not a browser quirk.

The band was active the whole time: the same `page.evaluate` measured `gapBelowLivePage: 58`, and every assertion that depends on the reserve being applied passed, including **(f)**, the one that separates `fixed` from `absolute`:

```
PASS (f) #legendToggleBtn offset is reserve-independent (slack 11px at 720 vs 11px at 1440)
PASS (f) #feedShowBtn   offset is reserve-independent (slack 123px at 720 vs 123px at 1440)
```

Under the old CSS those two numbers differ by the reserve. They match. The fix is demonstrated; the guard that was supposed to confirm the test is honest is the only thing that is not.

## Suggested fix

Assert the effect rather than the token. You already compute it:

```js
if (narrow.gapBelowLivePage > 0) pass('(a) reserve band active (.live-page ends ' + narrow.gapBelowLivePage + 'px above the viewport bottom)');
```

That is also the condition (f) actually depends on: that `.live-page` ends short of the viewport. If you would rather keep reading the property, resolve it by putting `height: var(--bottom-nav-reserve)` on a throwaway element and measuring it, which forces evaluation.

One thing not to do: assert `gap === 56`. The measured gap is **58px**, so something contributes 2px beyond the token, and I have not chased what. A `> 0` check does not care.

## On the rest

Your correction of the `.feed-show-btn` case is better than what either of us wrote: same anchor, same defect, but it missed the bar by 16px because `--legend-toggle-stack` parks it a row higher, so the change prevents a break rather than repairs one. Worth stating precisely, and you did.

Noted that this suite had never executed before this run and that you said so up front. That is the right way round: the CI line was the missing piece, and one run later it has already paid for itself.

Re-approving once (a) is green.

liquidraver:
Fixed in `611155a9` — same diagnosis you reached, and I confirmed the mechanism in Chromium against that exact declaration before changing anything:

```js
document.documentElement.style.setProperty(
  '--bottom-nav-reserve', 'calc(56px + env(safe-area-inset-bottom, 0px))');

getComputedStyle(document.documentElement)
  .getPropertyValue('--bottom-nav-reserve')   // "calc(56px + 0px)"
parseFloat("calc(56px + 0px)")                 // NaN

// throwaway element with height: var(--bottom-nav-reserve)
probe.getBoundingClientRect().height            // 56
```

So the returned string is the declaration, unevaluated, exactly as you said — not a browser quirk and nothing to do with the viewport.

**(a) now asserts `gapBelowLivePage > 0`.** Your framing is the right one: that is the whole precondition the bug needs, it is what (f) depends on, and it is measured rather than parsed. The token and its resolved length are still captured, but only for the failure message, so a future reader can see both numbers without anything asserting on them.

**Not pinning 56.** Taken, and I left the reasoning inline so the next person does not tighten it:

```js
// Deliberately `> 0` and not `=== 56`: the gap measures 58px on CI
// against a 56px token, and the extra 2px is unexplained (the height is
// derived from 100dvh, which need not agree with innerHeight to the
// pixel). Pinning the exact number would make this assertion fragile
// for no gain — the bug does not care how far short the box ends, only
// that it does.
```

I have not chased the 2px either. `100dvh` not agreeing with `innerHeight` to the pixel is the obvious suspect, but I would be guessing, and it does not affect anything here: every assertion that matters is a comparison between two measurements taken in the same run, so a constant offset cancels.

On the run itself — agreed, and worth saying plainly: the CI line was the missing piece, the suite's first execution caught a defect in my own guard, and the product assertions it was guarding passed. I would rather it surfaced that way than have shipped a test that could only ever pass.

liquidraver:
Green — `611155a9`, all 4 checks passing, and the suite is 14/14:

```
PASS: (a) reserve band active — .live-page ends 58px above the viewport bottom
          (token="calc(56px + 0px)", resolved=56px, gap=58px)
...
test-issue-1833-legend-toggle-clickable-e2e.js: 14 passed, 0 failed
```

The diagnostics land as intended: the raw token is still the unevaluated `calc(56px + 0px)`, the probe element resolves it to `56px`, and the assertion runs off the measured `58px` gap. All three visible in one line, none of them asserted except the gap — so the 2px stays on the record for whoever does chase it, without being able to break the suite.

efiten (approved):
Green, and the suite now reads the way it should:

```
PASS: (a) reserve band active — .live-page ends 58px above the viewport bottom
          (token="calc(56px + 0px)", resolved=56px, gap=58px)
...
test-issue-1833-legend-toggle-clickable-e2e.js: 14 passed, 0 failed
```

Keeping the token and the resolved length in the message while asserting only on the gap is better than either of the versions we discussed: the next person sees all three numbers and why they differ, without any of them being load-bearing.

For the record, since three runs of one suite told us more than the diff did:

- it never ran before this PR, and the first real execution failed on its own setup rather than on the product,
- the discriminating assertion **(f)** has passed on every run, at 11px and 123px slack identical between 720 and 1440, which is what proves the anchor change did its job,
- and (g) shows the desktop layout unmoved, which is the part a CSS change like this most easily breaks silently.

Approving.

Issue #1011, opened 2026-05-03
perf: lazy distance index — build on first request, not startup


Kpa-clawbot:
## Context

Split from #836 per startup audit (`specs/startup-audit.md`).

## Problem

Distance index is built eagerly on startup for all node pairs. At scale (2K+ nodes) this is O(n²) pairs × observation scan. Most operators never use the distance analytics page.

## Proposed fix

Build distance index lazily — only compute when `/api/analytics/distance` is first requested. Cache the result. Invalidate on new observations (debounced, not per-packet).

## Acceptance criteria

- Distance index NOT built at startup
- First request to distance endpoint triggers build (may be slow, return 202 + retry-after, or stream)
- Subsequent requests use cache
- Cache invalidated periodically (e.g., every 5 min if new data arrived)
- Startup time reduced by ~10-20s at 500K obs

## Related

- #836 (parent)
- `specs/startup-audit.md`

Kpa-clawbot:
**Severity:** P3 — startup-only cost; affects operators who never visit distance analytics.

**Claim:** confirmed by parent audit `specs/startup-audit.md`. Distance index is O(n²) pair-construction at startup regardless of whether any client requests `/api/analytics/distance`.

**Code today:** distance index built eagerly during `Load()` post-pass. No first-request hook; no invalidation policy beyond full rebuild.

**Findings:**
- (carmack): lazy-on-first-request is the right shape, but the first request will block for the full O(n²) build — return 202 + `Retry-After: N` or stream a progress object; do NOT silently block the HTTP handler for 10-20s while the user stares at a spinner.
- (kent-beck): test the no-request-no-build invariant directly — boot the server, scrape `/api/perf`, assert distance-index-built-count = 0; hit the endpoint once, assert = 1; hit again, assert still = 1 (cached). Then assert invalidation after N new observations. That's the contract.

**Fix path:** Move distance-index build out of `Load()` post-pass into a `sync.Once` triggered by first `/api/analytics/distance` request. Cache result; invalidate via debounced observation-counter (e.g., rebuild if Δobs > 5% since last build, max once per 5 min). First request returns 202+Retry-After while building. Effort: M.

**Dup check:** none. Parent: #836. Sibling splits: #1009 (chunked Load), #1010 (GOMEMLIMIT).

**Triage verdict:** ready-to-fix

Kpa-clawbot:
[...]

PR #2051, opened 2026-09-20
fix(analytics): show a building state on the distance index's 202, and stop caching it

### LINKED ISSUE #1997: bug(analytics): Distance tab crashes on the initial 202 index-building response
Found while replacing fixed browser sleeps with page-readiness checks for #1995, against `b5c7612`.

## Reproduction

1. Start an instance whose lazy distance index has not been requested.
2. Open Analytics → Distance.

The first `/api/analytics/distance` request correctly returns HTTP **202**, `Retry-After: 5`, and a body containing `status: "building"` and `retry_after_seconds: 5`, without `summary`.

`renderDistanceTab` treats that response as final data and accesses `data.summary.totalHops`. The page displays:

`Failed to load distance analytics: Cannot read properties of undefined (reading 'totalHops')`

A later API request returns a valid summary. A fresh fixture host independently reproduced the initial 202 response. The older accessibility sweep could report zero violations on the error view; the new readiness gate rejects it.

Source: [distance renderer](https://github.com/Kpa-clawbot/CoreScope/blob/b5c7612166e8e70a7a4075a209a786c12199f464/public/analytics.js#L2884). The backend's lazy-index 202 contract is already covered by `cmd/server/distance_lazy_index_test.go`.

## Expected / acceptance

- Show a building/loading state while the index is pending, then render the summary automatically.
- Respect the retry interval and avoid overlapping or unbounded requests.
- Stop obsolete work when the user changes tab or leaves the page.
- Add a browser regression for 202 followed by 200, plus navigation away while pending.

Keep the backend's nonblocking lazy-index behavior intact. This correction belongs with the audit's new browser-readiness coverage.

efiten:
Closes #1997.

## What was wrong

`/api/analytics/distance` answers `202 {status:"building", retry_after_seconds:5}` with no `summary` until the lazy index (#1011) has been built.

1. `renderDistanceTab` read `data.summary.totalHops` straight away. The TypeError is caught by the tab's own try/catch, so it never reaches `window.onerror`: it is painted into the tab as `Failed to load distance analytics: Cannot read properties of undefined (reading 'totalHops')`.
2. `api()` caches any `res.ok` body, and `res.ok` is true for 202, so the placeholder was stored for the `analyticsRF` TTL. Even a correct retry read the cached "building" body back. That is the half that made the broken state outlast the index build.

## What changed

- `public/app.js:173` skips the cache write when `res.status === 202`. A 200 still caches, unchanged.
- `public/analytics.js` renders a building notice and retries itself, honouring `retry_after_seconds` clamped to [1s, 30s]. The timer is cleared on tab switch and in `destroy()`, and a new render supersedes a pending retry, so two renders cannot write into the same tab.

The notice uses `.text-center`/`.text-muted` rather than the `.spinner` class used at `analytics.js:1683`, because `.spinner` has no CSS anywhere in the repo and renders nothing.

## Tests

`tests/unit/test-issue-1997-distance-building.js` (7 assertions, wired into `test-all.sh`) pins the two pure decisions the renderer makes and `api()`'s refusal to cache a 202 while still caching a 200. Red-run on the unfixed sources: 6 of 7 fail, and the "a 200 is still cached" control stays green.

`tests/e2e/test-issue-1997-distance-building-e2e.js` (classified in `scripts/non-unit-tests.json`, invoked from `deploy.yml` with `CHROMIUM_REQUIRE=1`) serves both responses by route interception, so it does not depend on whether the server under test has an index built. It asserts: the building state appears, the tab does not paint the error text, a retry arrives with no interaction, the retry replaces the placeholder once the server answers 200, and no retry fires after leaving the tab.

Check (2) deliberately asserts on the rendered text and not on `pageerror`: the TypeError is caught, so a `pageerror` assertion would pass on the broken build too.

## Verification

No local cgo toolchain here since #1992, so I could not build a server to run the E2E against. Instead I ran its five steps in Playwright against a live instance with this branch's `public/app.js` and `public/analytics.js` injected in place of the deployed ones (both files are byte-identical between that instance and upstream master, so the injection is faithful):

| check | deployed build | this branch |
|---|---|---|
| (1) building state shown | fail | pass |
| (2) not rendered as data | fail | pass |
| (3) retried on its own | fail (1 request) | pass (2 requests) |
| (4) real payload after retry | fail | pass |
| (5) no retry after leaving the tab | pass | pass |

(5) passes on the broken build too: it schedules no retry at all, so it is a control and only means anything together with (3).

The committed E2E suite has not been run as committed. CI is its first real run.

## Not done

- The server still recomputes on every 202 poll rather than signalling readiness.
- No other analytics tab was audited for the same assume-a-summary pattern.
- One pre-existing unit suite (`test-preflight-xss-gate.js`) fails on this Windows machine with a cp1252 `UnicodeEncodeError` from its Python helper, on a clean tree as well as with this change. Unrelated, and green on Linux CI.


efiten:
CI ran the new E2E suite for the first time (run 35512421581, Playwright job, 13:23:22-13:23:28Z):

```
=== E2E SUITE: test-issue-1997-distance-building-e2e.js ===
  ✓ (1) the 202 renders the building state
  ✓ (2) the building body is not rendered as data
  ✓ (3) the tab retried on its own (2 requests)
  ✓ (4) the retry replaces the placeholder with the real payload
  ✓ (5) no retry fires after leaving the tab
test-issue-1997-distance-building-e2e: 5 passed, 0 failed
```

All five assertions executed against the locally built server, so the suite is wired and not silently skipping. That closes the one open item in the PR description.

Issue #2052, opened 2026-09-20
style.css declares .nav-btn and .ch-icon-btn at both 48px and 44px


efiten:
Both selectors carry two conflicting minimum-size declarations in `public/style.css`. The later one wins, so 44px is what ships while the earlier block claims 48px.

| selector | first declaration | later declaration | effective |
|---|---|---|---|
| `.nav-btn` | `:561` `min-height/min-width: 48px` | `:890` `min-width: 44px; min-height: 44px` | 44 |
| `.ch-icon-btn` | `:567` `min-height/min-width: 48px` | `:1838` `min-width: 44px; min-height: 44px` | 44 |

The two blocks also disagree in their reasoning. The 48px block at `:610` says "Same 48x48 minimums + touch-action so all interactive surfaces meet WCAG 2.5.5", while the 44px rule at `:1836` says "WCAG 2.5.5 / Apple HIG: 44x44 CSS px minimum touch target". WCAG 2.5.5 is 44x44, so the later comment cites it correctly and the earlier one over-claims: 48 is the Material figure, not the WCAG one.

Nothing is visually broken today, and I am not proposing a size change: which number the project wants is a design call, not a lint. What is wrong is that the stylesheet states both, so a reader cannot tell what the rule is and a future edit to either block silently changes behaviour or silently does nothing.

Two ways out, both cheap:

1. Keep 44 (what ships, and what WCAG asks). Remove `.nav-btn` and `.ch-icon-btn` from the 48px blocks at `:561` and `:567`, and correct the `:610` comment to say the group is 48 by house preference rather than by WCAG 2.5.5.
2. Keep 48 (the house default for every other control in that block). Remove the `min-width`/`min-height` from the component rules at `:890` and `:1838`. This does change the rendered size of the navbar buttons and the channel row icons, so it wants a look on mobile first.

Found while reviving `tests/e2e/test-touch-targets.js`, which had gone unrun (#2037). That suite demanded a blanket 48 and so failed on exactly these two. Rather than lower the blanket or hide the failures, it now has a `MIN_OVERRIDES` table pinning these two at their effective 44 with a pointer here, so a third selector dropping to 44 still fails the build. Whichever option is taken above, that table entry should be removed in the same change.

Issue #1648, opened 2026-06-11
Migrate emoji UI to Phosphor Icons (regular weight) — tracking


Kpa-clawbot:
## Summary

Replace the emoji glyphs we are currently using as **UI iconography** across the CoreScope frontend with proper SVG icons from [Phosphor Icons](https://phosphoricons.com/) (regular weight). Emoji-as-icon is inconsistent across platforms (Apple/Google/Windows/Android render the same codepoint very differently), can't be themed via `currentColor`, and looks dated next to the rest of the UI.

This is a **tracking issue** for a multi-PR migration. Each milestone below should land as its own PR so we can ship incrementally and revert cleanly.

**Audit total: ~448 emoji-as-icon surfaces** across 32 files (excludes log lines, code comments, test fixtures, channel-name text, and other intentional text-emoji usage).

## Bundling decision (proposal — confirm before M1)

**Recommended: self-host an SVG sprite sheet.**

- Vendor only the icons we use into `public/icons/phosphor-sprite.svg` (one `<symbol id="ph-search">…</symbol>` per icon, regular weight only).
- Markup pattern: `<svg class="ph"><use href="/icons/phosphor-sprite.svg#ph-search"/></svg>`
- CSS: `.ph { width: 1em; height: 1em; fill: currentColor; vertical-align: -0.125em; }`
- Themable via `currentColor`, no FOUT, no CDN, ~6 KB gzipped for ~25–30 icons.
- One HTTP fetch (cacheable), works with our `__BUST__` cache-buster.

**Rejected alternative:** full Phosphor webfont (~150 KB, opaque to text selection, font loading flicker, harder to color individual icons). Counter only if there's a specific reason.

**Out of scope for this issue:** picking the exact SVG export script, naming convention beyond `ph-<kebab>`, or a build step. Keep it static — drop the sprite into `public/icons/` and reference by ID.

## Migration milestones

Ordered by user-visibility × blast radius. M1 proves the bundling pattern; subsequent milestones reuse it.

### M1 — Top-nav, bottom-nav, drawer, Compare entry points

Highest visibility, smallest blast radius. Establishes the sprite sheet and the `<svg><use>` pattern.

- [ ] `public/index.html` L123 🔴 → `ph-broadcast` (Live nav link)
- [ ] `public/index.html` L129 ⚡ → `ph-lightning` (Perf nav link)
[...]

Issue #1646, opened 2026-06-11
polish: observer compare follow-ups from #1645 visual review


Kpa-clawbot:
## polish: observer compare follow-ups from #1645 visual review

PR #1645 shipped the Tufte redesign of the observer-comparison surface. Visual review on staging surfaced 5 polish items from the parent reviewer + 3 additional findings from a Tufte pass. All items collapse into one follow-up PR (#1646 → see below).

### Coalesced findings (each item has tag, file, fix, screenshot evidence)

| # | Tag | Item | File:line | Fix |
|---|---|---|---|---|
| 1 | [both] | Native `<input type="checkbox">` rendered as bare white squares on the dark theme — observers list + everywhere | `public/style.css` line ~452 (new); `:root` blocks | Add a global `input[type=checkbox],input[type=radio] { accent-color: var(--accent); }` rule + `color-scheme: dark` on dark-theme blocks (themes UA scrollbars / selects too) |
| 2 | [parent] | "Compare" CTA next to the dropdowns is heavy-blue primary; redundant once both dropdowns have values | `public/compare.js` line 171; `public/style.css` line 3471 | Render `#compareBtn` as `.btn-ghost`; once both observers selected, CSS hides the button (the dropdown change IS the trigger). Drop the now-dead `.compare-btn` rule. |
| 3 | [parent] | "vs" label between dropdowns is at parity with the dropdowns | `public/style.css` `.compare-vs` block | Drop to 10px, center vertically, opacity 0.7, +letter-spacing |
| 4 | [parent] | SHARED middle column had three competing font weights and the count outshone the percentage | `public/compare.js` line 335; `public/style.css` mid-* blocks | Invert hierarchy: 87% gets `var(--fs-xl)` accent (the answer); "3,452 shared" demotes to `var(--fs-sm)`; "OF ALL UNIQUE" stays at 10px caps. New `.compare-strip-mid-pct` + `.compare-strip-mid-pct-unit` rules. |
| 5 | [parent] | Selector strip above the headline competes for "look here first" attention | `public/compare.js` `updateBtn`; `public/style.css` `.compare-controls.is-collapsed` | Add `is-collapsed` / `data-collapsed="true"` to `#compareControls` when both observers selected. CSS demotes: smaller padding, labels hidden, Compare button hidden, smaller dropdown font. Operators can still swap A/B via the dropdowns. |
[...]

Issue #1522, opened 2026-06-01
Trace tool: packet hash not reflected in URL (unsharable results)


syssi:
## Problem

When a user performs a trace in `#/tools/trace/`, the resulting packet hash is only held in component state — the URL stays at `#/tools/trace/` regardless of which hash was traced. Sharing the URL sends the recipient to an empty input, not to the trace result.

The `init()` function already reads a hash from the URL on page load (`routeParam` or `?hash=` query param), so the read path works — the write path is simply missing.

## Expected behavior

After tracing a hash, the URL updates to `#/tools/trace/<hash>`. Sharing or bookmarking that URL takes the recipient directly to the same trace result.

## Proposed fix

In `doTrace()` inside `traces.js`, call `history.replaceState` immediately after the hash is validated:

```js
currentHash = hash;
history.replaceState(null, '', `#/tools/trace/${encodeURIComponent(hash)}`);
```

One line — no architectural changes needed. The route parser already handles `#/tools/trace/<hash>` (incl. backward-compat redirect from the old `#/traces/<hash>` format).

Kpa-clawbot:
**Severity:** P3 — shareability gap, no breakage.

**Claim:** confirmed. `public/traces.js:53` sets `currentHash = hash` but never touches `location.hash`. The read path already works: `init(app, routeParam)` at `public/traces.js:8-30` accepts a hash from the route param or `?hash=` query and auto-fires `doTrace()` (`:30`). The backward-compat redirect `#/traces/<h>` → `#/tools/trace/<h>` is wired at `public/app.js:866-868`. Asymmetric: deep-links IN work, links OUT don't get generated.

**Code today:**
- `public/traces.js:49-53` — `doTrace()` validates hash, sets local state, no URL sync.
- `public/app.js:907` — tools sub-router parses `tools/trace/<hash>` and passes it as `routeParam` to `init()`.

**Findings:**
- (tufte): URL is the canonical "you are here" affordance for a SPA — when a view holds meaningful state (the traced hash IS the view), the URL must mirror it. Otherwise the share/bookmark/back-button mental model breaks silently. The fix restores parity with every other deep-link surface in the app (`#/map?packet=`, `#/packets/<h>`).
- (kent-beck): testable via a 3-line DOM test — set input, click `#traceBtn`, assert `location.hash === '#/tools/trace/<h>'` after the await. No mock needed beyond `fetch`. Belongs in `tests/issues/test-issue-1522.js`.

[...]

PR #2053, opened 2026-09-20
test: make every suite in tests/e2e run, and tell the truth when it does

### LINKED ISSUE #2037: 18 suites in tests/e2e are invoked by no runner, and one cannot run at all

efiten:
Closes #2037.

Step 1 (#2045) wired in the nine suites that already passed. This is steps 2 and 3: the four that ran and failed, and the five that could not run at all. After this, the count of suites in `tests/e2e` invoked by nothing goes from 18 to 0.

## Step 2 — the four that ran and failed

Triaged against a fixture server in CI with full output kept, not the four-line tail the first probe saved.

| suite | verdict |
|---|---|
| `test-packets-scope-column.js` | passes on master today. It failed on 2026-09-18, so something between the two fixed it. Wired in unchanged rather than investigated. |
| `test-node-reach-e2e.js` | test wrong. It waited for the reach map whenever any *link* had GPS; `public/node-reach.js` builds the map only when the *node* has coordinates. Guaranteed 10s timeout on a node with positioned neighbours and no position of its own. |
| `test-channel-modal-e2e.js` | both failures test-side. The Add button's visible label was shortened to "+ Add" with the accessible name moved to `aria-label` (`public/channels.js:746`); and `.ch-section-mychannels` is conditional on the visitor having added a channel, which has not happened at that point in the suite. |
| `test-touch-targets.js` | three test-side, two a product finding. |

The three test-side touch-target failures were the harness measuring controls that are not shown: `.compare-btn` (the CTA was removed in #1646, and `style.css` says so), `.ch-back-btn` (`display:none` outside the mobile channels layout), and `.filter-toggle-btn` (`display:none` on mobile since #1461; the control shown is the navbar mirror, which `mobile-page-actions.js:70` builds as a `.nav-btn`, so it was already measured). All three are dropped from the table with the reason recorded in the file.

The remaining two are **not** a test problem: `.nav-btn` and `.ch-icon-btn` are each declared twice in `public/style.css`, 48px in the touch-target block and 44px in their own component rule, and the later one wins. Rather than lower the blanket or hide the failures, the suite now has `DEFAULT_MIN = 48` plus a `MIN_OVERRIDES` table holding those two at their effective 44, so a third selector dropping to 44 still fails the build. The contradiction is **#2052**, with both ways out costed; the override entries should go when it is settled.

## Step 3 — the five that could not run

None is deleted. I checked each selector and seam against the product before deciding, and every one still targets a surface that exists and that nothing else covers.

Four were written against `@playwright/test`, a runner the project neither installs nor uses anywhere else. Adopting a second runner for twelve tests costs more than porting them, and the precedent is already set: `test-path-inspector-coverage-e2e.js` exists, as its own header says, because `test-path-inspector-e2e.js` could not run. So they are ported to the plain-node Chromium pattern the other 109 suites use.

- **`test-issue-1522-trace-url-sync-e2e.js`** — the trace hash in the URL, both directions. `test-e2e-playwright.js` covers that the page loads and searches; it never looks at the URL, which is the whole of #1522.
- **`test-marker-outline-weight.js`** — the canvas pulse ring never thins below 2px. There is no CSS rule to read and axe cannot see inside a canvas, so sampling the seam is the only way. Added a guard that the ring was actually visible, so the weight check cannot pass vacuously on a pulse that never rendered.
- **`test-pr-1490-live-map-gpu-animations-e2e.js`** — the queue drains, the engine sleeps again, the fading trails stay under the cap of 5, and the canvas sits on `animationsPane` rather than under the markers.
- **`test-path-inspector-e2e.js`** — reduced to what nothing else covers: the map side pane, the `/#/traces/<hash>` redirect, the tools landing. Its standalone-page test duplicated the wired coverage suite and is dropped. Its "switching candidate clears prior polyline" case ended after the click with a comment and no assertion, which is the same green-but-empty problem this issue is about; it now compares path counts, and skips loudly when the fixture yields too few candidates.

The fifth, **`test-table-sort.js`**, needed `jsdom`, which was declared nowhere. It is a unit test of `public/table-sort.js` filed under `tests/e2e`, so: `jsdom` is a devDependency (lockfile updated, `npm ci` stays consistent), the file moved to `tests/unit/`, and the `domIntegration` group in `scripts/non-unit-tests.json` is gone with its only member. It runs 22 tests. 20 passed immediately; 2 had rotted, because #1648 M2 replaced the up/down glyphs with Phosphor sprites and the direction moved out of `textContent` into the `<use href>`. Those two now read the sprite ref and the `aria-sort` value, so they also guard the accessible announcement.

## Verification

`tests/unit/test-table-sort.js` 22/22 and `test-test-inventory.js` pass locally; the E2E suites need a fixture server, which I cannot build here (no cgo toolchain since #1992), so CI is their first run as committed. The triage above was measured in CI, not assumed.

## Not done

The per-assertion skips named in the second comment on #2037 are untouched: the two flaky packet-detail cases, the fixture-data ones, and the two `clientRxCoverage` suites that skip wholesale while reporting success. Those need a fixture deployment with coverage enabled, which is its own change. I have not opened it.

Option 2 from the issue, making `test-test-inventory.js` require a `deploy.yml` line for every `tests/e2e` file, is also not here. It is the right guard and it is now enforceable, since the list is finally at zero, but it belongs in its own change where a red build means what it says.


efiten:
## Where this landed, after the first real runs

The suites had never run, so CI found things reading could not. Each is its own commit.

| what CI found | cause | fix |
|---|---|---|
| `test-table-sort.js` could not resolve `jsdom` | the JS unit gate has **no npm install at all**; all 179 unit suites use node builtins only. The original `domIntegration` reason said exactly this and I misread it | back to `tests/e2e`, own `deploy.yml` line in the job that runs `npm ci` |
| `test-node-reach-e2e.js` timed out at the map | Leaflet stamps `.leaflet-container` on the element handed to `L.map()`, so the class is on `#nqMap` itself. The selector had a space and waited for a descendant that never exists | selector corrected. My earlier GPS-condition change was right on its own but was not the bug |
| `test-pr-1490` queue never drained | the 1500ms budget came from the `@playwright/test` original, where it had never been exercised | 8000ms, with the reasoning recorded |
| `test-path-inspector-e2e.js` clicked a dead toggle | `initMapSidePane()` runs at the END of `loadNodes()` (`map.js:1795`), about 3s in. Clicking on sight is a race | click until the class appears, with real clicks so an overlay eating them still fails |
| `test-packets-scope-column.js` "No packets found" | **the packets page defaults to a 15-minute window.** CI freshens the fixture at job start and the E2E step runs longer than that, so a suite late in the list sees nothing. This PR moved it later by adding nine suites | the suite pins the window. Wider exposure is #2054 |
| `pickPrefixes` threw on a null pubkey | the graph records hops by prefix; the node behind one need not be known. The fixture has such entries | filtered |

Two corrections to the issue triage I posted earlier, both mine:

- I wrote that `test-packets-scope-column.js` "passes on master today, so something between 2026-09-18 and today fixed it". Wrong in both directions. It was never fixed and never broken; it is a function of how long the job has been running when it is reached.
- I wired the suites in and then reported "5 passed, 0 failed, 2 skipped" for the path inspector as if that were a result. The two skips were the only assertions about drawing a route, which is the green-but-covers-nothing shape this issue is about. Chasing them is the bulk of the commits above.

## The two skips that remain

`(4) Show on Map` and `(5) switching candidates` still skip on the fixture. What is established: the prefixes are valid (the endpoint answers 200), the graph is not empty (110 edges over 200 nodes), and the same selection returns 10 candidates against a populated instance where packet-path prefixes returned none. What is not established: why the beam search finds nothing in the fixture's graph. Seeding a chain it will score is fixture work, not a change to this suite, so I stopped there and the file records it.

If you would rather those two cases not ship as skips at all, say so and I will drop them from this file and open the fixture work separately.

PR #2057, opened 2026-09-21
fix(node-health): list only observers that heard the node on air

### LINKED ISSUE #2056: Node detail "Heard By" lists observers that never heard the node, with signal numbers that belong to another node
## What happens

The node detail page's "Heard By" card lists observers that never received anything from the node, and prints an average SNR and RSSI next to each of them.

It is easiest to see across frequency bands. `NL-BXE-RP01 | 433` is a 433 MHz repeater. Exactly one observer on the network runs on 433 MHz. Its card lists **13 observers**, twelve of which are 868 MHz only and cannot have heard it.

## Why

`GetNodeHealth` (`cmd/server/store.go`) builds the list from `s.byNode[pubkey]`. That index holds every transmission the node is associated with: as originator, as a `destPubKey`/`srcPubKey`, and as a **resolved relay hop**. For each such transmission it credits `pkt.ObserverID` with `pkt.SNR` / `pkt.RSSI`, the transmission-level representative observation. So the card answers "which observers saw traffic involving this node, anywhere in the path", and the signal numbers belong to whichever node last transmitted the copy that observer received.

The relay-hop half of that is also frequently wrong. Path hop hash size is chosen by the originator and defaults to one byte (`Packet.h:83`, `Mesh.cpp:649`, `CommonCLI.h:69` in the MeshCore firmware), so many packets carry one-byte hops. Measured on a live instance over 7 days: **28,204 of 80,217** path-carrying transmissions still use one-byte hops.

`prefixMap.resolveWithContext` never abstains on an ambiguous prefix: tiers 2, 3 and 4 always return a winner. For the `a4` prefix there are 11 candidate repeaters, and the geo_proximity tier picks the 433 node because it happens to sit in the middle of the 868 traffic it is being credited with:

```
GET /api/resolve-hops?hops=A4&observer=<an 868 observer>
→ {"name":"NL-BXE-RP01 | 433","ambiguous":true,"confidence":"geo_proximity"}
```

Over 14 days that node appears as an `A4` hop in 7,910 observations across 35 observers.

## What the data actually supports

Reading the firmware rather than guessing:

- Flood routes build the path up as they travel. A forwarding repeater appends its own hash before retransmitting (`Mesh.cpp:349`), so the **last hop is the node the observer heard**.
- Direct routes carry the **remaining** route. A forwarder matches itself against the head of the path and calls `removeSelfFromPath` (`Mesh.cpp:89,103`) before retransmitting, so the transmitter is not in the path at all. **38%** of transmissions over 7 days are direct routed.
- An empty flood path means the originator was heard directly. Only ADVERTs carry the originator's pubkey in the clear.

Applying that, plus refusing to guess on an ambiguous last hop, the card shrinks a lot, which is the honest reading:

| node | current card | direct-RF |
|---|---|---|
| BE-BRU-Moris | 36 observers | 3 |
| BE-KRO-RP01 \| ON1KW | 40 | 3 |
| BE-BRE-ON8AR | 38 | 2 |
| NL-BXE-RP01 \| 433 | 35 | 0 |

Network-wide over 7 days, **234 of 1,860 nodes** have at least one direct observer, 161 of them exactly one, maximum 8. Most nodes have no MQTT observer within radio range at all.

## Two other things this touches

- `GetBulkHealth` carries a second, near-identical copy of the same observer loop, and `public/analytics.js` sorts a leaderboard on its `observers.length`. Fixing only the node page would leave the two contradicting each other.
- Both builders return `[]map[string]interface{}` rows, which the AGENTS.md ratchet wants reduced.

## Out of scope

`resolveWithContext` keeps guessing on ambiguous hops for paths, neighbor edges and analytics. Making it abstain is a much larger change with network-wide effects and deserves its own issue.

efiten:
Closes #2056.

## What changes

The node detail "Heard By" card now lists only observers that received the node's **own transmission off the air**, and reports the rest as a count.

```
HEARD BY — DIRECT (8 OBSERVERS)
OBSERVER                REGION  PACKETS  AVG SNR   AVG RSSI
BE-DUF-SiSCD-01         —        16276    7.9 dB   -108 dBm
BE-BRU-Moris  repeater  —        13775   -5.6 dB   -122 dBm
...
Seen via relay by 29 observers. Those observers heard a repeater that
forwarded this node's traffic, not this node.
```

and for a node nothing hears:

```
HEARD BY — DIRECT (0 OBSERVERS)
No observer is within radio range of this node.
Seen via relay by 2 observers. …
```

## The rule, and where it comes from

Read out of the firmware rather than assumed:

| | |
|---|---|
| `Packet.h:83` | `setPathHashSizeAndCount(sz,n) { path_len = ((sz-1)<<6) \| (n&63); }` — hash size rides in the packet's `path_len` byte |
| `Mesh.cpp:649,678` | only `sendFlood()` sets it, so the **originator** decides; `CommonCLI.h:69` defaults `path_hash_mode = 0`, i.e. one byte |
| `Mesh.cpp:349` | a forwarding repeater appends its hash with the packet's size — it cannot upgrade a packet, and the **last hop is who was heard** |
| `Mesh.cpp:89,103` | on a direct route a forwarder matches the head of the path and calls `removeSelfFromPath` before retransmitting, so the path is the **remaining** route and the transmitter is not in it |

So an observation credits exactly one node:

1. Route type must be `ROUTE_TYPE_FLOOD` or `ROUTE_TYPE_TRANSPORT_FLOOD`. Direct routes never qualify (38% of transmissions over 7 days).
2. Empty path → the originator, known only for ADVERTs.
3. Otherwise the last hop.
4. The hop must resolve to exactly one candidate. Same gate `resolvePathForObsColdLoad` already applies: under-attribute rather than guess. It drops 418,530 of 1,455,721 flood observations with a path over 7 days (28.8%), and it is what stops the wrong-band credits.

## Measured effect

| node | before | after |
|---|---|---|
| BE-BRU-Moris | 36 observers | 3 |
| BE-KRO-RP01 \| ON1KW | 40 | 3 |
| BE-BRE-ON8AR | 38 | 2 |
| NL-BXE-RP01 \| 433 | 35 | 0 |

Network-wide over 7 days, 234 of 1,860 nodes have at least one direct observer (161 have exactly one, maximum 8). The direct list is therefore empty for most nodes, with the relay count below it. That is the correct reading: no observer is in radio range of them.

Independent corroboration on staging: for BE-WIL-3EIK-01 the eight direct observers are exactly the top eight entries of its Neighbors table by score and observation count.

## Perf justification

`GetNodeHealth` is fast today precisely because it never walks observations — it uses one representative observation per transmission. Direct-RF needs the per-observation path, and that cannot be a per-request walk: the reference store holds **232,928 transmissions / 2,887,861 observations**, one node's `byNode` slice alone holds **55,458 transmissions / 1,450,544 observations**, and `/api/nodes/bulk-health?limit=200` would multiply that.

So the aggregate is rebuilt by a background recomputer on the existing `newAnalyticsRecomputer` pattern, published into an `atomic.Value`. Reads are `O(direct observers)`, which is **cheaper than before** — the old code built per-observer sums over every transmission in `byNode` on every request.

Proof, `BenchmarkBuildDirectHeardIndex`:

```
BenchmarkBuildDirectHeardIndex-12    1    63067900 ns/op
```

3,000,000 observations (60,000 transmissions × 50 observations, 8-hop paths, 64 candidate repeaters) in **63 ms**, once per recompute interval.

Per observation the walk does one route-type check, one backward scan of `PathJSON` for the last quoted token (no allocation, no `json.Unmarshal`), one prefix-map lookup and one counter update.

Rebuilding wholesale also means eviction needs no bookkeeping: a pass simply does not see evicted transmissions. The alternative — a field on `StoreObs` updated incrementally — would have needed the call at five construction sites (`store.go:942,1264,2854,3179`, `chunked_load.go:609`), which is the duplication that caused #1558, plus matching decrements at eviction.

## API

Both `GetNodeHealth` and `GetBulkHealth` carried a near-identical copy of the observer loop; they now share one builder.

- `observers` — direct-RF only. Same field names, so no client migration. Rows are a named `HealthObserverRow` instead of `map[string]interface{}` (one fewer occurrence in a touched file, per the AGENTS.md ratchet).
- `relayObserverCount` — new integer, observers that saw traffic through the node without hearing it. `stats.totalPackets` and `stats.avgHops` still count relayed traffic, so without this number the card would contradict the figures printed beside it.

`docs/api-spec.md` is updated for both endpoints. It also documented an `iata` field on these rows that the endpoint has never emitted; removed.

## Tests

- `cmd/server/direct_heard_test.go` — table test over the rule: flood with empty path and known originator, flood whose last hop is the node, flood whose last hop is another node, direct and transport-direct routes (never credit), ambiguous last-hop prefix, listener-only candidate, 1-byte and 2-byte hop sizes; plus aggregation and row-building.
- `cmd/server/node_health_direct_rf_test.go` — end-to-end through the handler: an observer that only saw relayed traffic must not appear in `observers` but must be counted in `relayObserverCount`. Plus the benchmark.
- `tests/unit/test-direct-rf-heard-by.js` — slices the card template out of `public/nodes.js` and evaluates it, so it tests the shipped markup rather than a copy: heading, empty state, relay line, singular/plural, signal columns, listener/repeater badge tri-state.
- `cmd/server/node_health_can_relay_case_1290_test.go` — updated to seed a genuinely direct reception, since a relay-only observer no longer carries a badge.
- `cmd/server/analytics_recompute_after_load_test.go` — recomputer count 10 → 11.

Verified locally: `cmd/server` suite green, `sh test-all.sh` green (180 suites), `tests/e2e/test-e2e-playwright.js` 131/134 passed with 3 skipped and 0 failures against the seeded fixture, plus `test-issue-1147-section-order-e2e.js`, `test-issue-1151-orphan-separators-e2e.js` and `test-issue-1281-location-row-e2e.js`, which all assert on this card. `gofmt` clean, `vet` clean across all modules.

Browser-validated on staging: both the full detail page and the side pane, on a node with 8 direct observers and on the 433 MHz node with none. No console errors.

## What this does not do

`prefixMap.resolveWithContext` still guesses on ambiguous hops, so paths, neighbor edges and analytics keep their current attribution. Making it abstain is a much larger change and needs its own issue.

The "Regions" line and Region column on this card read `o.iata`, which this endpoint has never emitted, so both have always been dead. Left as found rather than widened into this change.


dborup:
Post-merge note on the "direct routes never attribute" rule: zero-hop adverts look like they are being dropped, even though they are the clearest direct-RF evidence available.

**What the firmware does**

`Mesh::sendZeroHop` sends with `ROUTE_TYPE_DIRECT` and `path_len = 0`. The transport overload uses `ROUTE_TYPE_TRANSPORT_DIRECT`, also with `path_len = 0` ([Mesh.cpp#L717-L734](https://github.com/meshcore-dev/MeshCore/blob/e94125987ed87497e706a0b54d1e80c709343980/src/Mesh.cpp#L717-L734)).

Both repeaters and companions send their own advert this way:

- Repeaters, for the periodic local advert: [simple_repeater/MyMesh.cpp#L1032-L1037](https://github.com/meshcore-dev/MeshCore/blob/e94125987ed87497e706a0b54d1e80c709343980/examples/simple_repeater/MyMesh.cpp#L1032-L1037) and [#L1302-L1303](https://github.com/meshcore-dev/MeshCore/blob/e94125987ed87497e706a0b54d1e80c709343980/examples/simple_repeater/MyMesh.cpp#L1302-L1303).
- Companions: [companion_radio/MyMesh.cpp#L1264](https://github.com/meshcore-dev/MeshCore/blob/e94125987ed87497e706a0b54d1e80c709343980/examples/companion_radio/MyMesh.cpp#L1264).

**Why this is direct RF**

A direct-routed packet with an empty path that carries an ADVERT cannot have been forwarded:

- Adverts only leave a node through `sendFlood` or `sendZeroHop`. They are never routed along a stored path.
- `path_len == 0` on a direct route is the firmware's own marker for "zero hop" (see the comment on L721).

So the observer received the advertiser's own transmission, and the advert carries that node's pubkey in the clear.

**Why it is currently dropped**

`directHeardNode` returns `""` for every route type other than `RouteFlood` and `RouteTransportFlood` before it looks at the path. So a node that observers only hear through its local zero-hop advert gets 0 direct observers and is listed under "Seen via relay" instead. I suspect this explains part of the low numbers in the description ("NL-BXE-RP01 | 433 → 0", 234 of 1,860 nodes with any direct observer). I have not measured how much.

**Suggested change**

Accept `RouteDirect` and `RouteTransportDirect` when the observation's path is empty and the transmission is an ADVERT, and attribute it to the advertised pubkey. Every other direct case keeps the current "never credit" rule, since a direct packet with a non-empty path, or an empty path with any other payload type, still says nothing reliable about the transmitter.

```go
hop := lastPathHop(obs.PathJSON)
switch *tx.RouteType {
case RouteFlood, RouteTransportFlood:
	// unchanged
case RouteDirect, RouteTransportDirect:
	if hop == "" {
		return advertOriginPubkey(tx) // zero-hop advert: sendZeroHop sets path_len = 0
	}
	return ""
default:
	return ""
}
```

`TestDirectHeardNode` currently covers direct routes only with `PayloadTXT_MSG`. Two more cases would pin this down:

- a `RouteDirect` + `PayloadADVERT` observation with an empty path should credit the advertiser;
- a `RouteTransportDirect` + `PayloadADVERT` observation with an empty path should do the same.

efiten:
Checked @dborup's note against the firmware and against a production database. The mechanism is right and the fix is #2064; the suspected scale is not supported.

**Confirmed.** `Mesh::sendZeroHop` sets `ROUTE_TYPE_DIRECT` with `path_len = 0` (commented there as "path_len of zero means Zero Hop"), the transport overload the same with `ROUTE_TYPE_TRANSPORT_DIRECT`. Repeaters send the periodic local advert through it, and `sendSelfAdvertisement` with `flood=false` does too. So an ADVERT on a direct route with an empty path cannot have been forwarded, and `directHeardNode` was rejecting exactly those at its route-type switch before reaching the `hop == ""` branch.

**Measured, 7 days:**

| | |
|---|---|
| zero-hop advert observations dropped | 8,166 |
| distinct nodes evidenced | 139 |
| node-observer pairs evidenced | 185 |
| pairs not already credited via an empty-path flood advert | **52** |

So 52 node-observer pairs of real direct evidence are invisible today, against 217 credited from flood adverts.

**Not supported: that this explains the low counts here.** Of 30 sampled nodes with zero-hop evidence, 29 already show at least one direct observer, because they also flood their adverts. And `NL-BXE-RP01 | 433`, the "35 → 0" row in the table above, has **zero** zero-hop adverts in the window, so its empty list has another cause. The fix mostly enriches lists that are already populated.

That leaves the empty state itself overclaiming, which is #2063: it says "No observer is within radio range of this node" on evidence that cannot establish it. Sampling 40 repeaters after this merged, 16 showed that line and all 16 had a non-zero `relayObserverCount` — the card contradicting itself two lines apart.

Issue #2058, opened 2026-09-21
GetChannels/GetEncryptedChannels: cold query cost is driven by payload_type, not region (~10-13s solo) — EXPLAIN QUERY PLAN + numbers


anieto:
Follow-up to #2029, per @efiten's ask there: numbers to judge a query-plan restructure by, not just "should be faster." This is a report, not a PR — I didn't attempt the actual rewrite (see Not done).

## Setup

All of this is against production `meshcore.db` (v3 schema, MeshTexas), read via a bind-mounted `keinos/sqlite3` container. Everything marked "production" below is `-readonly` against the live WAL-mode file; nothing wrote to it. The `ANALYZE` experiment further down ran against a `VACUUM INTO` copy in scratch space, verified with `PRAGMA quick_check` before and after — production was never touched.

Row counts:

| | |
|---|---|
| total transmissions | 464,667 |
| `payload_type=5` transmissions (channel messages, incl. `enc_*`) | 46,396 |
| total observations | 1,977,567 |
| total observers | 71 |
| observers with `iata=SAT` | 19 |
| observers with `iata=DFW` | 26 |
| observations belonging to SAT's 19 observers, **all payload types** | 303,220 |

## Current behavior

`GetChannels`'s outer query, timed cold, region-filtered:

| region | time |
|---|---|
| SAT | 13.19s (cold), 11.45s (rerun — so not a page-cache effect) |
| DFW | 10.44s |

`EXPLAIN QUERY PLAN`:
```
|--SEARCH t USING INDEX idx_transmissions_payload_type (payload_type=?)
|--SEARCH o USING COVERING INDEX idx_observations_dedup (transmission_id=?)
|--SEARCH obs USING INTEGER PRIMARY KEY (rowid=?) LEFT-JOIN
|--USE TEMP B-TREE FOR GROUP BY
|--CORRELATED SCALAR SUBQUERY 1
|  |--SEARCH t2 USING INDEX idx_transmissions_channel_hash_payload (channel_hash=? AND payload_type=?)
|  |--SEARCH o2 USING COVERING INDEX idx_observations_dedup (transmission_id=?)
|  `--SEARCH obs2 USING INTEGER PRIMARY KEY (rowid=?) LEFT-JOIN
`--USE TEMP B-TREE FOR ORDER BY
```
[...]

Issue #1910, opened 2026-08-23
perf: /api/stats and /api/observers degrade to 10-17s under concurrent load (18k+ observers, ~4M observer_upserts/5min cycle)


hb9tvw:
## Environment
- CoreScope version: v3.9.2
- Deployment: Docker (ghcr.io/kpa-clawbot/corescope:v3.9.2), single VPS, 11.68GB RAM allocated
- Fronted by: Apache reverse proxy + Cloudflare
- Scale: 18,125 observers total (9,644 online / 1 stale / 8,480 offline), 493 nodes, ~46,700 packets

## Symptom
The Observer Status page (`/#/observers`) hangs on "Loading..." for 10-20+ seconds, reproducible on both desktop and mobile browsers. Browser console (app's own [SLOW API] instrumentation) shows:

[SLOW API] /stats took 17167ms
[SLOW API] /stats took 16282ms
[SLOW API] /stats took 10150ms
[SLOW API] /observers took 3239ms
[SLOW API] /observers/clock-skew took 964ms

## Isolated single-endpoint timing (not slow)
curl -s http://localhost:3000/api/stats — consistently 7-15ms locally, 130-350ms via public domain, single sequential requests.

8x concurrent requests to /api/stats alone: all still under 70ms.

## Reproduces under mixed concurrent load
Running stats + observers + nodes + channels + clock-skew concurrently (mimicking what the Observers page actually fires on load):

time (
  curl -s http://localhost:3000/api/stats -o /dev/null &
  curl -s http://localhost:3000/api/observers -o /dev/null &
  curl -s "http://localhost:3000/api/nodes?limit=500&offset=0" -o /dev/null &
  curl -s http://localhost:3000/api/channels -o /dev/null &
  curl -s http://localhost:3000/api/observers/clock-skew -o /dev/null &
  wait
)
# real 0m6.905s

vs. single stats endpoint hit 8x concurrently staying under 70ms each — the slowdown only appears with the realistic multi-endpoint mix, not from raw concurrency on one endpoint.

## Possibly related: ingestor upsert volume
Ingestor stats log every ~5 minutes shows:

[ingestor] [stats] tx_inserted=6401 tx_dupes=11270 obs_inserted=17671 node_upserts=6658 observer_upserts=3968009 write_errors=0 sig_drops=0

[...]

Issue #1203, opened 2026-05-16
bug(path-inspector): 503 "service warming up" loop — ensureNeighborGraph rebuild storm, no singleflight, no stale-while-revalidate


Kpa-clawbot:
## Summary
`/api/paths/inspect` returns 503 `{"retry":true}` indefinitely on a busy store. Frontend shows "Service warming up, retry in a few seconds." but recovery never happens — each new request triggers its own synchronous full-graph rebuild that times out before completing.

## Repro
1. Open `/#/tools/path-inspector?prefixes=07,53` on a staging-sized store (≥70k tx).
2. UI displays "Service warming up, retry in a few seconds."
3. Retry indefinitely → same response.

## Root cause
`cmd/server/path_inspect.go:170-189` handler logic:

```go
go func() {
    s.store.ensureNeighborGraph()
    close(rebuilt)
}()
select {
case <-rebuilt:
    graph = s.store.graph
case <-time.After(2 * time.Second):
    writeError(503, "retry": true)
    return
}
```

And `cmd/server/path_inspect.go:421-427`:

```go
func (s *PacketStore) ensureNeighborGraph() {
    if s.graph != nil && !s.graph.IsStale() {
        return
    }
    g := BuildFromStore(s)
    s.graph = g
}
```

Three compounding bugs:

### Bug 1: No singleflight on `ensureNeighborGraph`
Each concurrent request that arrives after `builtAt + neighborGraphTTL` (5 min, `neighbor_graph.go:21`) spawns its OWN `BuildFromStore` goroutine. Under any non-trivial request rate, the store churns building multiple parallel graphs that all assign to `s.graph` — wasted CPU, cache pressure, and no convergence on a fresh `builtAt`.

### Bug 2: Synchronous rebuild gated on 2s deadline
`BuildFromStore` on a 70k-tx store is well over 2s when contended (verified on staging: persistence locks + `s.mu` contention slow it). The handler's 2s `time.After` always fires before the rebuild lands. Result: 503 returned, but the rebuild goroutine continues — and if it completes, the NEXT request might get a fresh graph. But under load every request hits an in-progress rebuild and times out anyway.

### Bug 3: No background refresh before stale
The 5-minute TTL means the graph goes from "fresh" to "stale" instantly at the boundary. There's no warmer / pre-refresh that rebuilds slightly before expiry, so every 5-minute boundary is a thundering herd if the path-inspector is being polled.

## Why this is showing up now
[...]

PR #2059, opened 2026-09-21
perf(channels): coalesce concurrent GetChannels/GetEncryptedChannels cache misses

### LINKED ISSUE #2029: GetChannels/GetEncryptedChannels: no request coalescing on cache-miss (same class as #1910)
`GetChannels`/`GetEncryptedChannels` in `cmd/server/db.go` cache their region-scoped result for 60s (`channelsCacheTTL`), but the cache-miss path has no request coalescing. Every request that arrives while the cache is expired reruns the full region-scoped `GROUP BY` scan itself.

This is the same bug class already fixed for `/api/stats`'s observations scan in #1910 (`statsSF singleflight.Group`) and for the region-membership cache (`regionMembershipSF`) — just never applied to the channels caches.

**Measured on a ~1.1GB production DB (41k `payload_type=5` rows, 260 distinct channels):**
- 1 solo request on a cache miss: 3-8s (query itself is expensive — see note below)
- 5 concurrent requests for the same region during a cache miss: ~8s **each**, i.e. no cheaper than running the query 5 separate times

A single page load that fires channels + stats + nodes + observers at once (or just two browser tabs open on the Channels view) reliably hits this every time the 60s cache lapses.

**Fix applied locally:** wrapped the query-building/execution/cache-populate block of both `GetChannels` and `GetEncryptedChannels` in a `singleflight.Group.Do(regionParam, ...)`, keyed per region, mirroring the exact pattern `regionMembershipSF` already uses (double-check the cache inside the flight in case a winner just populated it). Verified: 5 concurrent requests for the same never-before-cached region now finish within ~2ms of each other (one shared execution) instead of each independently taking the full query time. All existing `TestGetChannels*` tests pass; no new failures in the full `cmd/server` suite (only the pre-existing flaky `TestHandleNodePaths_*` group, confirmed flaky against unpatched `master` too).

**Separate, deeper issue also found while investigating (not fixed here):** the query itself is slow even without concurrency — `EXPLAIN QUERY PLAN` shows the outer query driven by `SEARCH t USING INDEX idx_transmissions_payload_type (payload_type=?)`, i.e. it scans every `payload_type=5` transmission and joins through `observations`/`observers` to filter by region, rather than starting from the region/observer side (which has a much smaller, indexed working set) and joining outward. On this DB a cold solo request already runs 3-8s before any concurrency is involved. Happy to open a separate issue with a proposed query restructure if that'd be useful — didn't want to bundle a bigger rewrite into this fix.

anieto:
Fixes #2029.

## What was wrong

`GetChannels`/`GetEncryptedChannels` (`cmd/server/db.go`) cache their region-scoped result for 60s but had no request coalescing on a cache miss, so every request that arrived while the cache was cold or expired ran the region-scoped `GROUP BY` scan itself. Measured on production: 5 concurrent requests for the same never-cached region each took ~8s, no cheaper than 5 independent runs.

`statsSF`/`regionMembershipSF` already fix the identical bug class elsewhere in this file (#1910), so this wraps both functions' query-build/execute/cache-populate block in a `singleflight.Group` the same way, keyed per region, double-checking the cache inside the flight in case a previous winner already refreshed it.

## Tests

The review on #2029 pointed out that a timing-based check ("finish within ~2ms of each other") doesn't actually prove coalescing happened — it would pass on a fast machine even without singleflight. `db_channels_singleflight_test.go` uses a call counter instead, same pattern as `TestEnsureNeighborGraph_Singleflight` (#1203 Pair A):

- `TestGetChannels_SingleflightCoalescesQueries` / `TestGetEncryptedChannels_SingleflightCoalescesQueries`: 10 concurrent callers against a cold cache, asserting the real query runs exactly once. A test-only hook (`channelsQueryHook`/`encChannelsQueryHook`, nil in production, same contract as `bgLoaderEntryHook`) increments the counter right where the query executes, since these functions hit `db.conn.Query` directly rather than going through an injectable builder function.
- `TestGetChannels_SingleflightPerRegion`: two regions queried concurrently (5 callers each) assert 2 queries, not 1 — pins that the flight is keyed per-region and a caller for one region can't receive another region's coalesced result.

Anti-tautology: reverting `channelsSF.Do`/`encChannelsSF.Do` back to a bare call makes the coalescing tests observe N instead of 1.

`go build ./...`, `go vet ./...`, `gofmt -l .` clean. Full `cmd/server` suite (race-enabled for the new concurrency tests) run in a `golang:1.22-alpine` container, mounted repo, workdir `cmd/server` so the sibling `internal/*` replace directives resolve:

```
=== RUN   TestGetChannels_SingleflightCoalescesQueries
--- PASS: TestGetChannels_SingleflightCoalescesQueries (0.06s)
=== RUN   TestGetChannels_SingleflightPerRegion
--- PASS: TestGetChannels_SingleflightPerRegion (0.05s)
=== RUN   TestGetEncryptedChannels_SingleflightCoalescesQueries
--- PASS: TestGetEncryptedChannels_SingleflightCoalescesQueries (0.07s)
```
Full suite: `FAIL github.com/corescope/server 98.261s`, but the only failures are `TestHandleNodePaths_PrefixCollision_1352`, `TestHandleNodePaths_FallbackUniquePrefix_1352`, and `TestHandleNodePaths_FallbackUnresolvableHop_1352`, all failing on a `503 {"error":"index loading","retryAfter":5}` — an index-build race in this container's timing, not this change. Confirmed by running the same three against an unmodified, freshly-cloned `master` in the same container: they fail there too (plus `TestHandleNodePaths_PrefixCollision_1352_FallbackBranch`, which this run happened not to hit). Nothing in this diff touches node-path handling.

## Not done

The deeper query-plan issue flagged in #2029 (the outer scan is driven by `payload_type`, not region, so a cold solo request still costs several seconds regardless of concurrency) is filed separately as #2058, with `EXPLAIN QUERY PLAN` output and row counts against production data. Coalescing makes one slow query serve everybody; it doesn't make the query itself fast.


efiten (approved):
Thanks for turning this round, and for taking the counter over the timing assertion — that was the half of #2029 that would have shipped a test proving nothing.

Checked the two things that go wrong with singleflight:

**The flight key covers the whole input.** `GetChannels(region ...string)` takes nothing else, and `db.channelsSF.Do(regionParam, ...)` uses the same `regionParam` the cache is keyed on, so a flight and a cache entry can never disagree about which region they hold. `TestGetChannels_SingleflightPerRegion` pins it from the other side by asserting 2 queries for 2 regions rather than 1.

**The in-flight re-check is there**, so a caller that joins a flight started just before a winner populated the cache still gets the fresh value instead of a second scan.

The test-only hooks are the right call here given these functions reach `db.conn.Query` directly rather than through an injectable builder, and `nil` in production costs a branch. Same contract as `bgLoaderEntryHook`, so it is one pattern in the file rather than two.

Also correct to split the query plan into #2058 instead of bundling it. Coalescing makes one slow query serve everybody; it does not make it fast, and those are separate claims that want separate evidence.

One observation, not a blocker and pre-existing: `"BE,NL"` and `"NL,BE"` are distinct keys for both the cache and the flight, so they miss each other's work even though `normalizeRegionCodes` would fold them. Cache and flight are wrong together, which is the safe way to be wrong. Worth a normalise-before-key only if callers actually vary the order.

Note on the checks: CI had not run at all on this PR — GitHub had it at `action_required`, which blocks workflows from a first-time contributor until a maintainer approves. That is on our side, not yours. I approved the run, so the pipeline is going now. Approving the change pending green.

PR #2061, opened 2026-09-22
fix(nodes): keep paginating past a page that post-LIMIT filtering shortened

liquidraver:
## Problem

`handleNodes` runs the geo-filter, `nodeBlacklist`, `hiddenNamePrefixes` and area
passes **after** the SQL `LIMIT/OFFSET`, and rewrites `total` to the filtered
length. A page that loses a row is therefore short **without being the last
page**, and neither the page length nor `total` can tell a client whether to ask
for another page.

#1606 added the pagination loop and chose the page length as the canonical stop.
That is correct only where nothing is ever filtered. Everywhere else the list
truncates at the first filtered page boundary and strands every node behind it —
the #1598 symptom reached by a different route: a node that is relaying right
now simply stops being in the list.

The comment at `app.js:240` rejects `total` for exactly the right reason, then
picks the signal the same code path also breaks.

## Measured on a live 2346-node deployment

Page sizes for the query the map issues:

```
offset=0     returned=500     ← full, loop continues
offset=500   returned=499     ← one row filtered AFTER the LIMIT → loop STOPS
offset=1000  returned=500     ← never requested
offset=1500  returned=500     ← never requested
offset=2000  returned=345     ← never requested
```

| stop rule | requests | nodes reached |
|---|---:|---:|
| short page (master) | 2 | **999** |
| `has_more`, else empty page | 6 | **2344** |

**1341 nodes, 57%, unreachable through the UI.**

### One hidden node truncates the whole list

The deployment this came from has no `geoFilter` (`/api/config/geo-filter`
returns `polygon: null`) and no `nodeBlacklist`. It has a single
`hiddenNamePrefixes` entry — a deliberate operator choice — and exactly one node
whose name starts with it:

```
public_key   d4a46ea2…1054      (64 clean hex chars)
name         🚫🔥☀️
role         repeater
last_seen    2026-09-22T10:09:38Z
```

`handleNodes` drops that row in the `IsNameHidden` pass, which runs after the SQL
`LIMIT`. The row is counted by the `LIMIT` and by `COUNT(*)`, so the page it lands
in comes back exactly one short — and stops every client that treats a short page
as the end.

Isolated against SQL on the same database, seconds apart:

```
SELECT lower(public_key) FROM nodes ORDER BY last_seen DESC LIMIT 500 OFFSET 500
  -> 500 rows
GET /api/nodes?limit=500&offset=500
  -> 499 rows
comm -23 sql.txt api.txt
  -> d4a46ea2e99cab132a3286ef3d9cce9099318790af7f25671fe83de453721054
```

Deterministic — `offset=500` returned 499 on three consecutive requests. Not a
CDN artifact either: `cf-cache-status: DYNAMIC`, origin `cache-control:
no-store`, no `age` header, and four requests with deliberately unique cache keys
all returned 499.

So **one deliberately hidden node makes 1341 of 2344 nodes unreachable.** The
hiding feature does exactly what it was asked to do for that one node, and takes
57% of the network with it, silently. A single `hiddenNamePrefixes` entry is
enough; no geo-filter, blacklist or area filter is needed to reach this state.

### The cutoff moves, which is why this reads as intermittent

The visible set is the sum of the pages up to and including the first short one,
so the boundary sits wherever the unreturnable row currently sorts by
`last_seen`, and jumps a whole page as ingest reorders the list. Same
deployment, same code, same config, ~2h apart:

| dropped row's rank | first short page | nodes visible |
|---|---|---:|
| inside 0–499 | page 1 | 499 |
| inside 500–999 | page 2 | 999 |

A node is visible or invisible purely by where it lands relative to that moving
line, so affected nodes appear to vanish and return on their own. Two operators
on this deployment reported exactly that, independently, while I was measuring.

### A named reproduction

`HU-ZA-Lentihegy` (`5287a33f…`), reported missing from the map by an operator
whose companion had logged its advert at 04:20 local the same morning.

Ingest was fine. The row is in `nodes` with `last_seen`
`2026-09-22T02:20:35Z` — the same advert, to the second — valid GPS, role
`repeater`, 1033 adverts, and `/api/nodes/search?q=lentihegy` returns it.

```
rank by last_seen : 1081
cutoff at the time:  999
```

It missed by 82 positions. Walking the same live endpoint, same moment:

| stop rule | requests | nodes reached | Lentihegy |
|---|---:|---:|---|
| short page (master) | 2 | 999 | **not reached** |
| `has_more`, else empty page | 6 | 2340 | reached |

The practical shape of this on a busy mesh: 1081 nodes had been heard more
recently than 9.4 hours, so on that deployment **anything last heard more than
~9 hours ago was invisible**, alive or not.

`#/nodes` compounds it — its search box filters client-side over the truncated
set, so the server-side `?search=` never runs and an operator cannot find the
node by searching for it either, even though the endpoint would return it.

## Change

**Server** — `NodeListResponse` gains `has_more`, computed from the raw SQL page
against the real `COUNT(*)` before the filter passes run, so it survives them:

```go
hasMore := offset+len(nodes) < total
```

Always emitted (no `omitempty`) so a client can tell `false` from an old server.
No extra request in the fixed path: `has_more` ends the loop exactly, where the
old rule needed a probe page.

**Clients** — `app.js` `fetchAllNodes`, `nodes.js` `loadNodes` and
`area-map.html`'s inline helper stop on `has_more`, falling back to a zero-length
page against a server that predates it. An empty page always ends the loop, so a
`has_more` against a concurrently-shrinking table cannot spin to `safetyCap`.

Left alone: the three loops are still three copies. Collapsing them onto
`fetchAllNodes` is a bigger change than this fix needs, and `nodes.js` has its
own inter-page progress UI. Happy to do it separately if you want it.

## Testing

- **Unit** (`tests/unit/test-fetch-all-nodes-pagination.js`): the fixture now
  models the real handler — a row counted by the LIMIT and by `COUNT(*)`, then
  removed from the page. Three new cases. Fails on the old rule at 499 of 1199.
- **E2E** (`tests/e2e/test-map-nodes-pagination-e2e.js`, already wired into
  `deploy.yml`): the mock drops a page-1 row and emits `has_more`.
  Mutation-checked — restoring master's stop rule fails 3 of its steps.
- **Go** (`cmd/server/nodes_pagination_has_more_test.go`): asserts `has_more`
  stays true on a page filtering shortened. Mutation-checked — recomputing it
  after the filter block fails the test.
- Full server suite `go test -race`: ok, 41.4s. `gofmt` clean, `go vet` passes.
- **Against a real binary**, not just mocks: fixture DB migrated with
  `corescope-migrate`, `hiddenNamePrefixes: ["SKCE"]`, `limit=3`. Page 1 returns
  2 of 3 with `total` rewritten to 2 and `has_more=true`. Walking the real server
  with master's rule reaches 2 nodes; with `has_more`, all 199 visible of 200,
  the hidden one still hidden. The real frontend against that server loads 199
  with no JS errors.

Two existing expectations changed, both deliberate:

1. `surfaces ALL nodes past the 500 server cap` — 3 → 4 requests. That mock emits
   no `has_more`, so the 200-row final page can no longer end the loop (a short
   page is exactly what a filtered page looks like) and a zero-length probe
   follows. Against a current server `has_more` still ends it at 3.
2. `rows missing public_key are NOT collapsed into one` — its stub returned a
   constant body, which would now be paged to `safetyCap`. It serves one page
   then empties.

Local `test-all.sh` exits 1 on two XSS-gate self-tests
(`good-2-tested.js`, `good-4-tested.js`) via a `UnicodeEncodeError` printing an
emoji under Windows cp1252. Identical on clean `origin/master` in a scratch
worktree, so it is pre-existing and platform-local, not this branch.

There is a second identical filter block further down `routes.go` on another list
endpoint. Likely the same class; not touched here.

If you would rather land your own version of this, say so and I will close mine.


efiten (approved):
Reviewed the contract, the loop and the fallback. Sound, and the diagnosis reproduces independently.

**`has_more` is derived in the one place it can be trusted.** `offset+len(nodes) < total` sits before the blacklist / hidden-prefix / geo-filter / area passes, so it uses the raw SQL page against the real `COUNT(*)`, and `TestNodesHasMoreSurvivesPostLimitFiltering` pins exactly that ordering. Moving that line below those passes would silently restore the bug, and the test would catch it.

**The loop cannot spin.** Three things I went looking for, all present: the empty-page check runs *before* `has_more`, so a server reporting `has_more` against a concurrently shrinking table ends rather than walking to the safety cap; the `typeof data.has_more === "boolean"` guard keeps old servers working, falling back to the zero-length page rather than the short page; and the safety cap is still there underneath.

**Corroborated on a second deployment.** On a 1895-node instance, `total` comes back as 500, 500, 500, 395, 0 across the five pages — it equals the page length every time. So `total` is the filtered length there too, and the description's reason for rejecting it as a stop signal holds beyond the deployment you measured.

That instance happens not to truncate: no page is short before the last one, because nothing currently filters a row out of a non-final page. Which is the point. It is one `hiddenNamePrefixes` entry away from losing the tail of its list, with nothing in the UI to indicate it. The fix is worth taking on deployments that look unaffected today.

One consequence worth stating rather than leaving implicit: `offset+len(nodes) < total` can read false if rows are deleted between the `COUNT(*)` and the page query, which stops a page early instead of looping. That is the right way round for this trade, and the next load recovers it.

Approving.

